PR #2, opened 2019-05-13
Add swift tests

gnuoy:


zaza/openstack/charm_tests/swift/tests.py
```diff
@@ -0,0 +1,82 @@
...
+        cls.image_name = 'zaza-swift-lts'
+        swift_session = openstack_utils.get_keystone_session_from_relation(
+            'swift-proxy')
+
+        cls.swift = openstack_utils.get_swift_session_client(
+            swift_session)
+        cls.glance_client = openstack_utils.get_glance_session_client(
+            cls.keystone_session)
+
+    def test_100_create_image(self):
```
ChrisMacNaughton: This looks a lot like a test that exists for [Ceph RadosGW](https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/charm_tests/ceph/tests.py#L580)

gnuoy: Good point, I'll try and get them both using the same test,

gnuoy: I'm backtracking here, but I have raised an issue saying two objects stores should share tests. However I think we should push ahead with the PR as it is based off what we currently do with amulet. 

https://github.com/openstack-charmers/zaza-openstack-tests/issues/3

PR #16, opened 2019-06-06
Add a new SSH function

ChrisMacNaughton:
It is likely that there are other commands that we
would like to run in our tests, so this is a new SSH
command runner that the existing SSH check has been
updated to use.


gnuoy:
Look great, thank you.

PR #18, opened 2019-06-11
Support initializing vault with no clients

gnuoy:
Currently when vault auto_inititialize [sic] is called a validation
step is run at the end to check that clients connected to vault
have the correct setup. However in sometimes there are no clients
to run validation with (as with cross model relation testing). To
facilitate this expose a new method 'auto_initialize_no_validation'
which skip the client validation.

Also correct the spelling of auto_inititialize but add a partial
to support tests that are already using the old spelling.

### DIFF
--- a/zaza/openstack/charm_tests/vault/setup.py
+++ b/zaza/openstack/charm_tests/vault/setup.py
@@ -14,6 +14,7 @@
 
 """Run configuration phase."""
 
+import functools
 import requests
 import tempfile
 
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/vault/setup.py
```diff
@@ -83,7 +87,16 @@ def auto_inititialize(cacert=None):
...
+    if validation_application:
+        validate_ca(cacertificate, application=validation_application)
+
+
+auto_initialize_no_validation = functools.partial(
+    auto_initialize,
+    validation_application=None)
+
+
+auto_inititialize = functools.partial(auto_initialize)
```
thedac: I'll fix the one place that calls this (keystone-saml-mellon). So we should not need the misspelled version.

PR #24, opened 2019-06-26
Add cinder-backup-swift functional tests

lolwww:
Adding functional tests for charm-cinder-backup-swift
https://github.com/openstack-charmers/charm-cinder-backup-swift

### DIFF
--- a/zaza/openstack/charm_tests/cinder_backup_swift/setup.py
+++ b/zaza/openstack/charm_tests/cinder_backup_swift/setup.py
@@ -1,0 +1,51 @@
+# Copyright 2018 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Code for configuring cinder-backup-swift."""
+
+import zaza.model as zaza_model
+import time
+
+
+def basic_setup():
+    """Run setup for testing cinder-backup-swift.
+
+    Cinder backup setup for testing cinder-backup is currently part of
+    cinder-backup functional tests.
+    Volume backup setup for other tests to use should go here.
+    """
+
+
+def configure_cinder_backup():
+    """Configure cinder-backup-swift."""
+    keystone_ip = zaza_model.get_app_ips(
+        'swift-keystone')[0]
+    swift_ip = zaza_model.get_app_ips(
+        'swift-proxy')[0]
+    auth_ver = (zaza_model.get_application_config('swift-keystone')
+                .get('preferred-api-version').get('value'))
+    if auth_ver == 2:
+        auth_url = 'http://{}:5000/v2.0'.format(keystone_ip)
+        endpoint_url = 'http://{}:8080/v1/AUTH_'.format(swift_ip)
+    else:
+        auth_url = 'http://{}:5000/v3'.format(keystone_ip)
+        endpoint_url = 'http://{}:8080/v1/AUTH'.format(swift_ip)
+    cinder_backup_swift_conf = {
+        'endpoint-url': endpoint_url,
+        'auth-url': auth_url
+    }
+    juju_service = 'cinder-backup-swift'
+    zaza_model.set_application_config(juju_service, cinder_backup_swift_conf)
+    zaza_model.wait_for_application_states()
+    time.sleep(300)
--- a/zaza/openstack/charm_tests/cinder_backup_swift/setup.py
+++ b/zaza/openstack/charm_tests/cinder_backup_swift/setup.py
@@ -15,17 +15,7 @@
 """Code for configuring cinder-backup-swift."""
 
 import zaza.model as zaza_model
-import time
-
-
-def basic_setup():
-    """Run setup for testing cinder-backup-swift.
-
-    Cinder backup setup for testing cinder-backup is currently part of
-    cinder-backup functional tests.
-    Volume backup setup for other tests to use should go here.
-    """
-
+import zaza.openstack.charm_tests.test_utils
 
 def configure_cinder_backup():
     """Configure cinder-backup-swift."""
@@ -48,4 +38,8 @@ def configure_cinder_backup():
     juju_service = 'cinder-backup-swift'
     zaza_model.set_application_config(juju_service, cinder_backup_swift_conf)
     zaza_model.wait_for_application_states()
-    time.sleep(300)
+    _singleton = zaza.openstack.charm_tests.test_utils.OpenStackBaseTest()
+    _singleton.setUpClass()
+    with _singleton.config_change(cinder_backup_swift_conf, cinder_backup_swift_conf):
+        # wait for configuration to be applied then return
+        pass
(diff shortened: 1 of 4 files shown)


zaza/openstack/charm_tests/cinder_backup_swift/setup.py
```diff
@@ -0,0 +1,51 @@
...
+        auth_url = 'http://{}:5000/v3'.format(keystone_ip)
+        endpoint_url = 'http://{}:8080/v1/AUTH'.format(swift_ip)
+    cinder_backup_swift_conf = {
+        'endpoint-url': endpoint_url,
+        'auth-url': auth_url
+    }
+    juju_service = 'cinder-backup-swift'
+    zaza_model.set_application_config(juju_service, cinder_backup_swift_conf)
+    zaza_model.wait_for_application_states()
+    time.sleep(300)
```
fnordahl: Sleep is never the right solution in a test, you need to find a way to poll until the deployment reaches the expected state.

You might be able to use this trick: https://github.com/openstack-charmers/zaza-openstack-tests/blob/bf93335017656dd6da5002a44e9783bd2c3106cc/zaza/openstack/charm_tests/octavia/setup.py#L88

zaza/openstack/charm_tests/cinder_backup_swift/setup.py
```diff
@@ -0,0 +1,51 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Code for configuring cinder-backup-swift."""
+
+import zaza.model as zaza_model
+import time
+
+
+def basic_setup():
```
fnordahl: There is no requirement for a ``basic_setup`` function, since it is empty here you could just remove it.

zaza/openstack/charm_tests/cinder_backup_swift/tests.py
```diff
@@ -0,0 +1,54 @@
...
+        size = 1
+        volume = openstack_utils.create_volume(
+            self.cinder_client,
+            size)
+        # Create volume backup
+        logging.info('Creating volume backup')
+        volume_backup = openstack_utils.create_volume_backup(
+            self.cinder_client,
+            volume.id)
+        # Delete volume backup
```
fnordahl: I think it would be pertinent to introspect the cinder backup metadata and assert that the back-end you expected to be used is actually the one used.

zaza/openstack/utilities/openstack.py
```diff
@@ -1690,6 +1690,30 @@ def delete_image(glance, img_id):
     delete_resource(glance.images, img_id, msg="glance image")
 
 
+@tenacity.retry(wait=tenacity.wait_exponential(multiplier=1, max=60),
```
fnordahl: resource_removed called from delete_resource is already using tenacity to retry the operation.  Why is the extra level of retrying necessary?

zaza/openstack/utilities/openstack.py
```diff
@@ -1690,6 +1690,30 @@ def delete_image(glance, img_id):
...
+    """Delete the given volume from cinder.
+    :param cinder: Authenticated cinderclient
+    :type cinder: cinderclient.Client
+    :param vol_id: unique name or id for the openstack resource
+    :type vol_id: str
+    """
+    delete_resource(cinder.volumes, vol_id, msg="deleting cinder volume")
+
+
+@tenacity.retry(wait=tenacity.wait_exponential(multiplier=1, max=60),
```
fnordahl: resource_removed called from delete_resource is already using tenacity to retry the operation.  Why is the extra level of retrying necessary?

zaza/openstack/utilities/openstack.py
```diff
@@ -1769,6 +1793,63 @@ def create_image(glance, image_url, image_name, image_cache_dir=None, tags=[]):
...
+    :param name: display name for new volume
+    :type name: Option[str, None]
+    :param image: Image to download to volume.
+    :type image: Option[str, None]
+    :returns: cinder volume pointer
+    :rtype: cinderclient.common.utils.RequestIdProxy
+    """
+    logging.debug('Creating volume')
+    if not size:
+        raise Exception("Size for volume not specified")
```
fnordahl: In general you would not use ``Exception`` class directly, create a new exception class so we know where the failure occurred.

However, in this case it might be sufficient to just pass the value or non-value of ``size`` on to cinder and let it raise an exception, or does it fail silently?

zaza/openstack/utilities/openstack.py
```diff
@@ -1769,6 +1793,63 @@ def create_image(glance, image_url, image_name, image_cache_dir=None, tags=[]):
...
+    :param volume_id: the source volume's id for backup
+    :type volume_id: str
+    :param name: display name for new volume backup
+    :type name: Option[str, None]
+    :returns: cinder volume backup pointer
+    :rtype: cinderclient.common.utils.RequestIdProxy
+    """
+    logging.debug('Creating volume backup')
+    if not volume_id:
+        raise Exception("volume_id not specified")
```
fnordahl: In general you would not use ``Exception`` class directly, create a new exception class so we know where the failure occurred.

However, in this case it might be sufficient to just pass the value or non-value of ``volume_id`` on to cinder and let it raise an exception, or does it fail silently?

zaza/openstack/charm_tests/cinder_backup_swift/setup.py
lolwww: done, please check

zaza/openstack/charm_tests/cinder_backup_swift/setup.py
lolwww: removed

zaza/openstack/charm_tests/cinder_backup_swift/tests.py
lolwww: Added check that backup_driver for backup in cinder is actually Swift

zaza/openstack/utilities/openstack.py
lolwww: removed

zaza/openstack/utilities/openstack.py
lolwww: removed

zaza/openstack/utilities/openstack.py
lolwww: I'm removing the Exception as it will be raised anyway.
Tried calling with no argument:
> TypeError: create_volume() missing 1 required positional argument: 'size'

Calling with "" or None as size:
> cinderclient.exceptions.BadRequest: Invalid input received: Volume size '' must be an integer and greater than 0 (HTTP 400)

zaza/openstack/utilities/openstack.py
lolwww: Same as with size.
I'm removing the Exception as it will be raised anyway.
> cinderclient.exceptions.BadRequest: Invalid input for field/attribute volume_id. Value: None. None is not of type 'string'

fnordahl:
@lolwww This is starting to look good.

A exercise of the test leads to failure, however if I re-run the test it succeeds.

I think what is going on is that you do the config change on ``cinder-backup-swift``, zaza waits for the config to be applied, and then goes on to run the test before the principal ``cinder`` charm has actually applied the configuration.

Due to how the ``cinder-backup`` interface is laid out in the ``cinder`` charm you don't really have any feedback when the config is done.

What you could do is add a second call to wait for something to happen after your config change, and then wait for everything to be idle again.

I think that would be sufficient for the test to always succeed.

Try to add a call to https://github.com/openstack-charmers/zaza/blob/6b912b7095f23d9aa707082f3821d92ed7817af8/zaza/model.py#L720

and then a call to https://github.com/openstack-charmers/zaza/blob/6b912b7095f23d9aa707082f3821d92ed7817af8/zaza/model.py#L756

after the config change at the end of your setup function.

lolwww:
I get constant first pass
https://pastebin.canonical.com/p/wGdHT52gn2/

PR #25, opened 2019-06-27
Catch unauthenticated keystone client separately

relaxdiego:
This change catches an error in the test_end_user_domain_admin_access of
zaza.openstack.charm_tests.keystone.tests.AuthenticationAuthorizationTest.
This is to avoid any ambiguity about the error.

This does not fix the bug but merely makes it obvious for easier
identification in the future.

Related-Bug: #1834287


relaxdiego:
Closing this PR since associated bug (https://bugs.launchpad.net/charm-keystone/+bug/1834287) is invalid.

PR #40, opened 2019-08-04
Add a subordinate blacklist for pausing services

ajkavanagh:
During the series upgrade tests, zaza pauses subordinate services.
However, the cinder-ceph charm has no pause/resume action, and so the
test fails.  This PR adds a blacklist that ensures that the cinder-ceph
subordinate is not paused during series upgrade.


gnuoy:
LGTM, thanks

PR #43, opened 2019-08-30
Fix domain scope settings.

thedac:


ajkavanagh:
Is this PR is getting a bit outdated? i.e. do we still need to make this change, and if so, we should merge it.  Thanks.

thedac:
Change already exists in master.

PR #49, opened 2019-09-02
Update versions to include Train

ChrisMacNaughton:


zaza/openstack/utilities/os_versions.py
```diff
@@ -61,7 +63,8 @@
     'yakkety_newton', 'xenial_ocata', 'zesty_ocata',
     'xenial_pike', 'artful_pike', 'xenial_queens',
     'bionic_queens', 'bionic_rocky', 'cosmic_rocky',
-    'bionic_stein', 'disco_stein']
+    'bionic_stein', 'disco_stein', 'bionic-train',
+    'eoan-train']
```
javacruft: underscores needed rather than dashes above.

PR #50, opened 2019-09-04
charm_tests/glance: migrate Amulet tests to Zaza

guoqiao:
This patch is the peer change for:

    https://review.opendev.org/#/c/679599/

charm-glance currently has following Amulet tests:

    https://github.com/openstack/charm-glance/blob/master/tests/basic_deployment.py

The 1xx, 2xx and 5xx ones are mostly covered in zaza keystone tests, or could
be skipped in zaza.

Add the only missing 900 one, to test glance restart on config change.

Test option `debug` instead of `use-syslog` to avoid conversion between
`-` and `_`, which is out the scope of this test.

Signed-off-by: Joe Guo <guoqiao@gmail.com>

### DIFF
--- a/zaza/openstack/charm_tests/glance/tests.py
+++ b/zaza/openstack/charm_tests/glance/tests.py
@@ -67,6 +67,44 @@ class GlanceTest(test_utils.OpenStackBaseTest):
             {'image_format': {'disk_formats': ['qcow2']}},
             ['glance-api'])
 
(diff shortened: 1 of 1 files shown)


guoqiao:
@thedac  Yes, I noticed that, but I suspected that was more about disk format config, so I still migrate this one, as practice. At least we test more options:) Thanks for the merging!

PR #53, opened 2019-09-04
use tenacity for retries

### LINKED ISSUE #46: time.sleep roundup on retries
The test_all_clients_authenticated test in vault/tests.py uses a hard-coded sleep time, wrapped by a for loop to retry.

It would be better to use tenacity retries to be consistent with the established approach to similar needs in other tests.
https://github.com/openstack-charmers/zaza-openstack-tests/blob/2244f131a058159a8329385b00d0157655328769/zaza/openstack/charm_tests/vault/tests.py#L131-L139

I feel the same about the is_initialized method in vault/utils.py:
https://github.com/openstack-charmers/zaza-openstack-tests/blob/2244f131a058159a8329385b00d0157655328769/zaza/openstack/charm_tests/vault/utils.py#L129-L141

And, same for mysql/tests.py:
https://github.com/openstack-charmers/zaza-openstack-tests/blob/2244f131a058159a8329385b00d0157655328769/zaza/openstack/charm_tests/mysql/tests.py#L380-L392

IMHO:

When we call time.sleep(), we need to affirm that we are introducing a possible race condition for our future selves. In the majority of cases where time.sleep() is sought, we should seek a different approach.

And if retries are what we're looking for, we should use a consistent and reusable approach, rather than constructing a one-off retry loop. That may require further distillation of the test procedure into one or more helper functions, but it will be better for it in the end.

guoqiao:
These code snippets were using a hard-coded sleep time, wrapped by a for loop
to retry.

Use tenacity retries to simplify the code, and be consistent with the
established approach to similar needs in other tests.

Closes #46

Signed-off-by: Joe Guo <guoqiao@gmail.com>


zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -117,30 +118,26 @@ def get_clients(units=None, cacert=None):
     return clients
 
 
+@tenacity.retry(
+    retry=tenacity.retry_if_exception_type((
+        ConnectionRefusedError,
+        urllib3.exceptions.NewConnectionError,
+        urllib3.exceptions.MaxRetryError,
+        requests.exceptions.ConnectionError)),
+    reraise=True,  # if all retries failed, reraise the last exception
```
guoqiao: If all retries failed, it will now raise the last exception, other than Exception("Cannot connect"), which slightly changed the original behavior. Since Exception is base class, it shouldn't affect the upstream code, in theory.

ajkavanagh:
This PR is getting a little outdated; @guoqiao do you want to update it and resolve the conflicts?

AurelienLourot (approved):
I'm +1 on this change and in fact I need this new behavior of re-raising the original exception instead of `Exception("Cannot connect")` (with this exception I don't understand what's going on)

So thanks a lot for doing this. I'm going to resurrect this PR by solving the merge conflicts.

AurelienLourot:
Validated the `vault/utils.py` part locally. This now raises a meaningful exception, thanks a lot! Landing

PR #59, opened 2019-09-18
Porting Ceilometer tests to Zaza

ed1000:
Porting the Amulet tests from Ceilometer to the Zaza framework.

https://opendev.org/openstack/charm-ceilometer/src/branch/master/tests/basic_deployment.py

### DIFF
--- a/zaza/openstack/charm_tests/ceilometer/setup.py
+++ b/zaza/openstack/charm_tests/ceilometer/setup.py
@@ -32,7 +32,8 @@ def basic_setup():
     xenial_pike = openstack_utils.get_os_release('xenial_pike')
 
     if current_release < xenial_pike:
-        raise unittest.SkipTest('Not checking ceilometer-upgrade')
+        raise unittest.SkipTest('Skipping ceilometer-upgrade as it is not '
+                                'supported before Pike')
 
     logging.debug('Checking ceilometer-upgrade')
 
--- a/zaza/openstack/charm_tests/ceilometer/setup.py
+++ b/zaza/openstack/charm_tests/ceilometer/setup.py
@@ -17,11 +17,12 @@
 """Code for configuring Ceilometer."""
 
 import logging
+import unittest
 import zaza.model as zaza_model
 import zaza.openstack.utilities.openstack as openstack_utils
 
 
-def basic_setup(unit='ceilometer/0'):
+def basic_setup():
     """Run setup for testing Ceilometer.
 
     Setup for testing Ceilometer is currently part of functional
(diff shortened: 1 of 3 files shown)


zaza/openstack/charm_tests/ceilometer/tests.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # Config file affected by juju set config change
+        conf_file = '/etc/ceilometer/ceilometer.conf'
+        services = {}
+        current_release = openstack_utils.get_os_release()
+        xenial_pike = openstack_utils.get_os_release('xenial_pike')
+        xenial_ocata = openstack_utils.get_os_release('xenial_ocata')
+        xenial_newton = openstack_utils.get_os_release('xenial_newton')
+        trusty_mitaka = openstack_utils.get_os_release('trusty_mitaka')
+        trusty_liberty = openstack_utils.get_os_release('trusty_liberty')
```
gnuoy: I think it makes sense to create xenial_pike, xenial_ocata etc  as class variables in the setUpClass so they can be reused by other tests

zaza/openstack/charm_tests/ceilometer/tests.py
```diff
@@ -0,0 +1,118 @@
...
+            if current_release < trusty_mitaka:
+                services['ceilometer-alarm-notifier'] = conf_file
+                services['ceilometer-alarm-evaluator'] = conf_file
+
+            if current_release >= trusty_liberty:
+                # Liberty and later
+                services['ceilometer-polling'] = conf_file
+            else:
+                # Juno and earlier
+                services['ceilometer-agent-central'] = conf_file
```
gnuoy: Please move the retrieval of the services list to a class method (or property) so that other tests can use it (see [cinder tests](https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/charm_tests/cinder/tests.py#L140)

zaza/openstack/charm_tests/ceilometer/tests.py
```diff
@@ -0,0 +1,118 @@
...
+            alternate_entry,
+            services)
+
+    def test_901_pause_resume(self):
+        """Run pause and resume tests.
+
+        Pause service and check services are stopped then resume and check
+        they are started.
+        """
+        self.pause_resume(['ceilometer'])
```
gnuoy: I don't think this is going to do anything. I think its supposed to be called as a context like [this](https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/charm_tests/cinder/tests.py#L186)

ed1000:
@gnuoy made the changes you proposed. Regarding the last item, I've seen it used both ways. Don't know which is correct though.

zaza/openstack/charm_tests/ceilometer/setup.py
```diff
@@ -0,0 +1,44 @@
...
+    """Run setup for testing Ceilometer.
+
+    Setup for testing Ceilometer is currently part of functional
+    tests.
+    """
+    current_release = openstack_utils.get_os_release()
+    xenial_pike = openstack_utils.get_os_release('xenial_pike')
+
+    if current_release < xenial_pike:
+        raise unittest.SkipTest('Not checking ceilometer-upgrade')
```
gnuoy: Please say why you are not checking it to the person running the tests knows. Somthing like: "Skipping ceilometer-upgrade as it is not supported before Pike"

PR #62, opened 2019-09-24
add plumb_dataport_method

andrewdmcleod:
if openstack version >= bionic-queens, add dataport to netplan
add "add_to_netplan" method

### DIFF
--- a/zaza/openstack/configure/network.py
+++ b/zaza/openstack/configure/network.py
@@ -219,6 +219,17 @@ def setup_gateway_ext_port(network_config, keystone_session=None):
         dvr_mode=network_config.get("dvr_enabled", False),
         net_id=net_id)
 
+    current_release = openstack_utils.get_os_release()
+    bionic_queens = openstack_utils.get_os_release('bionic_queens')
+    if current_release >= bionic_queens:
+        logging.warn("Adding second interface for dataport to guest netplan, \
+                     for bionic-queens and later")
+        openstack_utils.plumb_guest_nic(
+            nova_client,
+            neutron_client,
+            dvr_mode=network_config.get("dvr_enabled", False),
+            net_id=net_id)
+
 
 def run_from_cli(**kwargs):
     """Run network configurations from CLI.
--- a/zaza/openstack/configure/network.py
+++ b/zaza/openstack/configure/network.py
@@ -212,23 +212,23 @@ def setup_gateway_ext_port(network_config, keystone_session=None):
     else:
         net_id = None
 
+    # If we're using netplan, we need to add the new interface to the guest
+    current_release = openstack_utils.get_os_release()
+    bionic_queens = openstack_utils.get_os_release('bionic_queens')
+    if current_release >= bionic_queens:
+        logging.warn("Adding second interface for dataport to guest netplan "
+                     "for bionic-queens and later")
+        plumb_guest_nic = True
+    else:
+        plumb_guest_nic = False
+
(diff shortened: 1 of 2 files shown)


andrewdmcleod:
no longer wip

zaza/openstack/configure/network.py
```diff
@@ -212,12 +212,23 @@ def setup_gateway_ext_port(network_config, keystone_session=None):
     else:
         net_id = None
 
+    # If we're using netplan, we need to add the new interface to the guest
+    current_release = openstack_utils.get_os_release()
+    bionic_queens = openstack_utils.get_os_release('bionic_queens')
+    if current_release >= bionic_queens:
+        logging.warn("Adding second interface for dataport to guest netplan "
+                     "for bionic-queens and later")
+        plumb_guest_nic = True
```
gnuoy: Happy to discuss this but I don't think that this is really plumbing in the interface. It's adding it to operating system config. configure_gateway_ext_port, to my mind, does the plumbing when it attaches the port to the guest.

zaza/openstack/utilities/openstack.py
```diff
@@ -483,8 +484,68 @@ def get_admin_net(neutron_client):
             return net
 
 
+def _plumb_guest_nic(server_name, mac_address, dvr_mode=None):
```
gnuoy: As above, not sure that plumb is the right word here.

zaza/openstack/utilities/openstack.py
```diff
@@ -483,8 +484,68 @@ def get_admin_net(neutron_client):
...
+        application_name = 'neutron-openvswitch'
+    else:
+        application_name = 'neutron-gateway'
+
+    unit_name = juju_utils.get_unit_name_from_host_name(
+        server_name, application_name)
+
+    run_cmd_nic = "ip addr|grep {} -B1".format(mac_address)
+    interface = model.run_on_unit(unit_name, run_cmd_nic)
+    interface = interface['Stdout'].split(' ')[1].replace(':', '')
```
gnuoy: I think you can simplify this a lot by using ```ip -f link -br -o addr |grep {}``` It will only return one line and the interface name won't have a ':' appended.

zaza/openstack/utilities/openstack.py
```diff
@@ -483,8 +484,68 @@ def get_admin_net(neutron_client):
...
+
+    run_cmd_netplan = """sudo egrep -iR '{}|{}' /etc/netplan/
+                        """.format(mac_address, interface)
```
gnuoy: I think this needs tightening up, 'eth12' would match a search for 'eth1' wouldn't it ?

zaza/openstack/utilities/openstack.py
```diff
@@ -483,8 +484,68 @@ def get_admin_net(neutron_client):
...
+                        macaddress: {1}
+                    set-name: {0}
+            version: 2
+    """.format(interface, mac_address))
+    logging.debug("plumb guest interface debug info:")
+    logging.debug("body_value: {}\nunit_name: {}\ninterface: {}\nmac_address:"
+                  "{}\nserver_name: {}".format(body_value, unit_name,
+                                               interface, mac_address,
+                                               server_name))
+    netplan_file = open("60-dataport.yaml", "w")
```
gnuoy: I think its worth using a context manager to write a temporary file here, then it looks after creating it safely and the cleanup.

PR #68, opened 2019-09-29
Porting ceilometer-agent tests to Zaza

ed1000:
Porting the Amulet tests from ceilometer-agent to the Zaza framework.

https://opendev.org/openstack/charm-ceilometer-agent/src/branch/master/tests/basic_deployment.py

### DIFF
--- a/zaza/openstack/charm_tests/ceilometer_agent/setup.py
+++ b/zaza/openstack/charm_tests/ceilometer_agent/setup.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+
+# Copyright 2019 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


zaza/openstack/charm_tests/ceilometer_agent/setup.py
```diff
@@ -0,0 +1,45 @@
...
+        return
+
+    logging.debug('Checking ceilometer-upgrade')
+
+    action = zaza_model.run_action_on_leader(
+        'ceilometer',
+        'ceilometer-upgrade',
+        raise_on_failure=True)
+
+    return action
```
gnuoy: This whole file seems to be a duplicate of zaza/openstack/charm_tests/ceilometer/setup.py so it can be removed. If a test needs ceilometer setup it should just reference the existing function

zaza/openstack/charm_tests/ceilometer_agent/tests.py
```diff
@@ -0,0 +1,116 @@
...
+            alternate_entry,
+            self.services)
+
+    def test_901_pause_resume(self):
+        """Run pause and resume tests.
+
+        Pause service and check services are stopped then resume and check
+        they are started.
+        """
+        self.pause_resume(self.services)
```
gnuoy: This will not run the tests. pause_resume is a contextmanager, calling it needs be called as such "with self.pause_resume(...)

PR #70, opened 2019-09-30
Porting Cinder Backup tests to Zaza

natalytvinova:
Porting the Amulet tests from Cinder Backup to the Zaza framework.

https://opendev.org/openstack/charm-cinder-backup/src/commit/4273738b94b22cca187487af2bad982be49fdd69/tests/basic_deployment.py


zaza/openstack/charm_tests/cinder_backup/tests.py
```diff
@@ -0,0 +1,182 @@
...
+        openstack_utils.resource_reaches_status(
+            self.cinder_client.backups,
+            vol_backup.id,
+            expected_status='available',
+            msg='Backup status wait')
+        # Delete the backup
+        openstack_utils.delete_volume_backup(self.cinder_client, vol_backup.id)
+
+        # Re-check ceph cinder pool object count and disk usage
+        time.sleep(10)
```
gnuoy: Rather than sleeping, can you use resource_removed to wait for the resource to be removed ? There's an example of it being used in the [cinder tests](https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/charm_tests/cinder/tests.py#L135)

zaza/openstack/charm_tests/cinder_backup/tests.py
```diff
@@ -0,0 +1,182 @@
...
+
+        self.assertTrue(cinder_vols)
+
+        cinder_vol = cinder_vols[0]
+
+        # Delete restored cinder volume
+        openstack_utils.delete_volume(self.cinder_client, cinder_vol.id)
+
+        # Final check, ceph cinder pool object count and disk usage
+        time.sleep(10)
```
gnuoy: Same comment as before, look into replacing the sleep with a call to resource_removed

zaza/openstack/charm_tests/cinder_backup/tests.py
```diff
@@ -0,0 +1,182 @@
...
+            msg="Volume status wait")
+
+    def test_410_cinder_vol_create_backup_delete_restore_pool_inspect(self):
+        """Create, backup, delete, restore a ceph-backed cinder volume.
+
+        Create, backup, delete, restore a ceph-backed cinder volume, and
+        inspect ceph cinder pool object count as the volume is created
+        and deleted.
+        """
+        unit_name = 'ceph-mon/0'
```
gnuoy: This assumption about the unit name may be wrong. I think it'd be safer to use `zaza.model.get_lead_unit_name`

natalytvinova: Will `zaza.model.get_lead_unit_name('ceph-mon')` work?

gnuoy: Seems to:
```
>>> import zaza.model
>>> print(zaza.model.get_lead_unit_name('memcached'))
memcached/0
>>> print(type(zaza.model.get_lead_unit_name('memcached')))
<class 'str'>
```

natalytvinova: Great, tranks!

PR #85, opened 2019-10-08
Disabling service check pause_resume

ed1000:
These services are being disabled because of bug #1846390.
https://bugs.launchpad.net/charms/+source/ceilometer/+bug/1846390

### DIFF
--- a/zaza/openstack/charm_tests/ceilometer/tests.py
+++ b/zaza/openstack/charm_tests/ceilometer/tests.py
@@ -40,28 +40,33 @@ class CeilometerTest(test_utils.OpenStackBaseTest):
 
     @property
     def services(self):
-        """Return a list services for Openstack Release."""
+        """Return a list services for Openstack Release.
+
+        Note: disabling ceilometer-polling and ceilometer-agent-central due to
+        bug #1846390.
+        https://bugs.launchpad.net/charms/+source/ceilometer/+bug/1846390
+        """
         current_release = openstack_utils.get_os_release()
         services = []
 
         if current_release >= CeilometerTest.XENIAL_PIKE:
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/ceilometer/tests.py
```diff
@@ -40,41 +40,46 @@ def setUpClass(cls):
...
-            if current_release >= CeilometerTest.TRUSTY_LIBERTY:
+            # if current_release >= CeilometerTest.TRUSTY_LIBERTY:
                 # Liberty and later
-                services.append('ceilometer-polling')
-            else:
+                # services.append('ceilometer-polling')
+            # else:
                 # Juno and earlier
-                services.append('ceilometer-agent-central')
+                # services.append('ceilometer-agent-central')
```
coreycb: Hi Eduardo, In what release combination are you seeing ceilometer-agent-central end up in this list of services? I ask because trusty-mitaka should be the earliest release we're running tests for and based on the logic above ceilometer-agent-central shouldn't end up in the list for trusty-mitaka or above.

ed1000: I'm seeing that from trusty-mitaka onwards. I could change that part

coreycb: It seems like something is broken in this test code then. I'd suggest some debugging to see why ceilometer-agent-central is in the list of services for trusty-mitaka onward.

freyes:
this issue was taken care of by PRs https://github.com/openstack-charmers/zaza-openstack-tests/pull/80/files and https://github.com/openstack-charmers/zaza-openstack-tests/pull/163

PR #101, opened 2019-10-18
Handle OSCI environment variables

thedac:
Move to preferred TEST_* environment variables, while remaining backward
compatible.

### DIFF
--- a/zaza/openstack/utilities/generic.py
+++ b/zaza/openstack/utilities/generic.py
@@ -139,21 +139,35 @@ def get_undercloud_env_vars():
     :returns: Network environment variables
     :rtype: dict
     """
-    # Handle backward compatibile OSCI enviornment variables
+    # Handle OSCI environment variables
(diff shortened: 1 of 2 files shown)


zaza/openstack/utilities/generic.py
```diff
@@ -139,21 +139,35 @@ def get_undercloud_env_vars():
     :returns: Network environment variables
     :rtype: dict
     """
-    # Handle backward compatibile OSCI enviornment variables
+    # Handle OSCI environment variables
+    # Over time we have changed the names of prefixes. Try them all in
+    # ascending order of significance.
+    # Note: TEST_* is the preferred prefix
+    _prefixes = ["AMULET_", "MOJO_", "OS_", "", "TEST_"]
```
ryan-beisner: I'd rather not introduce a NONE prefix.  I'd rather make this reflect the expected realty of passenv, which is:

```
passenv = HOME TERM CS_* OS_* TEST_*
```

PR #105, opened 2019-10-25
Add heat functional tests

jose-delarosa:
This Pull Request adds ported Amulet functional tests for the heat charm.

https://bugs.launchpad.net/charm-heat/+bug/1828424


zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+        logging.info('Deleting keypair...')
+        openstack_utils.delete_resource(self.nova_client.keypairs,
+                                        KEYPAIR_NAME, msg="nova keypair")
+
+    def test_100_domain_setup(self):
+        """Run required action for a working Heat unit."""
+        # Action is REQUIRED to run for a functioning heat deployment
+        logging.info('Running domain-setup action on heat unit...')
+        unit = zaza.model.get_units(self.application_name)[0]
+        assert unit.workload_status == "active"
```
ajkavanagh: Doesn't this assume that the 100 test is the first one (after deployment).  Better to wait for active before proceeding and let the timeout kill it if it never becomes active.

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+        zaza.model.run_action(unit.entity_id, "domain-setup")
+        zaza.model.block_until_unit_wl_status(unit.entity_id, "active")
+        unit = zaza.model.get_unit_from_name(unit.entity_id)
+        assert unit.workload_status == "active"
```
ajkavanagh: This assert isn't necessary.  At 230 the test just waited for the unit to become active.

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+
+    def test_400_heat_resource_types_list(self):
+        """Check default resource list behavior and confirm functionality."""
+        logging.info('Checking default heat resource list...')
+        try:
+            types = list(self.heat_client.resource_types.list())
+            if type(types) is list:
```
ajkavanagh: How could `type(types)` not be a `list`?  It was just set to one the line above?

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+                msg = 'Resource type list is not a list!'
+                logging.error('{}'.format(msg))
+                raise
+            if len(types) > 0:
+                logging.info('Resource type list is populated '
+                             '({}, ok).'.format(len(types)))
+            else:
+                msg = 'Resource type list length is zero!'
+                logging.error(msg)
+                raise
```
ajkavanagh: This should be an assert; it's in a test.

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+        """Check default resource list behavior and confirm functionality."""
+        logging.info('Checking default heat resource list...')
+        try:
+            types = list(self.heat_client.resource_types.list())
+            if type(types) is list:
+                logging.info('Resource type list check is ok.')
+            else:
+                msg = 'Resource type list is not a list!'
+                logging.error('{}'.format(msg))
+                raise
```
ajkavanagh: This should be an assert, it's in a test.  (note path can never be reached as the type of types is always a `list`).

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+                logging.info('Resource type list is populated '
+                             '({}, ok).'.format(len(types)))
+            else:
+                msg = 'Resource type list length is zero!'
+                logging.error(msg)
+                raise
+        except Exception as e:
+            msg = 'Resource type list failed: {}'.format(e)
+            logging.error(msg)
+            raise
```
ajkavanagh: This should be an assert.

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+
+    def test_402_heat_stack_list(self):
+        """Check default heat stack list behavior, confirm functionality."""
+        logging.info('Checking default heat stack list...')
+        try:
+            stacks = list(self.heat_client.stacks.list())
+            if type(stacks) is list:
```
ajkavanagh: Again, `stacks` is always a list if the code gets this far.

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+                logging.info("Stack list check is ok.")
+            else:
+                msg = 'Stack list returned something other than a list.'
+                logging.error(msg)
+                raise
```
ajkavanagh: Again, an assert here, not a raise.

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+        except Exception as e:
+            msg = 'Heat stack list failed: {}'.format(e)
+            logging.error(msg)
+            raise
```
ajkavanagh: And here.

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+    def test_410_heat_stack_create_delete(self):
+        """Create stack, confirm nova compute resource, delete stack."""
+        logging.info('Creating, deleting heat stack (compute)...')
+        self._image_create()
+        self._keypair_create()
+        self._stack_create()
+        self._stack_resource_compute()
+        self._stack_delete()
+        self._image_delete()
+        self._keypair_delete()
```
ajkavanagh: Are any of these helper functions used outside of this test?  If not, then they should either be placed after the test, or just inlined into the test, please.

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+                   .format("{}={}".format(k, v) for k, v in keys.items()))
+            logging.error("Error: {}".format(msg))
+
+    @staticmethod
+    def _run_arbitrary(command, timeout=300):
+        """Run an arbitrary command (as root), but not necessarily on a unit.
+
+        (Otherwise the self.run(...) command could have been used for the unit
+
+        :param str command: The command to run.
```
ajkavanagh: Please could we use the standard form for these.  i.e.

    :param command: The command to run
    :type command: str

Thanks.

Also, this function may raise subprocess.CalledProcessError and so also needs:

    :raises: subprocess.CalledProcessError.

Finally, the return type:

    :returns: A pair containing the output of the command and exit value
    :rtype: (str, int)

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+        """Test the auth_encryption_key in heat.conf is same on all units."""
+        logging.info("Checking the 'auth_encryption_key' is the same on "
+                     "all units.")
+        output, ret = self._run_arbitrary(
+            "--application heat "
+            "--format json "
+            "grep auth_encryption_key /etc/heat/heat.conf")
+        if ret:
+            msg = "juju run returned error: ({}) -> {}".format(ret, output)
+            logging.error("Error: {}".format(msg))
```
ajkavanagh: This should bail at this point:  e.g.

    self.assertEqual(ret, 0, "juju run returned error: ret: {}. output: {}".format(ret, msg))

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+        output = json.loads(output)
+        keys = {}
+        for r in output:
+            k = r['Stdout'].split('=')[1].strip()
+            keys[r['UnitId']] = k
+        # see if keys are different.
+        ks = list(keys.values())
+        if any(((k != ks[0]) for k in ks[1:])):
```
ajkavanagh: A quicker/easier way to do this is to use a set:

    ks = set(keys.values())
    self.assertEqual(len(ks), 1, "'auth_enc....: {}".format(", ".join(ks)))

thedac:
Just to be official, adding that I have requested the yaml files be moved to the heat charm rather than having them in the zaza python module.
Example: Keystone SAML Mellon does this with samltest.xml

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,355 @@
...
+            msg = 'Failed to get heat stack: {}'.format(e)
+            logging.error(msg)
+
+        # Confirm stack name.
+        logging.info('Expected, actual stack name: {}, '
+                     '{}'.format(STACK_NAME, _stack.stack_name))
+        if STACK_NAME != _stack.stack_name:
+            msg = 'Stack name mismatch, {} != {}'.format(STACK_NAME,
+                                                         _stack.stack_name)
+            logging.error(msg)
```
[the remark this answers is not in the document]
jose-delarosa: Does an assert bail out of the method or out of the entire functional test run (does the program exit?) I wasn't sure if we need to completely quit, or just mark the specific test as failed. Will switch to using `assert False, msg` on all of these.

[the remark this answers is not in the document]
ajkavanagh: Yes, the assert will bail out of the test as a test failure.  Technically, the `assert` keyword raises the `AssertionError` behind the scenes, which is what the test framework is looking for.

zaza/openstack/charm_tests/heat/tests.py
jose-delarosa: Thanks @ajkavanagh. Admittedly, I copied & pasted this method from the Amulet implementation as-is.

zaza/openstack/charm_tests/heat/tests.py
jose-delarosa: @ajkavanagh I don't believe these functions are used anywhere else. However, I don't quite follow where they should be placed instead. You mean put them in a separate class/file like utils.py? Or just don't have separate functions and put all the code in the 410 function?

zaza/openstack/charm_tests/heat/tests.py
jose-delarosa: @ajkavanagh Are you suggesting a change here? I did remove the last 2 lines in this function per your next suggestion, but not sure what else needs to be done here.

jose-delarosa:
@thedac Acknowledged. Have already implemented change, expect it in next commit.

zaza/openstack/charm_tests/heat/tests.py
jose-delarosa: @ajkavanagh Trying to understand what the right code should be here. If it's OK with you, I will treat this as Optional for now (or at least put it at the bottom of my list).

Thank you.

zaza/openstack/charm_tests/heat/tests.py
jose-delarosa: @ajkavanagh Upon closer inspection, I am going to remove the 402() function since the heat stack is empty at this point, and I do check the stack list in _stack_create() after a stack is created.

zaza/openstack/charm_tests/heat/tests.py
ajkavanagh: The assert assumes that the workload status is 'active'  -- this assumes that this test immediately follows deployment; another test may have, say, changed config back and the workload may not have reached active before this test runs.  If you want to ensure that it is at the active state, then use:

    zaza.model.block_until_unit_wl_status(unit.entity_id, "active")

This is the code from 2 lines below this line.

zaza/openstack/charm_tests/heat/tests.py
ajkavanagh: I suspect this is now wrong; `self.heat_client.resource_types.list()` may return a generator, not a list, which was why it was forced to a list.  You need to check the return type for that function and see what it can return to determine what the code below should be.

zaza/openstack/charm_tests/heat/tests.py
ajkavanagh: The 'flow' in the file is awkward as the functions get defined before the are used; I'm questioning why they are functions at all; i.e. why aren't they just inline code in the test?  What value is gained from them being made into functions?

zaza/openstack/charm_tests/heat/tests.py
ajkavanagh: Okay, which is easier to understand:

```python
ks = list(keys.values())
if any(((k != ks[0]) for k in ks[1:])):
```

or

```python
ks = set(keys.values())
if len(ks) != 1:
```

because they do exactly the same thing.   It's just that a `set()` automatically only has unique members.  i.e. if the keys are different then there will be more than 1 key in the set.

jose-delarosa: Thanks @ajkavanagh, this is much simpler. Full disclaimer: I was using the same code that was in the original Amulet testing.

zaza/openstack/charm_tests/heat/tests.py
jose-delarosa: Got it, makes sense, thank you.

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,348 @@
...
+    def services(self):
+        """Return a list services for OpenStack release.
+
+        :returns: List of services
+        :rtype: [str]
+        """
+        services = ['heat-api', 'heat-api-cfn', 'heat-engine']
+        return services
+
+    def _image_create(self):
```
ryan-beisner: It feels like you're re-creating something that already exists in zaza-openstack-tests:

https://github.com/openstack-charmers/zaza-openstack-tests/blob/0c2f36fb4e7313e78b6982a594bbf595bc263c8e/zaza/openstack/charm_tests/glance/setup.py#L65-L77

jose-delarosa: @ryan-beisner I understand the desire to use existing libraries/helper routines, but I thought that's what `openstack_utils` was for. I assumed that every set of tests in `zaza/openstack/charm_tests` was independent of each other, but now heat will be dependent on the glance setup test routines. Anyway, I will do what is best long-term, I will reuse helper functions from glance and nova as suggested.

jose-delarosa: One more thing, I am also concerned about the functional test failure in the proposed tests, but I am using the exact same `openstack_utils` functions as the glance helper routine is, so I don't see that changing with the glance helper routine.

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+                assert False, msg
+        except Exception as e:
+            msg = 'Resource type list failed: {}'.format(e)
+            assert False, msg
+
+    def test_410_heat_stack_create_delete(self):
+        """Create stack, confirm nova compute resource, delete stack."""
+        # Create an image for use by the heat template
+        logging.info('Creating glance image ({})...'.format(IMAGE_NAME))
+        glance_setup.add_cirros_image(image_name=IMAGE_NAME)
```
ajkavanagh: Instead of calling the helper here, use a `configure` line in the tests.yaml of the test.  i.e.

```yaml
configure:
  - zaza.charm_tests.glance.setup.add_cirros_image
```

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # Verify new image name
+        images_list = list(self.glance_client.images.list())
+        if images_list[0].name != IMAGE_NAME:
+            message = ('glance image create failed or unexpected '
+                       'image name {}'.format(images_list[0].name))
+            assert False, message
```
ajkavanagh: This whole section should become:

```python
images_list = list(...)
self.assertNotEqual(images_list[0].name, IMAGE_NAME, "glance image ...")
```

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def test_400_heat_resource_types_list(self):
+        """Check default resource list behavior and confirm functionality."""
+        logging.info('Checking default heat resource list...')
+        try:
+            types = self.heat_client.resource_types.list()
+            if type(types) is list:
+                logging.info('Resource type list check is ok.')
+            else:
+                msg = 'Resource type list is not a list!'
```
ajkavanagh: Replace with:

```python
self.assertIsInstance(types, list, "resource type is not a list!")
```

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+                assert False, msg
+            if len(types) > 0:
+                logging.info('Resource type list is populated '
+                             '({}, ok).'.format(len(types)))
+            else:
+                msg = 'Resource type list length is zero!'
+                assert False, msg
```
ajkavanagh: Replace with:

```python
self.assertGreater(len(types), 0, "Resource type list len is zero")
```

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+        except Exception as e:
+            msg = 'Resource type list failed: {}'.format(e)
+            assert False, msg
```
ajkavanagh: Replace with:

```python
self.fail(msg)
```

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+        try:
+            stack = self.heat_client.stacks.create(**fields)
+            logging.info('Stack data: {}'.format(stack))
+            stack_id = stack['stack']['id']
+            logging.info('Creating new stack, ID: {}'.format(stack_id))
+        except Exception as e:
+            # Generally, an api or cloud config error if this is hit.
+            msg = 'Failed to create heat stack: {}'.format(e)
+            logging.error(msg)
+            raise
```
ajkavanagh: Replace with:

```python
self.fail("Failed to ....")
```

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+        stacks = list(self.heat_client.stacks.list())
+        logging.info('All stacks: {}'.format(stacks))
+
+        # Get stack information
+        try:
+            stack = self.heat_client.stacks.get(STACK_NAME)
+        except Exception as e:
+            # Generally, a resource availability issue if this is hit.
+            msg = 'Failed to get heat stack: {}'.format(e)
+            assert False, msg
```
ajkavanagh: Either drop the try/except (and let the exception just fail the test), or replace this bit with:

```python
self.fail('Failed to get ...')
```

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # Confirm stack name.
+        logging.info('Expected, actual stack name: {}, '
+                     '{}'.format(STACK_NAME, stack.stack_name))
+        if STACK_NAME != stack.stack_name:
+            msg = 'Stack name mismatch, {} != {}'.format(STACK_NAME,
+                                                         stack.stack_name)
+            assert False, msg
```
ajkavanagh: Replace with:

```python
self.assertEqual(stack.stack_name, STACK_NAME, "Stack name ...")
```

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+        logging.info('Confirming heat stack resource status...')
+        resource = self.heat_client.resources.get(STACK_NAME, RESOURCE_TYPE)
+        server_id = resource.physical_resource_id
+        if server_id:
+            logging.debug('Heat template spawned nova instance, '
+                          'ID: {}'.format(server_id))
+        else:
+            msg = 'Stack failed to spawn a nova compute resource (instance).'
+            logging.error(msg)
+
```
ajkavanagh: Either an error has occurred and the test should bail out or it hasn't and there is nothing to log?  Which is it?  I suspect this should just be:

```python
self.assertTrue(server_id, "Stack failed to spawn a nova compute ...")
```

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+        # Delete stack
+        logging.info('Deleting heat stack...')
+        openstack_utils.delete_resource(self.heat_client.stacks,
+                                        STACK_NAME, msg="heat stack")
+
+        # Delete image
+        logging.info('Deleting glance image...')
+        image = self.nova_client.glance.find_image(IMAGE_NAME)
+        openstack_utils.delete_resource(self.glance_client.images,
+                                        image.id, msg="glance image")
```
ajkavanagh: No need to delete the image.

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # Delete keypair
+        logging.info('Deleting keypair...')
+        openstack_utils.delete_resource(self.nova_client.keypairs,
+                                        nova_utils.KEYPAIR_NAME,
+                                        msg="nova keypair")
```
ajkavanagh: No need to delete the keypair

zaza/openstack/charm_tests/heat/tests.py
```diff
@@ -0,0 +1,311 @@
...
+        keys = {}
+        for r in output:
+            k = r['Stdout'].split('=')[1].strip()
+            keys[r['UnitId']] = k
+        # see if keys are different
+        ks = set(keys.values())
+        if len(ks) != 1:
+            msg = ("'auth_encryption_key' is not identical on every unit: {}"
+                   .format("{}={}".format(k, v) for k, v in keys.items()))
+            assert False, msg
```
ajkavanagh: Replace with:

```python
ks = set(keys.values())
self.assertEqual(len(ks), 1, "...failure message...")
```

zaza/openstack/charm_tests/heat/tests.py
jose-delarosa: This should be `self.assertEqual()`, it should only fail if they image names are different.

PR #107, opened 2019-11-01
MySQL 8 specific tests

thedac:
* Re-arrange the MySQL, Percona-cluster classes to avoid code
      duplication
* cluster-status action

### DIFF
--- a/zaza/openstack/charm_tests/mysql/tests.py
+++ b/zaza/openstack/charm_tests/mysql/tests.py
@@ -14,6 +14,7 @@
 
 """MySQL/Percona Cluster Testing."""
 
+import json
 import logging
 import os
 import re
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/mysql/tests.py
```diff
@@ -398,3 +400,31 @@ def test_100_kill_crm_master(self):
...
+        cls.application = "mysql-innodb-cluster"
+        cls.services = ["mysqld"]
+
+    def test_801_cluster_status(self):
+        """Checking cluster status.
+
+        Run the cluster-status action.
+        """
+        # Update which node is the leader and which are not
+        self.update_leaders_and_non_leaders()
```
ajkavanagh: This is being run for its side-effect of updating the class vars leader/non-leaders ... it would be "better" if the function just returned `(leader, non_leaders)` as a tuple and test functions just called the helper method directly.  i.e. `cls.leader` is just a global in disguise, and it's being updated, which means that its hard to track the side-effects.

thedac:
Alex, I think I have addressed your concerns.

PR #115, opened 2019-11-07
Add octavia policyd tests

ajkavanagh:
Providing a policy.d test for Octavia is a bit more complicated as the
obvious policy is just to prevent listing of load balancers;
unfortunately it doesn't work in the upstream, as the code doesn't check
the policy in octavia.  However, "showing" an load balancer does involve
a policy check and so the test is more complicated than 'normal' policy
tests in that a resource is created, then the policy override is checked
and then the resource is deleted.

### DIFF
--- a/zaza/openstack/charm_tests/policyd/tests.py
+++ b/zaza/openstack/charm_tests/policyd/tests.py
@@ -35,13 +35,15 @@ import logging
 import os
 import shutil
 import tempfile
+import tenacity
 import unittest
 import zipfile
 
-import keystoneauth1
-import glanceclient.common.exceptions
+from octaviaclient.api.v2 import octavia as octaviaclient
 import cinderclient.exceptions
 import heatclient.exc
+import glanceclient.common.exceptions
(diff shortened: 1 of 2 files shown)


zaza/openstack/charm_tests/policyd/tests.py
```diff
@@ -593,3 +625,109 @@ def get_client_and_attempt_operation(self, ip):
...
+        @tenacity.retry(wait=tenacity.wait_fixed(1),
+                        reraise=True, stop=tenacity.stop_after_delay(900))
+        def wait_for_lb_resource(client, resource_id):
+            resp = client.load_balancer_show(resource_id)
+            logging.info(resp['provisioning_status'])
+            assert resp['provisioning_status'] == 'ACTIVE', (
+                'load balancer resource has not reached '
+                'expected provisioning status: {}'
+                .format(resp))
+            return resp
```
fnordahl: future todo: We should move this to a central and re-usable location

PR #117, opened 2019-11-11
Increase number of attempts for ping_response

fnordahl:
Some configurations require more time to settle before we get the
expected response. (LP: #1851710)


fnordahl:
Log of successful run of the job referenced in [bug 1851710](https://bugs.launchpad.net/charm-neutron-api/+bug/1851710)

PR #118, opened 2019-11-12
net: Add support for DVR deployments with NGW

fnordahl:


fnordahl:
Output of `bionic-train-dvr` job run with this PR in place at [0].

0: https://gist.github.com/fnordahl/7ec7b858254eb9a429fef521cdea0e17

PR #125, opened 2019-11-15
Add manila API test

ChrisMacNaughton:


zaza/openstack/charm_tests/manila/tests.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class ManilaTests(test_utils.OpenStackBaseTest):
+    """Encapsulate Manila  tests."""
+
+    RESOURCE_PREFIX = 'zaza-manilatests'
+    INSTANCE_USERDATA = """#cloud-config
+packages:
+- nfs-common
+"""
```
ajkavanagh: Formatting looks a bit strange?

zaza/openstack/charm_tests/manila/tests.py
```diff
@@ -0,0 +1,48 @@
...
+        super(ManilaTests, cls).setUpClass()
+        cls.nova_client = (
+            openstack_utils.get_nova_session_client(cls.keystone_session))
+        cls.manila_client = manilaclient.Client(
+            session=cls.keystone_session, client_version='2')
+
+    def test_manila_api(self):
+        """Test that the Manila API is working."""
+        # now just try a list the shares
+        # NOTE(AJK) the 'search_opts={}' is to work around Bug#1707303
```
ajkavanagh: This is fix released in 2017-09-13; is it still relevant?

zaza/openstack/charm_tests/manila/tests.py
ChrisMacNaughton: it's a multiline cloud-config string - I think that the #cloud-config line has to be on the first line of the user-data, and then, indentation matters, so `packages:` has to be on the left-most column

zaza/openstack/charm_tests/manila/tests.py
ChrisMacNaughton: it looks like it's backported to pike - given that we seem to just grab whatever version is published for the clients, I bet it'll work :) will remove it

PR #135, opened 2019-11-28
Guard against run_resource_cleanup being unset

gnuoy:
A test may not call the base class setup so may be missing
run_resource_cleanup in which case teardown will fail. Guard against
that and default to False.

### DIFF
--- a/zaza/openstack/charm_tests/test_utils.py
+++ b/zaza/openstack/charm_tests/test_utils.py
@@ -93,6 +93,8 @@ def audit_assertions(action,
 class OpenStackBaseTest(unittest.TestCase):
     """Generic helpers for testing OpenStack API charms."""
 
+    run_resource_cleanup = False
+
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/test_utils.py
```diff
@@ -107,7 +107,7 @@ def resource_cleanup(cls):
     @classmethod
     def tearDown(cls):
         """Run teardown for test class."""
-        if cls.run_resource_cleanup:
+        if getattr(cls, 'run_resource_cleanup', False):
```
sahid: I probably missed something, why not to declare it as a class attribute? That would avoid using a check each time we want to use it.

```
class OpenStackBaseTest(unittest.TestCase):
    run_resource_cleanup = False
    ...
```

sahid:
+1

PR #141, opened 2019-12-06
py3.7+ StopIteration handling

thedac:
In python 3.7 StopIteration is no longer an acceptable way to end generation pre-maturely. RuntimeError is now thrown. In other circumstances, StopIteration can be thrown when not expected.

This change re-arranges some code that relied on the old behavior. It should be backward and forward compatible.

[0] https://www.python.org/dev/peps/pep-0479/

### DIFF
--- a/unit_tests/utilities/test_zaza_utilities_openstack.py
+++ b/unit_tests/utilities/test_zaza_utilities_openstack.py
@@ -816,7 +816,7 @@ class TestOpenStackUtils(ut_utils.BaseTestCase):
         _machine = mock.MagicMock()
 
         # No machine returned
-        self._get_machines.side_effect = KeyError
+        self._get_machines.return_value = []
         with self.assertRaises(exceptions.ApplicationNotFound):
             openstack_utils.get_current_os_release_pair()
         self._get_machines.side_effect = None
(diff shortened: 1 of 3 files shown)


zaza/openstack/utilities/openstack.py
```diff
@@ -674,31 +674,35 @@ def configure_gateway_ext_port(novaclient, neutronclient, net_id=None,
...
-                                         mac_address=mac_address)
+                port = neutronclient.create_port(body=body_value)
+                server.interface_attach(port_id=port['port']['id'],
+                                        net_id=None, fixed_ip=None)
+                if add_dataport_to_netplan:
+                    mac_address = get_mac_from_port(port, neutronclient)
+                    add_interface_to_netplan(server.name,
+                                             mac_address=mac_address)
+    except StopIteration:
+        pass
```
ajkavanagh: Not sure what this `StopIteration` refers to?

unit_tests/utilities/test_zaza_utilities_openstack.py
```diff
@@ -814,14 +813,16 @@ def test_get_current_openstack_release_pair(self):
             name='_get_machine_series'
         )
 
+        _machine = mock.MagicMock()
+
         # No machine returned
-        self._get_machines.side_effect = StopIteration
+        self._get_machines.side_effect = KeyError
```
ajkavanagh: This is not really testing the thing you want to test?  I think it would be better to say:

```python
self._get_machines.return_value = []
```

And that means that the test is testing that the target code will handle the empty list, rather than make the code implementation dependent on the test raising `KeyError`?

PR #142, opened 2019-12-10
Neutron Gateway zaza tests

jguedez:
Port functional tests from the Amulet to the Zaza framework - https://bugs.launchpad.net/charm-neutron-gateway/+bug/1828424

- Dropped some tests, to bring testing for this charm in line with the current functional testing strategy for OpenStack charms. In particular, the tests from the following series were impacted 1XX (Keystone catalog), 2XX (relations), and 3XX (config files).

- Additional functional tests will be included in the neutron-gateway - **NeutronNetworkingTest** (originally part of the neutron-api functional tests). Those will be invoked via the charm itself, and have been validated already (no changes needed in this repository for that).

### DIFF
--- a/zaza/openstack/charm_tests/neutron/tests.py
+++ b/zaza/openstack/charm_tests/neutron/tests.py
@@ -17,7 +17,6 @@
 """Encapsulating `neutron-openvswitch` testing."""
 
 
-import time
 import logging
 import tenacity
 import unittest
@@ -90,21 +89,7 @@ class NeutronGatewayTest(test_utils.OpenStackBaseTest):
                     {'enable-qos': 'True'},
                     application_name="neutron-api"):
 
-                # wait for neutron-gateway to complete as well
-                time.sleep(60)
-
-                # obtain dhcp agent to identify the neutron-gateway host
-                dhcp_agent = self.neutron_client.list_agents(
-                    binary='neutron-dhcp-agent')['agents'][0]
-                neutron_gw_host = dhcp_agent['host']
-                logging.debug('neutron gw host: {}'.format(neutron_gw_host))
-
-                # check extensions on the ovs agent to validate qos
-                ovs_agent = self.neutron_client.list_agents(
-                    binary='neutron-openvswitch-agent',
-                    host=neutron_gw_host)['agents'][0]
-
-                self.assertIn('qos', ovs_agent['configurations']['extensions'])
+                self._validate_openvswitch_agent_qos()
 
     def test_900_restart_on_config_change(self):
         """Checking restart happens on config change.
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/neutron/tests.py
```diff
@@ -29,6 +30,206 @@
...
+            services.append('neutron-vpn-agent')
+        if cls.current_os_release < xenial_newton:
+            services.append('neutron-lbaas-agent')
+        if xenial_newton <= cls.current_os_release < bionic_train:
+            services.append('neutron-lbaasv2-agent')
+
+        return services
+
+
+class NeutronGatewaySecurityTest(test_utils.OpenStackBaseTest):
```
thedac: Would it be possible to generalize this class with the existing SecurityTest class. Just some intelligence about the application name seems necessary.

zaza/openstack/charm_tests/neutron/tests.py
```diff
@@ -29,6 +30,206 @@
...
+        # set up clients
+        cls.neutron_client = (
+            openstack_utils.get_neutron_session_client(cls.keystone_session))
+
+        bionic_stein = openstack_utils.get_os_release('bionic_stein')
+
+        cls.pgrep_full = (True if cls.current_os_release >= bionic_stein
+                          else False)
+
+    def test_400_create_network(self):
```
thedac: Not a blocker, but the 4xx tests feel much more about testing the neutron-api than the gateway. Ironically NeutronNetworkingTest is much more about the gateway. It might be good to add these to the NeutronAPITest class rather than this one.

zaza/openstack/charm_tests/neutron/tests.py
```diff
@@ -29,6 +30,206 @@
...
+                openstack_utils.get_os_release('trusty_mitaka')):
+            logging.info('running qos check')
+
+            with self.config_change(
+                    {'enable-qos': 'False'},
+                    {'enable-qos': 'True'},
+                    application_name="neutron-api"):
+
+                # wait for neutron-gateway to complete as well
+                time.sleep(60)
```
thedac: The time.sleep may have been in amulet. But we have a standard for zaza that we never use these. See tenacity in various other tests as a way of waiting or retrying.

zaza/openstack/charm_tests/neutron/tests.py
jguedez: done - [commit](https://github.com/openstack-charmers/zaza-openstack-tests/pull/142/commits/79bb61243f25d3264ab81edfd9279036d930dd5d)

I considered doing this initially, but hesitated to touch another charm at the same time. Func-smoke is happy for both neutron-api and neutron-gateway with this commit (should essentially be a noop for neutron-api)

zaza/openstack/charm_tests/neutron/tests.py
jguedez: done - [commit](https://github.com/openstack-charmers/zaza-openstack-tests/pull/142/commits/f79f2fb54c062651324f7140e9f90941c565c302)

Validated via func-smoke. Tenacity is nice for sure :)

zaza/openstack/charm_tests/neutron/tests.py
jguedez: done - [commit](https://github.com/openstack-charmers/zaza-openstack-tests/pull/142/commits/55fc718f981f22cb7a555de993960841d8d5196a)

Validated via func-smoke on both neutron-api and neutron-gateway

PR #151, opened 2020-01-10
Add a short exponential retry to the mount setup bits

ChrisMacNaughton:


ChrisMacNaughton:
I have completed a test run using this branch on the manila-ganesha charm, and it works as expected, resolving the transient test failure that's been observed in the gate

PR #155, opened 2020-01-17
Fix for NDR testing race condition

thedac:
Previous attempts to fix LP Bug#1784083 added a workaround (commit
820ed808) which is being removed here.

The root cause seems to be upstream in the dragent. It may never have
been envisioned to run the agent by itself the way the charm does.

So that even if neutron-api completes its amqp relation first,
neutron-dynamic-routing can still see
oslo_messaging.exceptions.MessagingTimeout errors. Some operation must
occur against neutron before dragent is truly ready. i.e. some post
deploy openstack command. So it is outside the purview of the charm.

This change adds a service restart late.

Partial-Bug: #1841459

### DIFF
--- a/zaza/openstack/utilities/generic.py
+++ b/zaza/openstack/utilities/generic.py
@@ -824,3 +824,29 @@ def get_series(unit):
     result = model.run_on_unit(unit.entity_id,
                                "lsb_release -cs")
     return result['Stdout'].strip()
+
+
+def systemctl(unit, service, command="restart"):
+    """Run systemctl command on a unit.
+
+    :param unit: Unit object or unit name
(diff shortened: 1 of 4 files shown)


zaza/openstack/utilities/generic.py
```diff
@@ -824,3 +824,29 @@ def get_series(unit):
     result = model.run_on_unit(unit.entity_id,
                                "lsb_release -cs")
     return result['Stdout'].strip()
+
+
+def systemctl(unit, service, command="restart"):
+    """Run systemctl command on a unit.
```
addyess: :+1:

PR #156, opened 2020-01-21
Bug/1828424

addyess:
Port keystone-ldap [basic tests](https://opendev.org/openstack/charm-keystone-ldap/src/branch/master/src/tests/basic_deployment.py) to zaza

### DIFF
--- a/zaza/openstack/charm_tests/keystone/tests.py
+++ b/zaza/openstack/charm_tests/keystone/tests.py
@@ -405,7 +405,10 @@ class LdapTests(BaseKeystoneTest):
         :return: return username if found
         :rtype: Optional[str]
         """
-        client = self.admin_keystone_client
+        client = openstack_utils.get_keystone_session_client(
+            self.admin_keystone_session,
+            client_api_version=3)
+
         domain_users = client.users.list(
             domain=client.domains.find(name=domain).id
         )
--- a/zaza/openstack/charm_tests/keystone/tests.py
+++ b/zaza/openstack/charm_tests/keystone/tests.py
@@ -17,7 +17,7 @@ import collections
 import json
 import logging
 import pprint
-import tenacity
+import unittest
 import keystoneauth1
 
 import zaza.model
@@ -145,11 +145,11 @@ class CharmOperationTest(BaseKeystoneTest):
                 if unit_repo != lead_repo:
                     raise zaza_exceptions.KeystoneKeyRepositoryError(
                         'expect: "{}" actual({}): "{}"'
-                        ''.format(pprint.pformat(lead_repo), unit.entity_id,
-                                  pprint.pformat(unit_repo)))
+                        .format(pprint.pformat(lead_repo), unit.entity_id,
+                                pprint.pformat(unit_repo)))
                 logging.info('"{}" == "{}"'
-                             ''.format(pprint.pformat(unit_repo),
-                                       pprint.pformat(lead_repo)))
+                             .format(pprint.pformat(unit_repo),
+                                     pprint.pformat(lead_repo)))
 
 
 class AuthenticationAuthorizationTest(BaseKeystoneTest):
@@ -258,8 +258,7 @@ class AuthenticationAuthorizationTest(BaseKeystoneTest):
         logging.info('OK')
 
     def test_end_user_access_and_token(self):
-        """
-        Verify regular end-user access resources and validate token data.
+        """Verify regular end-user access resources and validate token data.
 
         In effect this also validates user creation, presence of standard
         roles (`_member_`, `Member`), effect of policy and configuration
@@ -280,7 +279,7 @@ class AuthenticationAuthorizationTest(BaseKeystoneTest):
                 if len(token) != 32:
                     raise zaza_exceptions.KeystoneWrongTokenProvider(
                         'We expected a UUID token and got this: "{}"'
-                        ''.format(token))
+                        .format(token))
             else:
                 if len(token) < 180:
                     raise zaza_exceptions.KeystoneWrongTokenProvider(
@@ -382,36 +381,19 @@ class LdapTests(BaseKeystoneTest):
         """Run class setup for running Keystone ldap-tests."""
         super(LdapTests, cls).setUpClass()
 
-    def _get_ldap_config(self):
+    @staticmethod
+    def _get_ldap_config():
         ldap_ips = zaza.model.get_app_ips("ldap-server")
-        ldap_server_ip = ldap_ips[0]
-
-        keystone_ldap_config = {
-            'ldap-server': "ldap://{}".format(ldap_server_ip),
-            'ldap-user': 'cn=admin,dc=test,dc=com',
-            'ldap-password': 'crapper',
-            'ldap-suffix': 'dc=test,dc=com',
-            'domain-name': 'userdomain',
-        }
-        if all(keystone_ldap_config.values()):
-            self.ldap_configured = True
-            return keystone_ldap_config
-        else:
-            # NOTE(jamespage): Use mock values to check deployment only
-            #                  as no test fixture has been supplied
-            self.ldap_configured = False
-            return {
-                'ldap-server': 'myserver',
-                'ldap-user': 'myuser',
-                'ldap-password': 'mypassword',
-                'ldap-suffix': 'mysuffix',
+        if ldap_ips:
+            return True, {
+                'ldap-server': "ldap://{}".format(ldap_ips[0]),
+                'ldap-user': 'cn=admin,dc=test,dc=com',
+                'ldap-password': 'crapper',
+                'ldap-suffix': 'dc=test,dc=com',
                 'domain-name': 'userdomain',
             }
+        return False, {}
 
-    # NOTE: intermittent authentication fails. Wrap in a retry loop.
-    @tenacity.retry(wait=tenacity.wait_exponential(multiplier=1,
-                                                   min=5, max=10),
-                    reraise=True)
     def _find_keystone_v3_user(self, username, domain):
         """Find a user within a specified keystone v3 domain."""
         client = self.admin_keystone_client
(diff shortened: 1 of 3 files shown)


zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -371,3 +372,87 @@ def test_security_checklist(self):
...
+            'keystone-ldap': {
+                'workload-status': 'idle',
+                'workload-status-messages': 'Unit is ready'
+            },
+            'keystone': {
+                'workload-status': 'idle',
+                'workload-status-messages': 'Unit is ready'
+            }
+        }
+        zaza.model.wait_for_application_states(states=states)
```
addyess: this feels awkward.  What's the "RIGHT" way to wait for a config change to flutter around and make things happy?"

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -371,3 +372,87 @@ def test_security_checklist(self):
...
+            self.ldap_configured = False
+            return {
+                'ldap-server': 'myserver',
+                'ldap-user': 'myuser',
+                'ldap-password': 'mypassword',
+                'ldap-suffix': 'mysuffix',
+                'domain-name': 'userdomain',
+            }
+
+    # NOTE: intermittent authentication fails. Wrap in a retry loop.
```
addyess: this helped in cases where the user may have not yet been accessible from keystone?  Maybe there's a better way here?

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -145,11 +145,11 @@ def test_key_distribution_and_rotation(self):
                 if unit_repo != lead_repo:
                     raise zaza_exceptions.KeystoneKeyRepositoryError(
                         'expect: "{}" actual({}): "{}"'
-                        .format(pprint.pformat(lead_repo), unit.entity_id,
-                                pprint.pformat(unit_repo)))
+                        ''.format(pprint.pformat(lead_repo), unit.entity_id,
```
ajkavanagh: You don't need the '' before the `.format(...)` statement.

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -279,12 +280,12 @@ def _validate_token_data(openrc):
                 if len(token) != 32:
                     raise zaza_exceptions.KeystoneWrongTokenProvider(
                         'We expected a UUID token and got this: "{}"'
-                        .format(token))
+                        ''.format(token))
```
ajkavanagh: This is not needed.

zaza/openstack/charm_tests/keystone/tests.py
ajkavanagh: It's more about keystone not being ready by the time the code gets to here.  As everything is async retries tend to be the only way to get things back in "sync" again.

zaza/openstack/charm_tests/keystone/tests.py
ajkavanagh: Try taking a look at `config_change` in `test_utils.py` (of this library).  It let's you make a change to the config, and it waits for it to settle before moving on to what ever needs checking.

This should be done at them top where the config is performed., along with the wait.  I suspect that this wait is okay as it's an async "let's wait for things to settle".

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -371,3 +372,87 @@ def test_security_checklist(self):
...
+
+    def test_100_keystone_ldap_users(self):
+        """Validate basic functionality of keystone API with ldap."""
+        zaza.model.set_application_config('keystone-ldap',
+                                          self._get_ldap_config())
+
+        if not self.ldap_configured:
+            msg = 'Skipping API tests as no LDAP test fixture'
+            logging.info(msg)
+            return
```
ajkavanagh: Consider using `unittest.SkipTest(...)`; it gives a nicer output.  usage:

```python
if not self.ldap_configured:
    raise unittest.SkipTest("Skipping LDAP API tests as it's not configured")
```

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -371,3 +372,87 @@ def test_security_checklist(self):
...
+                'workload-status': 'idle',
+                'workload-status-messages': 'Unit is ready'
+            }
+        }
+        zaza.model.wait_for_application_states(states=states)
+
+        # NOTE(jamespage): Test fixture should have johndoe and janedoe
+        #                  accounts
+        johndoe = self._find_keystone_v3_user('john doe', 'userdomain')
+        assert johndoe is not None
```
ajkavanagh: This is nicer as:

```python
self.assertIsNotNone(johndoe, "'john doe' user not found")
```

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -371,3 +372,87 @@ def test_security_checklist(self):
...
+            if username.lower() == user.name.lower():
+                return user
+        logging.debug("User {} was not in these users: {}. Returning None."
+                      "".format(username, usernames))
+        return None
+
+    def test_100_keystone_ldap_users(self):
+        """Validate basic functionality of keystone API with ldap."""
+        zaza.model.set_application_config('keystone-ldap',
+                                          self._get_ldap_config())
```
ajkavanagh: Take a look at `config_change(...)` in `test_utils.py`

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -371,3 +372,87 @@ def test_security_checklist(self):
...
+
+        if not self.ldap_configured:
```
ajkavanagh: It would be better if `_get_ldap_config()` returned a tuple of `(configured, config_data)` or raises an Exception if the LDAP couldn't be configured.  By setting an instance var in the method, this is effectively a side-effect leak which is much harder to refactor and reason about.

e.g.
```python
(configured, ldap_config) = self._get_ldap_config()
if not configured:
    raise unittest.SkipTest(...)
# now do config
with config_change(...):
     # now do the  checks
```

This way you can avoid starting the config change before you have the config (rather than setting it to `None`) and it's more obvious *where* self.ldap_configured is actually coming from, as it a return value from the function rather than an object instance var.

The alternative is to raise the `SkipTest` *in* `_get_ldap_config()` which would skip the test and bail early; however, I tend to prefer a more explicit approach.

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -371,3 +371,73 @@ def test_security_checklist(self):
...
+        """Validate basic functionality of keystone API with ldap."""
+        application_name = 'keystone-ldap'
+        can_config, config = self._get_ldap_config()
+        if not can_config:
+            raise unittest.SkipTest(
+                "Skipping API tests as no LDAP test fixture"
+            )
+
+        with self.config_change(
+                self.config_current(application_name),
```
addyess: here's the use of a new method `config_current`

zaza/openstack/charm_tests/test_utils.py
```diff
@@ -133,6 +133,29 @@ def setUpClass(cls, application_name=None, model_alias=None):
             model_name=cls.model_name)
         logging.debug('Leader unit is {}'.format(cls.lead_unit))
 
+    def config_current(self, application_name=None, keys=None):
```
addyess: new method here to get the current config from an application.  You can supply `keys` if you only want certain items from the current config.

zaza/openstack/charm_tests/test_utils.py
```diff
@@ -133,6 +133,29 @@ def setUpClass(cls, application_name=None, model_alias=None):
...
+        """Get Current Config of an application normalized into key-values.
+
+        :param application_name: String application name for use when called
+                                 by a charm under test other than the object's
+                                 application.
+        :type application_name: str
+        :param keys: iterable of strs to index into the current config.  If
+                     None, return all keys from the config
+        :type keys: Union[iterable[str], None]
```
ajkavanagh: It's great that you are using types, and an easier/simpler type here is: `Optional[iterable[str]]` which is the same as `Union[iterable[str], None]`

This is also missing the :return: and :rtype: items in the docstring for the return type.

ajkavanagh:
Hi Adam

I not sure, but it looks like you've switched all the docstring params/types to `@something:` from `:something:`  .. the latter (`:something:`) is more correct.  We used to use `@param` in Juju and older code, but we switched to the current form as it's more widespread/accepted by the Python community.  I realise it's confusing!  (particularly when you look at charm-helpers and see that both styles are in use!)

addyess:
force of habit.  I've been doing `@param:` for too long.  Heck i may even have my editor set-up that way.  

I'll fix that up

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -371,3 +370,79 @@ def test_security_checklist(self):
...
+
+    def _get_ldap_config(self):
+        """Generate ldap config for current model.
+
+        :return: tuple of whether ldap-server is running and if so, config
+            for the keystone-ldap application.
+        :rtype: Tuple[bool, Dict[str,str]]
+        """
+        ldap_ips = zaza.model.get_app_ips("ldap-server")
+        self.assertTrue(ldap_ips, "Should be at least one ldap server")
```
addyess: @ajkavanagh @thedac let's assert if the ldap-server isn't there.

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -371,3 +370,79 @@ def test_security_checklist(self):
...
+        with self.config_change(
+                self.config_current(application_name),
+                config,
+                application_name=application_name):
+            logging.info(
+                'Waiting for users to become available in keystone...'
+            )
+            states = {
+                'keystone': {
+                    'workload-status': 'idle',
```
thedac: There is another method for checking for idleness. workload-state should be active here. This is, however, the default, making the setting of states unecessary.
The best practice is to check the tests.yaml which allows future developers to change expectations:
        test_config = lifecycle_utils.get_charm_config(fatal=False)
        zaza.model.wait_for_application_states(
            states=test_config.get("target_deploy_status", {}))

zaza/openstack/charm_tests/test_utils.py
```diff
@@ -133,6 +133,31 @@ def setUpClass(cls, application_name=None, model_alias=None):
...
+        :return: Dictionary of requested config from application
+        :rtype: Dict[str, str]
+        """
+        if not application_name:
+            application_name = self.application_name
+        _app_config = model.get_application_config(application_name)
+        # convert the more elaborate config structure from libjuju to key-value
+        keys = keys or _app_config.keys()
+        # note that conversion to string for all values is due to
+        # attempting to set any config with other types lead to Traceback
```
thedac: s/lead/leads
Though the whole sentence could be clearer.

Beyond the comment, I am not sold on changing the types. I would be interested to see the traceback when you don't set things to a string.

addyess: if i remember rightly, the traceback was trying to cast a go native type into a python type?  That may be why str(*) is used here.  Ill see if i can reproduce.  

Also, I think I remember that passing the config back with `model.set_application_config` fussed about the types?

addyess: yeah... here's the issue.  You can 'get' the current config such that the types are preserved (aka -- not mapped through str) but when you play them back into the model.set_application_config -- it doesn't like anything that isn't str typed.  
```
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/zaza/model.py", line 531, in async_set_application_config
2020-01-24 12:40:44 [INFO]     .set_config(configuration))
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/juju/application.py", line 383, in set_config
2020-01-24 12:40:44 [INFO]     return await app_facade.Set(application=self.name, options=config)
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/juju/client/facade.py", line 471, in wrapper
2020-01-24 12:40:44 [INFO]     reply = await f(*args, **kwargs)
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/juju/client/_client8.py", line 1124, in Set
2020-01-24 12:40:44 [INFO]     reply = await self.rpc(msg)
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/juju/client/facade.py", line 607, in rpc
2020-01-24 12:40:44 [INFO]     result = await self.connection.rpc(msg, encoder=TypeEncoder)
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/juju/client/connection.py", line 456, in rpc
2020-01-24 12:40:44 [INFO]     raise errors.JujuAPIError(result)
2020-01-24 12:40:44 [INFO] juju.errors.JujuAPIError: json: cannot unmarshal bool into Go struct field ApplicationSet.options of type string
```

zaza/openstack/charm_tests/keystone/tests.py
addyess: did this!

PR #164, opened 2020-01-31
Add osd removal test

ChrisMacNaughton:


ajkavanagh:
@ChrisMacNaughton is this still needed?  Added stale so we can track any further progress.

PR #170, opened 2020-02-04
designate zaza tests

addyess:
In order to migrate [**charm-designate**](https://review.opendev.org/#/c/705864) and [**charm-designate-bind**](https://review.opendev.org/#/c/705726/) to use zaza test framework [bug](https://launchpad.net/bugs/1828424)


zaza/openstack/charm_tests/designate/tests.py
```diff
@@ -0,0 +1,491 @@
...
+        for ep in endpoints:
+            logging.debug(ep)
+            self.assertIsNotNone(ep.id)
+            self.assertEqual(ep.region, "RegionOne")
+            self.assertIsNotNone(ep.service_id)
+            self.assertIn(ep.interface, self.VALID_INTERFACE)
+            self.assertRegexpMatches(ep.url, self.VALID_URL)
+
+
+class KeystoneIdentityRelationTests(BaseDesignateTest):
```
thedac: All relation validation can be removed.

zaza/openstack/charm_tests/designate/tests.py
```diff
@@ -0,0 +1,491 @@
...
+            "service_password missing"
+        )
+
+        self.assertIsNotNone(
+            relation.get('service_tenant_id'),
+            "service_tenant_id missing"
+        )
+
+
+class AmqpRelationTests(BaseDesignateTest):
```
thedac: All relation validation can be removed.

zaza/openstack/charm_tests/designate/tests.py
```diff
@@ -0,0 +1,491 @@
...
+        rel_private_ip = relation.get('private-address')
+        hostname = relation.get('hostname')
+        password = relation.get('password')
+
+        self.assertEqual(rel_private_ip, remote_ip)
+        self.assertEqual(hostname, remote_ip)
+        self.assertIsNotNone(password)
+
+
+class NeutronApiRelationTests(BaseDesignateTest):
```
thedac: All relation validation can be removed.

zaza/openstack/charm_tests/designate/tests.py
```diff
@@ -0,0 +1,491 @@
...
+            remote_unit_name,
+            relation_name
+        )
+        # Get private-address in relation
+        rel_private_ip = relation.get('private-address')
+        # The private address in relation should match designate-bind/0 address
+        self.assertEqual(rel_private_ip, remote_ip)
+
+
+class BindRelationTests(BaseDesignateTest):
```
thedac: All relation validation can be removed.

zaza/openstack/charm_tests/designate/tests.py
```diff
@@ -0,0 +1,491 @@
...
+        Pause service and check services are stopped then resume and check
+        they are started
+        """
+        with self.pause_resume(
+                self.designate_svcs,
+                pgrep_full=False):
+            logging.info("Testing pause resume")
+
+
+class ServerCreationTest(BaseDesignateTest):
```
thedac: I think we need one overarching class DesignateTests which includes ServerCreationTest and DomainCreationTests and the restart and puase/resume tests.

zaza/openstack/charm_tests/designate/tests.py
```diff
@@ -0,0 +1,197 @@
...
+        wait=tenacity.wait_exponential(multiplier=1, min=5, max=10),
+        reraise=True
+    )
+    def _wait_on_server_gone(self):
+        logging.debug('Waiting for server to disappear')
+        return not self._get_test_server_id()
+
+    def test_400_server_creation(self):
+        """Simple api calls to create domain."""
+        # Designate does not allow the last server to be delete so ensure ns1
```
ChrisMacNaughton: "to be delete" -> "to be deleted"

zaza/openstack/charm_tests/designate/tests.py
```diff
@@ -0,0 +1,197 @@
...
+            self.designate.recordsets.create(
+                _domain_id, _record.name, _record.type, [_record.data])
+        else:
+            _domain_id = new_domain.id
+            self.designate.records.create(_domain_id, _record)
+
+        self._wait_to_resolve_test_record()
+
+        logging.debug('Tidy up delete test record')
+        self.zones_delete(_domain_id)
```
ajkavanagh: I'd wait for the zone to be deleted as well; at some point this test could contaminate another test.

zaza/openstack/charm_tests/designate/__init__.py
```diff
@@ -0,0 +1,72 @@
+# Copyright 2020 Canonical Ltd.
```
ajkavanagh: I'm curious as to why this base class definition is in `__init__.py` and not just before the only usage of the class? Why not put it in the `tests.py` file?

addyess: i think i was following the keystone tests as a template.  Maybe it's the one off instead of the standard.  I'll move it.

zaza/openstack/charm_tests/designate/tests.py
addyess: @ajkavanagh, i'm now confused at this point about this test. 

Here's the [original](https://opendev.org/openstack/charm-designate-bind/src/branch/master/src/tests/basic_deployment.py#L274).  it seems likes we are deletings the zone, but not the domain?  Is that because the domain is contained within the zone?  Does that hold true for cases where zones weren't available (pre-xenial-queens)

addyess:
@thedac -- thanks for the review.  I'm at my best when I code with 'backspace'

zaza/openstack/charm_tests/designate/tests.py
ajkavanagh: It was more that the one further up in this test function looks like this:

            self.zones_delete(old_dom_id)
            self._wait_on_domain_gone()

i.e. wait until it is gone.  It would be useful if this test ensured that it cleaned up after itself (which it does), but it exits *before* it is sure that the test record is gone.  I'd just think that the equivalent of `_wait_on_domain_gone()` for the new domain that's been created should be done before the function exits.  My reasoning is that the function ought to be idempotent with respect to the resources it creates/removes.

addyess:
does anyone think I ought to port in this test too? 
https://review.opendev.org/#/c/597009

zaza/openstack/charm_tests/designate/tests.py
```diff
@@ -194,4 +250,5 @@ def test_400_domain_creation(self):
 
         logging.debug('Tidy up delete test record')
         self.zones_delete(_domain_id)
+        self._wait_on_domain_gone()
```
ajkavanagh: I think this will wait on the wrong domain_id?  I meant a function that waits on `_domain_id`?  However, I don't now as it's very confusing what is going on.  Does this wait on the right domain id?

addyess: @ajkavanagh , i think i have refactored this in a way that is more straight forward... let me know if it doesn't

zaza/openstack/charm_tests/designate/tests.py
ajkavanagh: I can't tell.  Does it wait on `_domain_id` which is `new_domain['id']` and is that different or the same as `old_dom_id` above; if different, then does it actually wait?  if the same then why a different variable name?

addyess: honestly... i don't know.  i'll try to understand what the original amulet test was trying to accomplish as much of this structure remains as a straight port of old tests.

ajkavanagh: Okay, I appreciate this is complicated!

addyess: @ajkavanagh -- i think i cracked it.  Lemme know what you think

PR #171, opened 2020-02-05
ovn: Update bridge configuration code

fnordahl:
After power user feedback we want to change how interfaces are
attached to bridges on the OVN charms.

Ref: openstack-charmers/charm-layer-ovn#6


fnordahl:
Functional test results with a `ovn-chassis` charm that includes changes from openstack-charmers/charm-layer-ovn#6 can be viewed in https://review.opendev.org/705951

PR #173, opened 2020-02-06
Standardize preferred-api-version as an int

thedac:


thedac:
Validated against keystone manually and running in OSCI against mysql-router and mysql-innodb-cluster now.

https://review.opendev.org/706033
https://review.opendev.org/706034

PR #190, opened 2020-03-10
Fixes for masakari tests

gnuoy:
* Deal with hypervisors now using fqdn
* Skip instant restart tests until bug is resolved


ChrisMacNaughton:
Looks good, thanks!

PR #196, opened 2020-03-16
mysql: Add tests for 'seeded' file presence

freyes:
This patchset adds 2 tests: 
1. Check the presence of /var/lib/percona-xtradb-cluster/seeded file in all the deployed units.
1. Test the mark-seeded action by removing the `seeded` file and recreating it with the help of the action.

Related-Bug: #1867456


freyes:
A patch that implements the mark-seeded action and that uses this branch to test it can be found at https://review.opendev.org/#/c/713316/

freyes:
I will split the commits to make them 2 different pull requests with their respective associated patches in percona-cluster charm.

PR #203, opened 2020-03-24
mysql: add check for seeded file in pxc

freyes:
Verify the seeded file is present with the expected content since it will
be used by *-relation-changed hook to determine if it can process requests
of related units.

Related-Bug: #1868326


ajkavanagh (approved):
LGTM

ajkavanagh:
Note; this didn't get merged at the same time as the gerrit review; hopefully it will add to the test coverage for seed, which is still broken for series upgrade.

PR #204, opened 2020-03-25
mysql: add test for mark-seeded action

freyes:
This test removes the seeded file by ssh, then runs the mark-seeded
action and checks the output of the action.

Related-bug: #1867456


ajkavanagh:
@freyes is this still needed; the associated bug is still open.  Added stale label to track future progress.

freyes:
I'm closing this PR, since it has been stale, and I have no time revisit, also percona-cluster is in maintenance mode, deployments now are expected to use mysql-innodb-cluster

PR #206, opened 2020-03-26
Add functional tests for charm-keystone-kerberos

camille-rodriguez:
Add functional tests for the charm-keystone-kerberos. Setup.py configures the environment post-deploy. The main test is to verify that a keystone authentication is successful via a verification of the user against a kerberos test server. The tests are done in a ubuntu server, since the test host had to be a kerberos client, which requires the install of several packages. 

Dependent on this other PR : https://github.com/openstack-charmers/charm-keystone-kerberos/pull/1


zaza/openstack/charm_tests/kerberos/setup.py
```diff
@@ -0,0 +1,239 @@
+# Copyright 2020 Canonical Ltd.
```
ryan-beisner: As discussed 1:1, several methods in zaza/openstack/charm_tests/kerberos/setup.py feels like code that is on the wrong side of the charm boundary, and should not typically be inside the test code.  My view is that the test fixture charm should typically be handling a lot of that functionality.

In this case, as a tactical measure for this tricky test fixture, we've agreed to go forward as-is. Otherwise, a fair amount of additional test fixture development time, including layers and interfaces, would be necessary to avoid the boundary concerns.

zaza/openstack/charm_tests/kerberos/tests.py
```diff
@@ -0,0 +1,74 @@
+# Copyright 2020 Canonical Ltd.
```
ryan-beisner: This test creatively introduces a new approach, not used elsewhere in Zaza to-date:  spinning up a generic cs:ubuntu unit and using that unit to execute client commands.  This is necessary here because the system needs to have Kerberos packages installed and the host needs to be configured with kinit, etc.  I think this is a sound and interesting approach.  We may want to distill some of this into general-purpose code at a later time, then update this test.

zaza/openstack/charm_tests/kerberos/setup.py
```diff
@@ -0,0 +1,239 @@
...
+        logging.info('Installing {}'.format(package))
+        result = zaza.model.run_on_unit(ubuntu_test_host.name,
+                                        "apt install {} -y".format(package))
+        assert result['Code'] == '0', result['Stderr']
+
+
+def install_python_packages_on_ubuntu_test_host():
+    """Install python packages on a zaza unit."""
+    ubuntu_test_host = zaza.model.get_units('ubuntu-test-host')[0]
+    package = 'keystoneauth1[kerberos]'
```
ryan-beisner: Is there a reason that we can't use the apt package here too?  If so, we would prefer that here, for the sake of having just one package (dependency) management system in play on the host.

Mixing userspace pip (ie. outside of a venv) with system-level packages is known to invite eventual trouble when dependency requirements clash.

zaza/openstack/charm_tests/kerberos/setup.py
```diff
@@ -0,0 +1,239 @@
...
+    logging.info('Installing {}'.format(package))
+    result = zaza.model.run_on_unit(ubuntu_test_host.name,
+                                    "pip install {} ".format(package))
+    assert result['Code'] == '0', result['Stderr']
+
+
+def run_all_tests():
```
ryan-beisner: I think the run_all_tests method name is a bit misleading, as they are all really configuration steps, and it doesn't actually include the most meaningful test here (CharmKeystoneKerberosTest).

You could either rename this method, or include ALL tests. I would lean toward a rename, mainly because of the Zaza test.yaml primitives of .configure and .test.

zaza/openstack/charm_tests/kerberos/setup.py
```diff
@@ -0,0 +1,239 @@
...
+            ubuntu_test_host.name,
+            tmp_file,
+            temp_krb5_path)
+        logging.info('Moving {} to {} on {}.'.format(temp_krb5_path,
+                     remote_file, ubuntu_test_host.name))
+        zaza.model.run_on_unit(ubuntu_test_host.name, ('sudo mv {} {}'.
+                               format(temp_krb5_path, remote_file)))
+
+
+def install_apt_packages_on_ubuntu_test_host():
```
ryan-beisner: Per 1:1 discussion, we always initialize OpenStack clients pythonically so that it tracks the same dependencies as defined in the requirements and test-requirements.

In this case, the client libraries are not coming from there, so we need to take note that of the risk.  As-is, the client package versions will always be aligned to the series of the cs:ubuntu unit, instead of being defined by the requirements.txt files.  At some point, this is nearly certain to break, as things are not always forward-compatible.

The best way to mitigate that risk, if there is no way to use the already-maintained requirements/test-requirements files, is to enable the Ubuntu Cloud Archive which aligns with the OpenStack version under test.

I'm not going to block on this now, but we need to log this as a bug for tech-debt to pay off here.

zaza/openstack/charm_tests/kerberos/setup.py
```diff
@@ -0,0 +1,239 @@
...
+    zaza.model.attach_resource('keystone-kerberos',
+                               'keystone_keytab',
+                               tmp_file)
+    logging.info('Waiting for keystone-kerberos unit to be active and idle')
+    unit_name = zaza.model.get_units('keystone-kerberos')[0].name
+    zaza.model.block_until_unit_wl_status(unit_name, "active")
+    zaza.model.block_until_all_units_idle()
+
+
+def add_dns_entry(kerberos_hostname="kerberos.testubuntu.com"):
```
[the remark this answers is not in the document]
camille-rodriguez: Right, I agree that it might not be ideal to have to edit this file. 

I think one way to get around that would be to configure the kerberos server with its original hostname, instead of changing it for "kerberos.testubuntu.com". I went this way to make it more obvious that requests towards that server would indeed be for the kerberos server, and to keep the same name every time the test fixture is deployed. 

Otherwise, hostnames of nodes deployed by juju are something like this "juju-75c90b-zaza-2b19a3604ff8-0.project.serverstack". But it would be possible to change the test-fixture to keep its original hostname, and the name resolution should work without editing the /etc/hosts file.

zaza/openstack/charm_tests/kerberos/setup.py
camille-rodriguez: I'm not sure actually. The Openstack docs uses pip to install this extra module. https://docs.openstack.org/keystoneauth/latest/api/keystoneauth1.extras.kerberos.html 

I know that python-keystoneauth1 can be installed with apt, but I believe it does not include the kerberos module. I can test and see if it is actually the case.

zaza/openstack/charm_tests/kerberos/setup.py
camille-rodriguez: True, rename it is!

zaza/openstack/charm_tests/kerberos/setup.py
camille-rodriguez: I've tried installing python-keystoneauth1 or python-keystoneauth1[kerberos] instead of pip install and it doesn't work. If there is another way to do this that I'm not away of, please let me know, otherwise I think we need to go forward with the pip install.

ryan-beisner: I think this still needs more work.  AFAICT, python3-keystoneauth1 contains the extras.kerberos bits.  Please evaluate with purely python3 packages via apt.

camille-rodriguez: This has been replaced by python3-openstackclient and python3-requests-kerberos.

PR #212, opened 2020-04-01
Add custom assertions

gnuoy:
Add custom assertions to check that remote actions and commands
run ok.

### DIFF
--- a/zaza/openstack/charm_tests/test_utils.py
+++ b/zaza/openstack/charm_tests/test_utils.py
@@ -89,51 +89,7 @@ def audit_assertions(action,
             assert value == "PASS", "Unexpected failure: {}".format(key)
 
 
-class CharmTestAssertions:
-    """Custom assertions to support zaza testings."""
-
-    def assertActionRanOK(self, action):
-        """Assert that the remote action ran successfully.
-
-        Example usage::
-
-            self.assertActionRanOK(model.run_action(
-                unit,
-                'pause',
-                model_name=self.model_name))
-
-            self.assertActionRanOK(model.run_action_on_leader(
-                unit,
-                'pause',
-                model_name=self.model_name))
-
-        :param action: Action object to check.
-        :type action: juju.action.Action
-        :raises: AssertionError if the assertion fails.
-        """
-        if action.status != 'completed':
-            msg = ("Action '{name}' exited with status '{status}': "
-                   "'{message}'").format(**action.data)
-            raise AssertionError(msg)
-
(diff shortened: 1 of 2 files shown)


zaza/openstack/charm_tests/test_utils.py
```diff
@@ -89,7 +89,53 @@ def audit_assertions(action,
             assert value == "PASS", "Unexpected failure: {}".format(key)
 
 
-class BaseCharmTest(unittest.TestCase):
+class CharmTestAssertions:
+    """Custom assertions to support zaza testings."""
+
```
ajkavanagh, line 94: Not sure why it's a class?  Is it for consistency in terms of using `self.someassertion(...)` in a test file.

zaza/openstack/charm_tests/test_utils.py
```diff
@@ -89,7 +89,53 @@ def audit_assertions(action,
...
+        :type action: juju.action.Action
+        :raises: AssertionError if the assertion fails.
+        """
+        print("Checking action")
+        if action.status != 'completed':
+            msg = ("Action '{name}' exited with status '{status}': "
+                   "'{message}'").format(**action.data)
+            raise AssertionError(msg)
+
+    def assertRemoteRunOK(self, run_output):
```
ajkavanagh, line 119: The function doesn't use self.  It can be a `@staticmethod`.

zaza/openstack/charm_tests/test_utils.py
```diff
@@ -89,7 +89,53 @@ def audit_assertions(action,
             assert value == "PASS", "Unexpected failure: {}".format(key)
 
 
-class BaseCharmTest(unittest.TestCase):
+class CharmTestAssertions:
+    """Custom assertions to support zaza testings."""
+
+    def assertActionRanOK(self, action):
```
ajkavanagh, line 95: This function doesn't use `self`.  Define using `@staticmethod`?

zaza/openstack/charm_tests/test_utils.py
gnuoy: On second thoughts these are useful outside of the test classes so I will move them.

zaza/openstack/charm_tests/test_utils.py
gnuoy: I'll move them

zaza/openstack/charm_tests/test_utils.py
gnuoy: I'll move it

PR #217, opened 2020-04-02
Neutron gateway managed bridges

AurelienLourot:


tests/tests.yaml
```diff
@@ -6,6 +6,6 @@ target_deploy_status:
     workload-status: active
     workload-status-message: icmp ok, local hostname ok, dns ok
 configure:
-- zaza.openstack.charm_tests.noop.setup.basic_setup
+- zaza.charm_tests.noop.setup.basic_setup
 tests:
-- zaza.openstack.charm_tests.noop.tests.NoopTest
+- zaza.charm_tests.noop.tests.NoopTest
```
AurelienLourot, line 11: This has been wrong for a year and there is no tox target pointing to this `tests/` folder, although the readme in this folder says:

> This directory contains functional test definition for functional test of Zaza
> itself.

Is this `tests/` folder really still valid?

ChrisMacNaughton: This is absolutely 100% a correct import path. This is coming from zaza itself, rather than from zaza.openstack

AurelienLourot: @ChrisMacNaughton @fnordahl As discussed offline I got rid of the entire `tests/` folder for now. 

For the record what we noticed is:
1. `tox.ini` doesn't have a `func` target so probably these tests haven't been run for a long time.
2. These tests can't be run at the moment because of conflicting version ranges in python deps pulled via `requirements.txt` and `setup.py`, e.g.

```
error: PyYAML 3.13 is installed but pyyaml<=6.0,>=5.1.2 is required by {'juju'}
error: python-keystoneclient 3.21.0 is installed but python-keystoneclient>=3.22.0 is required by {'python-openstackclient'}
```

AurelienLourot:
CI runs for the new neutron-gateway test: https://review.opendev.org/717074

tests/tests.yaml
ajkavanagh: I'm not sure why the `tests/` directory is being removed as part of the PR?  It may be that it *should* be removed, but it shouldn't be done as part of this PR unless it is directly relevant to the functionality that is being tested (neutron gateway managed bridges).

The reason is, should we wish to re-instate the `tests` directory it would be more awkward as it is mixed in with this functionality.

zaza/openstack/charm_tests/neutron/tests.py
```diff
@@ -118,7 +146,7 @@ def test_900_restart_on_config_change(self):
         """
         # Expected default and alternate values
         current_value = zaza.model.get_application_config(
-            'neutron-gateway')['debug']['value']
+            self._APP_NAME)['debug']['value']
```
ajkavanagh, line 149: Where does `_APP_NAME` come from?  I've grepped the code and can't seem to find it.  Is it Juju specific itself?

AurelienLourot: new line 116 of this file :)

tests/tests.yaml
AurelienLourot: Thanks, makes sense, moved [here](https://github.com/openstack-charmers/zaza-openstack-tests/pull/219)

PR #218, opened 2020-04-02
Update path to the noop test

ChrisMacNaughton:


ChrisMacNaughton:
This is already proposed in https://github.com/openstack-charmers/zaza-openstack-tests/pull/217/files#diff-04c6e90faac2675aa89e2176d2eec7d8

ajkavanagh:
@ChrisMacNaughton I think you should re-open this and do it here.  It's mixing concerns to do the change in the other branch; it will be cleaner if it's done here first.

PR #222, opened 2020-04-07
octavia: tear down resources created during LBaasV2 test

### LINKED ISSUE #221: octavia: High probability of running over quota during test
Test artifacts showing the issue: https://openstack-ci-reports.ubuntu.com/artifacts/test_charm_pipeline_func_full/openstack/charm-octavia/709079/7/5267/index.html

The test has ``loadbalancer-topology`` set to 'ACTIVE_STANDBY' which means that each load balancer consumes two instances, we also have ``spare-pool-size`` set to '2' which means two load balancers (consuming 4 instances) will be kept as spares at any given time.  As part of test setup we start two instances to serve payload data.  The LBaasV2 test does not clean up after itself, so at the point in time where we are testing the policyd stuff chances are high we would be over the quota of '10' running instances.

fnordahl:
Fixes #221

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,8 @@ matrix:
     - name: "Python 3.7"
       python: 3.7
       env: ENV=pep8,py3
+    - name: "Python 3.8"
+      python: 3.7
+      env: ENV=pep8,py3
 script:
   - tox -c tox.ini -e $ENV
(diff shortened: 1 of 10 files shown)


fnordahl:
A set of successful test runs with this code can be viewed in [0], while a deployment at the end failed due to unrelated issues it still demonstrates that the code in this PR works for the successful runs.

0: https://openstack-ci-reports.ubuntu.com/artifacts/test_charm_pipeline_func_full/openstack/charm-octavia/709079/8/5281/consoleText.test_charm_func_full_8453.txt

PR #224, opened 2020-04-08
Series upgrade can be done in massively parallel batches

ChrisMacNaughton:
Due to the nature of the control plane outage required for a series upgrade, most of the API services can be upgraded massively in parallel. The idea is that it is safe to upgrade things in batches like:

```python
SERVICE_GROUPS = collections.OrderedDict([
    ('Stateful Services', ['percona-cluster', 'rabbitmq-server']),
    ('Core Identity', ['keystone', 'ceph-mon']),
    ('Control Plane', [
        'aodh', 'barbican', 'ceilometer', 'ceph-fs',
        'ceph-radosgw', 'cinder', 'designate',
        'designate-bind', 'glance', 'gnocchi', 'heat', 'manila',
        'manila-generic', 'neutron-api', 'neutron-gateway', 'placement',
        'nova-cloud-controller', 'openstack-dashboard']),
    ('Data Plane', [
        'nova-compute', 'ceph-osd', 'swift-proxy', 'swift-storage'])
])
```

Given the above ordering (Stateful Services, Core Identity, etc), the Control Plane group can series-upgrade every unit in parallel. This can drastically reduce the time to perform a series upgrade!

### DIFF
--- a/zaza/openstack/utilities/parallel_series_upgrade.py
+++ b/zaza/openstack/utilities/parallel_series_upgrade.py
@@ -125,6 +125,16 @@ async def parallel_series_upgrade(
     :param origin: The configuration setting variable name for changing origin
                    source. (openstack-origin or source)
     :type origin: str
+    :param pause_non_leader_primary: Whether the non-leader applications should
+                                     be paused
+    :type pause_non_leader_primary: bool
+    :param pause_non_leader_subordinate: Whether the non-leader subordinate
+                                         hacluster applications should be
+                                         paused
+    :type pause_non_leader_subordinate: bool
+    :param post_upgrade_functions: A list of Zaza functions to call when
+                                   the upgrade is complete on each machine
+    :type post_upgrade_functions: List[str]
     :param files: Workaround files to scp to unit under upgrade
     :type files: list
     :param workaround_script: Workaround script to run during series upgrade
@@ -170,14 +180,14 @@ async def parallel_series_upgrade(
     await prepare_series_upgrade(leader_machine, to_series=to_series)
     if leader_machine not in completed_machines:
         machines.append(leader_machine)
-    # do the dist upgrade
-    await dist_upgrades(machines)
-    # do a do-release-upgrade
-    await do_release_upgrades(machines)
-    # do a reboot
-    await reboots(machines)
-
-    await complete_series_upgrade(machines, to_series)
+    upgrade_group = [
+        series_upgrade_machine(
+            machine, from_series=from_series, to_series=to_series,
+            files=files, workaround_script=None,
+            post_upgrade_functions=post_upgrade_functions)
+        for machine in machines
+    ]
+    await asyncio.gather(*upgrade_group)
     if origin:
         await os_utils.async_set_origin(application, origin)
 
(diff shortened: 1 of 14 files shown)


ChrisMacNaughton:
Something that's worth discussing is that the current versions of the series upgrade code have `files` and `workaround_script` as options. This version currently doesn't as I couldn't find anything using those bits of code. I may have missed something so I'd love feedback from @thedac about it

zaza/openstack/utilities/parallel_series_upgrade.py
```diff
@@ -0,0 +1,639 @@
...
+    pause_non_leader_subordinate=True,
+    pre_upgrade_functions=None,
+    post_upgrade_functions=None,
+    post_application_upgrade_functions=None,
+    completed_machines=None,
+    follower_first=False,
+    files=None,
+    workaround_script=None
+):
+    """Perform series upgrade on an application in parallel.
```
gnuoy, line 251: s/parallel/series/

zaza/openstack/utilities/parallel_series_upgrade.py
```diff
@@ -0,0 +1,639 @@
...
+    """
+    try:
+        await run_on_machine(machine, 'sudo init 6 & exit')
+        # await run_on_machine(unit, "sudo reboot && exit")
+    except subprocess.CalledProcessError as e:
+        logging.warn("Error doing reboot: {}".format(e))
+        pass
+
+
+async def run_on_machine(machine, command, model_name=None, timeout=None):
```
gnuoy, line 540: Is there not a method in zaza.model that does this already?

zaza/openstack/utilities/parallel_series_upgrade.py
```diff
@@ -0,0 +1,639 @@
...
+    """
+    logging.info('Upgrading ' + machine)
+    do_release_upgrade_cmd = (
+        'yes | sudo DEBIAN_FRONTEND=noninteractive '
+        'do-release-upgrade -f DistUpgradeViewNonInteractive')
+
+    await run_on_machine(machine, do_release_upgrade_cmd, timeout="120m")
+
+
+# TODO: Move these functions into zaza.model
```
gnuoy, line 600: +1

zaza/openstack/utilities/parallel_series_upgrade.py
```diff
@@ -0,0 +1,639 @@
...
+
+
+def _unit_idle(app, unit_name):
+    async def f():
+        x = await get_agent_status(app, unit_name)
+        return x == "idle"
+    return f
+
+
+async def get_agent_status(app, unit_name):
```
gnuoy, line 628: Seems like another method that could go over to zaza.model

zaza/openstack/utilities/parallel_series_upgrade.py
ChrisMacNaughton: This is performing a series upgrade, in parallel

zaza/openstack/utilities/parallel_series_upgrade.py
ChrisMacNaughton: Currently, nope; there's just `run_on_unit` and `run_on_leader`, both of which expect an actual application unit name

PR #226, opened 2020-04-10
Add codecov configuration to exclude the charm tests

ChrisMacNaughton:
This does not, however, exclude the configure code in
charm_tests as that can, and should, be unit tested :)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+ignore:
+  - "zaza/openstack/charm_tests/**/*tests.py"


codecov.yml
```diff
@@ -0,0 +1,2 @@
+ignore:
+  - "zaza/openstack/charm_tests/**/*tests.py"  # ignore folders and all its contents
```
AurelienLourot, line 2: Why not just `zaza/openstack/charm_tests/`? https://docs.codecov.io/docs/ignoring-paths
Also nitpick: missing newline at the end of the file

ChrisMacNaughton: because I want to include "zaza/openstack/charm_tests/*/configure.py" ;-)

AurelienLourot: you mean zaza/openstack/charm_tests/dragent/configure.py ? (it's the only one I could find)
I don't see it being imported in the local tests. Do you think it should?

ChrisMacNaughton: Broadly, I mean everything that isn't actually a `unittest.TestCase`. There are a lot of helpers in the zaza/openstack/charm_tests directories that are standalone functions (ex: [Glance's image setup](https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/charm_tests/glance/setup.py#L33)) that absolutely could & should have unit test coverage!

PR #231, opened 2020-04-16
Fix copy-paste mistakes and typos and remove dead code

AurelienLourot:


zaza/openstack/charm_tests/cinder/tests.py
```diff
@@ -153,7 +153,7 @@ def test_120_volume_force_delete(self):
 
     @property
     def services(self):
-        """Return a list services for OpenStack release."""
+        """Return a list services for the selected Openstack release."""
```
ajkavanagh, line 156: According to https://www.openstack.org/ it really is "OpenStack".  Perhaps a global search/replace to pick them all up in comments?

PR #242, opened 2020-04-21
Add a helper for testing restart on changed

gnuoy:
Most api charms test restart on changed by flipping the value of
debug. This has led to the same test being copied around. To reduce
the boiler plate code this change adds a new test helper which can
be used instead.

### DIFF
--- a/zaza/openstack/charm_tests/test_utils.py
+++ b/zaza/openstack/charm_tests/test_utils.py
@@ -254,6 +254,41 @@ class BaseCharmTest(unittest.TestCase):
         # TODO: Optimize with a block on a specific application until idle.
         model.block_until_all_units_idle()
 
+    def restart_on_changed_use_debug(self, config_file, services):
+        """Check restart happens on config change by flipping debug mode.
+
+        Change debug mode and assert that change propagates to the correct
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/test_utils.py
```diff
@@ -254,6 +254,41 @@ def config_change(self, default_config, alternate_config,
         # TODO: Optimize with a block on a specific application until idle.
         model.block_until_all_units_idle()
 
+    def restart_on_changed_use_debug(self, config_file, services):
+        """Check restart happens on config change by flipping debug mode.
+
+        Change debug mode and assert that change propagates to the correct
```
AurelienLourot, line 260: actually it needs more documentation, or maybe it's not a good abstraction, because:
- it checks that config_file will be touched, and that isn't reflected in the name and description of the function.
- it works only for an OSLO config file where `debug` ends up in a `DEFAULT` section.

What if the provided config_file follows a different pattern. As a user when I see the function name I think "nice, this will allow me to check restart on debug, super easy" but in the end it requires me to know about a config file with quite some hard constraints on its syntax.

Still I'll approve this review because I think it's better than before.

PR #246, opened 2020-04-21
Migrate 499 cinder-ceph test to zaza

ajkavanagh:
This takes the 499 test from basic_deployment.py (Amulet) test and
ports it over to zaza.

### DIFF
--- a/zaza/openstack/charm_tests/ceph/mon/tests.py
+++ b/zaza/openstack/charm_tests/ceph/mon/tests.py
@@ -22,6 +22,7 @@ import zaza.model
 from zaza.openstack.utilities import (
     generic as generic_utils,
     openstack as openstack_utils,
+    exceptions as zaza_exceptions
 )
 import zaza.openstack.charm_tests.test_utils as test_utils
 
@@ -43,7 +44,7 @@ class CinderCephMonTest(test_utils.OpenStackBaseTest):
         current_release = openstack_utils.get_os_release()
         bionic_train = openstack_utils.get_os_release('bionic_train')
         if current_release < bionic_train:
-            units.extend(zaza.model.get_utils("cinder-ceph",
+            units.extend(zaza.model.get_units("cinder-ceph",
                                               model_name=self.model_name))
 
         commands = [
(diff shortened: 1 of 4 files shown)


zaza/openstack/charm_tests/ceph/mon/tests.py
```diff
@@ -0,0 +1,200 @@
...
+import zaza.openstack.charm_tests.test_utils as test_utils
+
+
+class CinderCephMonTest(test_utils.OpenStackBaseTest):
+    """Verify that the ceph mon units are healthy."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Run class setup for running ceph mon tests with cinder."""
+        super().setUpClass()
```
ChrisMacNaughton, line 35: This seems a bit boiler-plate - do we really need to include it if all we're doing is deferring to `super`?

zaza/openstack/charm_tests/ceph/mon/tests.py
```diff
@@ -0,0 +1,200 @@
...
+            does_relation_exist(
+                u.name, "ceph-mon", "cinder-ceph", "ceph", self.model_name)
+            for u in ceph_mon_units]
+        logging.info("Wait till model is idle ...")
+        zaza.model.block_until(*conditions)
+        zaza.model.block_until_all_units_idle()
+        logging.info("... Done.")
+
+
+def does_relation_exist(unit_name,
```
ChrisMacNaughton, line 126: This should probably be a more generic helper in Zaza. I'm tentatively OK not blocking on this so we can get it landed more quickly, but I'd like to see a bug to track moving this into Zaza

zaza/openstack/charm_tests/ceph/mon/tests.py
ajkavanagh: We probably don't, but I tend to leave them in as a reminder that there is that entry point. It can be dropped though.

zaza/openstack/charm_tests/ceph/mon/tests.py
ajkavanagh: Yes, I'll raise a bug.

ajkavanagh:
Bug #289 raised for refactoring bits to zaza library.

zaza/openstack/charm_tests/ceph/mon/tests.py
ChrisMacNaughton: Fair enough

zaza/openstack/charm_tests/ceph/mon/tests.py
ChrisMacNaughton: Thanks!

PR #249, opened 2020-04-22
Disable the pristine disk check on focal

ajkavanagh:
The test won't pass on serverstack due to the snapd taking a copy of the
/mnt/vdb info at boot time and thus not being able to umount it is root
without also running through all of the procs and checking if they also
have it mounted.  Thus, this just disables the test on focal and newer.

### DIFF
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -371,6 +371,13 @@ class CephTest(test_utils.OpenStackBaseTest):
         As the ephemeral device will have data on it we can use it to validate
         that these checks work as intended.
         """
+        current_release = zaza_openstack.get_os_release()
+        focal_ussuri = zaza_openstack.get_os_release('focal_ussuri')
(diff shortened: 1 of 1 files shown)


ajkavanagh:
It has actually passed its test if you check on travis; it looks like the notification didn't get back to github for some reason.

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -371,6 +371,13 @@ def test_blocked_when_non_pristine_disk_appears(self):
         As the ephemeral device will have data on it we can use it to validate
         that these checks work as intended.
         """
+        current_release = zaza_openstack.get_os_release()
+        focal_ussuri = zaza_openstack.get_os_release('focal_ussuri')
+        if current_release >= focal_ussuri:
+            # NOTE(ajkavanagh) - focal (on ServerStack) is broken for /dev/vdb
```
AurelienLourot, line 377: If you have a link to a bug it would be nice to put it here, thanks!

PR #250, opened 2020-04-25
Add test cases for nova shelving accounting in quotas

hemanthnakkina:
Add functional test case to verify new config param
quota-count-usage-from-placement in nova-cloud-controller
charm.

Charm changes - https://review.opendev.org/#/c/715182/

Related-Bug: #1864859


zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -384,6 +385,56 @@ def test_902_quota_settings(self):
...
+        """
+        # Expected default and alternate values
+        current_value = zaza.model.get_application_config(
+            'nova-cloud-controller')['quota-count-usage-from-placement']
+        try:
+            current_value = current_value['value']
+        except KeyError:
+            current_value = False
+        new_value = True
+        new_value_str = str(new_value).title()
```
ajkavanagh, line 404: What will the test do if the `current_value` is `True`?

hemanthnakkina: Wrong assumption that cloud will be deployed with default values. Corrected the functional test case to perform tests irrespective of juju config of quota-count-usage-from-placement on the deployed cloud. Please re-review.

PR #252, opened 2020-04-29
Add zaza tests for TrilioVault

javacruft:
Add setup and tests for trilio-{data-mover,dm-api,wlm} charms.

The actual functional test for TrilioVault still needs to be developed
but the configure steps validate quite a bit of the deployment.

### DIFF
--- a/zaza/openstack/charm_tests/trilio/setup.py
+++ b/zaza/openstack/charm_tests/trilio/setup.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python3
+
+# Copyright 2019 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Code for configuring Trilio."""
+
+import logging
+import os
+
+import zaza.model as zaza_model
+import zaza.openstack.utilities.juju as juju_utils
+import zaza.openstack.utilities.generic as generic_utils
+
+
+def basic_setup():
+    """Run setup for testing Trilio.
+
+    Setup for testing Trilio is currently part of functional
+    tests.
+    """
+    logging.info("Configuring NFS Server")
+    nfs_server_ip = zaza_model.get_app_ips("nfs-server-test-fixture")[0]
+    trilio_wlm_unit = zaza_model.get_first_unit_name("trilio-wlm")
+
+    nfs_shares_conf = {"nfs-shares": "{}:/srv/testing".format(nfs_server_ip)}
+    _trilio_services = ["trilio-wlm", "trilio-data-mover"]
+
+    conf_changed = False
+    for juju_service in _trilio_services:
+        app_config = zaza_model.get_application_config(juju_service)
+        if app_config["nfs-shares"] != nfs_shares_conf["nfs-shares"]:
+            zaza_model.set_application_config(juju_service, nfs_shares_conf)
+            conf_changed = True
+
+    if conf_changed:
+        zaza_model.wait_for_agent_status()
+        # NOTE(jamespage): wlm-api service must be running in order
+        #                  to execute the setup actions
+        zaza_model.block_until_service_status(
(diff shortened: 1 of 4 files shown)


zaza/openstack/charm_tests/trilio/setup.py
```diff
@@ -0,0 +1,80 @@
...
+    tests.
+    """
+    logging.info("Configuring NFS Server")
+    nfs_server_ip = zaza_model.get_app_ips("nfs-server-test-fixture")[0]
+    trilio_wlm_unit = zaza_model.get_first_unit_name("trilio-wlm")
+
+    nfs_shares_conf = {"nfs-shares": "{}:/srv/testing".format(nfs_server_ip)}
+    _trilio_services = ["trilio-wlm", "trilio-data-mover"]
+    for juju_service in _trilio_services:
+        zaza_model.set_application_config(juju_service, nfs_shares_conf)
```
fnordahl, line 39: In general doing this may be tricky as you don't know what the model will look like before you or someone else executes this code. If the config value happens to be the same as you are trying to set the test code will execute indefinitely waiting for something to happen that won't.

Having said that since you are setting a value from the underlay that most likely will be different every time it will most likely be ok here.

The [config_change](https://github.com/openstack-charmers/zaza-openstack-tests/blob/719c4bfacf7b6d662e21452195e576d136484963/zaza/openstack/charm_tests/test_utils.py#L177) context manager can be useful for these kinds of changes as it checks the already set value for you.

zaza/openstack/charm_tests/trilio/setup.py
```diff
@@ -0,0 +1,80 @@
...
+        target_status="active",
+    )
+
+    # NOTE(jamespage: due to a bug in the data-mover codebase, we
+    #                 have to use a big hammer and restart it once
+    #                 its been started for the first time.
+    logging.info("Restarting tvault-contego services")
+    for data_mover_unit in zaza_model.get_units('trilio-data-mover'):
+        zaza_model.run_action(unit_name=data_mover_unit.name,
+                              action_name='restart-services')
```
fnordahl, line 56: Perhaps add an assertion on the result of the action

zaza/openstack/charm_tests/trilio/tests.py
```diff
@@ -0,0 +1,95 @@
...
+    conf_file = None
+
+    def test_restart_on_config_change(self):
+        """Checking restart happens on config change.
+
+        Change debug mode and assert that change propagates to the correct
+        file and that services are restarted as a result
+        """
+        # Expected default and alternate values
+        set_default = {"debug": "False"}
```
fnordahl, line 36: Use `bool` values here, otherwise the config_change helper will not be able to detect if you are trying to set a configuration option to a value that is already set in the model.

zaza/openstack/utilities/openstack.py
```diff
@@ -487,6 +487,24 @@ def get_project_id(ks_client, project_name, api_version=2, domain_name=None):
     return None
 
 
+def get_domain_id(ks_client, domain_name, api_version=2):
```
dshcherb, line 490: @javacruft api_version looks unused.

zaza/openstack/utilities/openstack.py
```diff
@@ -487,6 +487,24 @@ def get_project_id(ks_client, project_name, api_version=2, domain_name=None):
...
+    """Return domain ID.
+
+    :param ks_client: Authenticated keystoneclient
+    :type ks_client: keystoneclient.v3.Client object
+    :param domain_name: Name of the domain
+    :type domain_name: string
+    :param api_version: API version number
```
dshcherb, line 497: Same as above.

zaza/openstack/charm_tests/trilio/tests.py
```diff
@@ -0,0 +1,379 @@
...
+        juju_utils.remote_run(
+            self.trilio_wlm_unit,
+            remote_cmd=self.ONECLICK_RESTORE_CMD.format(
+                auth_args=self.auth_args, snapshot_id=snapshot_id
+            ),
+            timeout=180,
+            fatal=True,
+        )
+
+        # TODO validate restore but currently failing with 4.0
```
dshcherb, line 245: OK, let's discuss this later.

dshcherb:
Looks good to me overall.

I made 1 comment about an extra argument in `get_domain_id` and we also need to figure out what to do about the restore functionality validation with 4.0.

javacruft:
I'm working on getting the restore test working still - there are some bugs/configuration issues in the data-mover which are currently blocking it from succeeding

dshcherb:
I see, I think we can do it in a follow-up PR then.

Just waiting for the full func test to pass on https://review.opendev.org/#/c/721683/ acknowledging that it needs a full cloud to be stood up.

Issue #256, opened 2020-05-04
OpenStackBaseTest doesn't pass setUpClass args to setUpClass of BaseCharmTest


dshcherb:
`OpenStackBaseTest` doesn't pass `setUpClass` method args to `setUpClass` of `BaseCharmTest`.

This leads to a quite unpleasant behavior seen in https://review.opendev.org/#/c/712980/1 (see my `-1` comment for the first patch set).

PR #257, opened 2020-05-04
Pass arguments BaseCharmTest.setUpClass

dshcherb:
Not doing so triggers an incorrect behavior leading to functional test
failures as the application name is not set correctly.

https://github.com/openstack-charmers/zaza-openstack-tests/issues/256
https://review.opendev.org/#/c/712980/1

### DIFF
--- a/unit_tests/charm_tests/test_utils.py
+++ b/unit_tests/charm_tests/test_utils.py
@@ -1,0 +1,32 @@
+# Copyright 2020 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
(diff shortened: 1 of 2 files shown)


ajkavanagh:
This is a great catch.  It's also a breaking change?  It would be good to see some test runs against a few charms to check that there's no unexpected behaviour in other tests that may cause us trouble -- we're 3 days from freeze!  Please could we test against, say (in a bastion):

 - aodh
 - cinder
 - nova-compute

and verify there's no weird  behaviour.

Note. I may be being over-cautious here, in which case, please feel free to comment!

dshcherb:
@ajkavanagh The change that introduced the issue was done quite recently: https://github.com/openstack-charmers/zaza-openstack-tests/commit/9cb7876e

So I think it's not a breaking change as far as the older tests go but the ones merged between then and now might be affected.

To find out whether any of them are I did the following:
```
# None of the lines are prefixed with `+` which means that the lines below are just the lines around new changes that were there before
git log -p 9cb7876e..master | grep -P '\.setUpClass\((\w|\d)+\)'
         super(NeutronOpenvSwitchTest, cls).setUpClass(cls)
         super(GenericPolicydTest, cls).setUpClass(application_name)
         super(BasePolicydSpecialization, cls).setUpClass(application_name)
         super(NeutronGatewayTest, cls).setUpClass(cls)

# Gives more results just for sanity checking and catching multi-line cases
git log -p 9cb7876e..master | grep -P 'setUpClass\('
```

There is a lesser possibility that some of those test cases began to pass after 9cb7876e was merged.

This change managed to fix it for https://review.opendev.org/#/c/712980/

ajkavanagh:
@dshcherb okay, I think we are probably pretty safe then.  Let's get this merged :)

PR #260, opened 2020-05-06
The octavia tests require multiple LTS images

ajkavanagh:
This patch ensures that bionic and focal images are available for
the LTS octavia tests.


fnordahl:
Note that the Octavia test uses the create LTS guest to create instances that provide the payload for the load balancer end to end test, not for creating the Amphora images.

Images for creating the amphora must come from GSS to get the required metadata for discovery.

PR #270, opened 2020-05-13
libjuju-2.8 breaks with juju 2.7.6 and 2.8.0 - pinning

ajkavanagh:
This is a temporary pin until a fix is provided.


ChrisMacNaughton:
This break is represented in https://github.com/juju/python-libjuju/issues/413

PR #277, opened 2020-05-16
Minor improvements to RmqTests.test_913_list_unconsumed_queues

AurelienLourot:
CI runs: https://review.opendev.org/725909

PR #283, opened 2020-05-18
Add cleanup to the unit removal case

dshcherb:
## Add cleanup to the unit removal case

For some reason test cases are sometimes executed out of order and so
test_921_remove_unit is sometimes run before the pause_and_resume test
case which results in an error.

While the root cause for it must be found it would also be good to avoid
side-effects in individual test cases and return the environment back to
its original state.

There is no 'start' hook implementation for charm-rabbitmq-server,
however, changes that close to the 20.05 release are discouraged so this
change uses an upgrade-charm event simulation to re-trigger the addition
of a unit (which was previously removed) to the cluster.

NOTE: after an execution of hooks/upgrade-charm finishes, the charm will
stay in the waiting state with the following status until the next
update-status event: 'Unit has peers, but RabbitMQ not clustered'

Related-Bug: #1730709

### DIFF
--- a/zaza/openstack/charm_tests/rabbitmq_server/tests.py
+++ b/zaza/openstack/charm_tests/rabbitmq_server/tests.py
@@ -76,7 +76,9 @@ class RmqTests(test_utils.OpenStackBaseTest):
             rmq_utils.configure_ssl_off(units)
 
         # Publish and get amqp messages in all possible unit combinations.
-        # Qty of checks == (qty of units) ^ 2
+        # Qty of checks == qty_of_units * (qty_of_units - 1)
+        assert len(units) >= 2, 'Test is useful only with 2 units or more.'
+
         amqp_msg_counter = 1
         host_names = generic_utils.get_unit_hostnames(units)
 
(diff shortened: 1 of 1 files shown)


dshcherb:
Merged with https://github.com/openstack-charmers/zaza-openstack-tests/pull/277 in order to add it to https://review.opendev.org/#/c/725909/ as a single PR reference.

zaza/openstack/charm_tests/rabbitmq_server/tests.py
```diff
@@ -356,25 +370,48 @@ def test_921_remove_unit(self):
...
+            for u in units:
+                e = self._retry_check_unit_cluster_nodes(u,
+                                                         unit_node_names)
+                if e:
+                    errors.append(e)
+
+            self.assertFalse(errors, msg=errors)
+
+        logging.info('Checking that all units except for {} are present'
+                     'in the cluster'.format(removed_unit))
```
AurelienLourot, line 401: I think you should use `format(removed_unit.entity_id)` to have a more human-readable output

zaza/openstack/charm_tests/rabbitmq_server/tests.py
```diff
@@ -356,25 +370,48 @@ def test_921_remove_unit(self):
...
-        for unit in unit_host_names:
-            unit_node_names.append('rabbit@{}'.format(unit_host_names[unit]))
-        errors = []
-
-        for u in left_units:
-            e = self._retry_check_unit_cluster_nodes(u,
-                                                     unit_node_names)
-            if e:
-                errors.append(e)
+        def check_units(units):
```
ajkavanagh, line 384: I tend to avoid embedding functions inside other functions as they don't seem to get the same level of scrutiny as top level functions, and are also subject to interesting closure rules.  Perhaps consider making this a static method of the class as a helper function, with a docstring?  Not for this review, though.

PR #286, opened 2020-05-18
Initial enablement of Bionic-to-Focal series upgrade

ryan-beisner:
Use the -d (development) flag on the do-release-upgrade command since do-release-upgrade does not recognize the new LTS until its first .1 point release.


AurelienLourot:
Thanks, I believe this isn't needed anymore

PR #287, opened 2020-05-19
Introduce topology changes in remove_and_add_unit

dshcherb:
dfc62353620d0229fec14b8f4a05fcfd6332c34c added a simulation of a node
re-addition to the cluster to complement the existing logic of a
simulated node removal from the cluster. This has proven to not be
enough after CI test runs, therefore, this change actually introduces
topology changes to the test case:

* a unit is removed from the cluster at the beginning of the test case;
* a new unit is added to the cluster at the end of it to restore the
  initial cluster count.

Related-Bug: #1730709

### DIFF
--- a/zaza/openstack/charm_tests/rabbitmq_server/tests.py
+++ b/zaza/openstack/charm_tests/rabbitmq_server/tests.py
@@ -354,15 +354,37 @@ class RmqTests(test_utils.OpenStackBaseTest):
     def _retry_check_unit_cluster_nodes(self, u, unit_node_names):
         return rmq_utils.check_unit_cluster_nodes(u, unit_node_names)
 
+    def _check_units_in_cluster(self, units):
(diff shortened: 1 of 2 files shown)


dshcherb:
Relies on the recent addition to Zaza https://github.com/openstack-charmers/zaza/pull/353.

I tested this on ServerStack by forcing the test execution order to be as follows:

* remove and add;
* pause and resume.
https://github.com/openstack-charmers/zaza-openstack-tests/compare/master...dshcherb:2020-05-18-730709-reorder-and-fix

But this change has the old order and just the changes to rework the test case.

ajkavanagh:
@dshcherb is this still required; it needs a rebase/conflict resolution, but would otherwise be good to go.  Added stale label to monitor progress.  Please remove the label if it's still a work in progress.

PR #288, opened 2020-05-19
Skip test_921_remove_and_add_unit

dshcherb:
Due to test stability issues we need to skip it and work on it after
release.


dshcherb:
@ajkavanagh A bug reference would definitely help: I added https://github.com/openstack-charmers/zaza-openstack-tests/issues/290 with an explanation of the current state of affairs.

PR #292, opened 2020-05-20
Add ceph tests to nova-compute

rodrigogansobarbieri:
This patch adds 3 new tests:

1) Test to make sure ceph broker messaging is being
   processed correctly.
2) Test to make sure there are not additional restarts
   caused by invoking a config-changed hook without changes.
3) Test to make sure there are not additional restarts
   caused by invoking ceph-relation-changed without changes.

Tests 1 and 3 need to be run having nova-comptue
configured to use a RBD backend, otherwise they are skipped.


zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -141,6 +142,107 @@ def test_930_check_virsh_default_network(self):
...
+                    if broker_rsp_dict['request-id'] in output:
+                        logging.info(
+                            'Found request-id in output: '
+                            '{}'.format(broker_rsp_dict['request-id']))
+                        marked_done = True
+
+            if not marked_done:
+                msg = "Ceph broker message was not processed " \
+                      "successfully for {}".format(nova_unit_name)
+                raise Exception("Test failed: {}".format(msg))
```
ajkavanagh, line 194: This looks a bit like the test is peeking a bit too much into the implementation of the effect being tested rather than block box testing the outcome (behaviours) of the the system-under-test.  i.e. the test probably ought to test "what it should do" with a particular setup, rather than "how is is implemented".  This would allow the implementation to change, but the behaviour to remain the same, which is what is important.  The issue with testing implementation at this testing level, is it 'locks in' the implementation.

I think the purpose is to check that the "nova" pool gets created, which is the purpose of the next test, but there may be something that I don't understand or I'm missing.

zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -141,6 +142,107 @@ def test_930_check_virsh_default_network(self):
...
+        logging.info('test_1001_config_changed_cause_restarts successful')
+
+    def test_1002_ceph_relation_changed_cause_restarts(self):
+        """Verify service is not restarted on ceph-relation-changed hook."""
+        backend = zaza.model.get_application_config(
+            self.application_name)['libvirt-image-backend']
+        rbd_enabled = backend['value'].lower() == "rbd"
+        if not rbd_enabled:
+            raise unittest.SkipTest(
+                'RBD backend is not enabled, skipping test')
```
ajkavanagh, line 220: This is repeated above, so can probably become a predicate helper in the class?

zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -141,6 +142,107 @@ def test_930_check_virsh_default_network(self):
...
+        if "nova" not in pools:
+            msg = "Nova ceph pool was not created successfully"
+            raise Exception("Test failed: {}".format(msg))
+
+        logging.info('test_1000_ceph_broker_message_marked_done successful')
+
+    def test_1001_config_changed_cause_restarts(self):
+        """Verify service is not restarted on config-changed hook."""
+        self._test_hook_cause_restarts('config-changed')
+        logging.info('test_1001_config_changed_cause_restarts successful')
```
ajkavanagh, line 211: I'm a bit dubious about this test; it's forcing a config-changed hook run by calling the actual hook, rather than doing some config change and recording whether the proposed thing change.

I think the test is probably fragile.  e.g. if nova-compute were optimised such that no actual config change caused a restart regardless of whether the actual hook is called, then this test would fail -- it's assuming implementation rather than behaviour.

There is a config_changed context manager (`config_change()`) which can be used, or even the `restart_on_changed_debug_oslo_config_file()` helper which can check for restarts.  Please review the `charm_tests/test_utils.py` for a suitable helper?

zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -141,6 +142,107 @@ def test_930_check_virsh_default_network(self):
...
+    def test_1002_ceph_relation_changed_cause_restarts(self):
+        """Verify service is not restarted on ceph-relation-changed hook."""
+        backend = zaza.model.get_application_config(
+            self.application_name)['libvirt-image-backend']
+        rbd_enabled = backend['value'].lower() == "rbd"
+        if not rbd_enabled:
+            raise unittest.SkipTest(
+                'RBD backend is not enabled, skipping test')
+
+        self._test_hook_cause_restarts('ceph-relation-changed')
```
ajkavanagh, line 222: Again, I'm dubious of forcing this in this way, by calling a hook.  It would be better to get Juju to call the hook, by changing something in ceph.

rodrigogansobarbieri:
Hi Alex. Thanks for the feedback. I'm sorry, I should have included more context to help reviewing this.

This functional test change is to validate the correctness of PR [1] 

Indeed, I went a bit too far on test_1000, checking if the pool is created should be enough. I will update this.

Regarding test 1001, it is to address the problem caused by bug [2], which motivated PR [1]. Changing a debug option should cause a restart. However, the test is validating that that invoking a config-changed hook *does not* cause a restart. Previously it used to, but after fix [3] the bug was addressed (without a functional test). I think the main value of this test is to capture if the nova-compute service is restarting multiple times for no reason. Before, it was due to a bug in the code that was explicitely invoking a restart. Please let me know what you think after this clarification.

I coded those tests a long time ago and now I realize that Test 1002 shouldn't be invoking ceph-relation-changed without a remote context (and to do that manually is ugly-ish [4]), so yes, I will see to updating the test to change something in ceph to trigger it, if possible. 

[1] https://github.com/juju/charm-helpers/pull/363

[2] https://bugs.launchpad.net/charm-nova-compute/+bug/1835045

[3] https://review.opendev.org/#/c/676407

[4] https://discourse.juju.is/t/how-to-run-any-hook-with-a-remote-context/2789

zaza/openstack/charm_tests/nova/tests.py
rodrigogansobarbieri: test removed

zaza/openstack/charm_tests/nova/tests.py
rodrigogansobarbieri: refactored into a helper test (although no longer needed)

zaza/openstack/charm_tests/nova/tests.py
rodrigogansobarbieri: removed that and now only checks if the pool has been created

zaza/openstack/charm_tests/nova/tests.py
rodrigogansobarbieri: As I stated in my comment, I believe this test is valuable. Please check my comment with extra context and let me know what you think.

rodrigogansobarbieri:
I just pushed a new change, CI has run successfully on https://review.opendev.org/#/c/729768

Test 1002 has been removed and test 1000 no longer checks messaging details, only the pool.

Waiting on reviews now.

ajkavanagh:
@ChrisMacNaughton Is this still needed/desirable?  I didn't merge it 2 years ago, but we need to decide whether to keep this or not.  Adding stale label to track progress.

PR #294, opened 2020-05-26
Enable guest restart test on ussuri

gnuoy:
As documented in the 20.05 release note the guest restart function
of masakari is currently supported on ussuri+ so enable the test
for bionic-ussuri+


gnuoy:
@ajkavanagh I think that is an excellent idea: https://review.opendev.org/730739

PR #306, opened 2020-06-04
Fix hacluster tests

gnuoy:
Tidy up hacluster tests. In the process this fixes the error:

```
UnboundLocalError: local variable 'primary_status' referenced before assignment
```

This was caused by libjuju now returning an empty dict rather than
None when listing a subordinates units.

Depends-On: https://github.com/openstack-charmers/zaza/pull/364


gnuoy:
Tested here: https://review.opendev.org/#/c/732645/

PR #315, opened 2020-06-10
Migrate Manila Generic to Zaza

alejandro-sg:


zaza/openstack/charm_tests/manila_generic/tests.py
```diff
@@ -0,0 +1,65 @@
...
+        remote_ip = remote_unit.public_address
+        relation = zaza_juju.get_relation_from_unit(
+            unit_name,
+            remote_unit_name,
+            relation_name
+        )
+        # Get private-address in relation
+        rel_private_ip = relation.get('private-address')
+        # The private address in relation should match manila/0 address
+        self.assertEqual(rel_private_ip, remote_ip)
```
fnordahl, line 65: We typically do not want to port these types of tests over from Amulet. The problem with them is that they introspect implementation details that are subject to change over time and that makes them brittle and increases our maintenance burden.

What we do want instead is a black-box type test that consumes the actual service the charm provides to validate that it is working correctly.

zaza/openstack/charm_tests/manila_generic/tests.py
```diff
@@ -0,0 +1,65 @@
...
+    """Encapsulate Manila Generic tests."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Run class setup for running tests."""
+        super(ManilaGenericTests, cls).setUpClass()
+        cls.manila_client = manilaclient.Client(
+            session=cls.keystone_session, client_version='2')
+
+    def test_manila_api(self):
```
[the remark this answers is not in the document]
alejandro-sg: Done as suggested

zaza/openstack/charm_tests/manila_generic/tests.py
alejandro-sg: Understood, working on more relevant functional tests to add

ryan-beisner: Let's not add relation introspection tests, agree with Frode.

PR #321, opened 2020-06-12
Add setup for glance-simplestreams-sync

javacruft:
Use action to complete initial image sync for the gss charm.

This avoids races where the images end up in the wrong locations
and allows the tests to actually know when images should be
discoverable.

Update tests to wait for at least four images (20.04 is synced
by default).


thedac:
Actually, this can land now as nothing is calling it.

PR #324, opened 2020-06-15
Move tempest workspace to user's home directory

coreycb:
Prior to this change the tempest workspace was stored
in the current working directory. Not having the tempest
workspace in a single location was proving to be difficult.
The directory is referenced in both setup.py and tests.py
and there's not a good way to share the location between
the two. Additionally, ~/.tempest/workspace.yaml could get
out of sync with the tempest workspace directory location
causing tempest to not know where the workspace directory
was located. Having a well known location allows us to
completely recreate the workspace directory on each run.
The workspace directory will be put in the existing
~/.tempest directory that tempest already creates.

### DIFF
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -794,6 +794,50 @@ class CephPrometheusTest(unittest.TestCase):
             '3', _get_mon_count_from_prometheus(unit.public_address))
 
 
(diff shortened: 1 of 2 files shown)


coreycb:
Good point and thanks for the comment. I need to think about how to do that without getting a leak of creating a directory/workspace on each run. I think we wouldn't want to clean it up at the end of each test run.

PR #326, opened 2020-06-18
Fix racy swift global replication testing

afreiberger:
During some runs of functional testing, swift global replication would
only have 2 copies rather than 3 of the object that was just written.

This patch attempts to resolve the race condition and also adds ordering
to more recently added tests.

Closes-Bug: #1882247


afreiberger:
non-obvious from the implementation, but the resolution to the race condition is waiting for the region2 model to settle before uploading the object in either test to ensure the rings have been completely distributed.  I accomplished this block_until by making the configuration changes in both models' proxy configurations, which is necessary anyways, to affect write_affinity for the second region.

PR #336, opened 2020-06-22
Add functional tests for neutron-arista

AurelienLourot:
Proving that I didn't break existing tests: https://review.opendev.org/#/c/737277/

### DIFF
--- a/zaza/openstack/charm_tests/neutron/tests.py
+++ b/zaza/openstack/charm_tests/neutron/tests.py
@@ -293,8 +293,23 @@ class NeutronCreateNetworkTest(test_utils.OpenStackBaseTest):
         cls.neutron_client = (
             openstack_utils.get_neutron_session_client(cls.keystone_session))
 
+    def _test_400_additional_validation(self, expected_network_names):
+        """Additional assertions for test_400_create_network.
+
+        Can be overridden in derived classes.
+
+        :type expected_network_names: List[str]
+        """
+        pass
+
     def test_400_create_network(self):
-        """Create a network, verify that it exists, and then delete it."""
+        """Create a network, verify that it exists, and then delete it.
+
+        Additional verifications on the created network can be performed by
+        deriving this class and overriding _test_400_additional_validation().
+        """
+        self._test_400_additional_validation([])
+
         logging.debug('Creating neutron network...')
         self.neutron_client.format = 'json'
         net_name = 'test_net'
@@ -319,9 +334,13 @@ class NeutronCreateNetworkTest(test_utils.OpenStackBaseTest):
         network = networks['networks'][0]
         assert network['name'] == net_name, "network ext_net not found"
 
+        self._test_400_additional_validation([net_name])
+
         # Cleanup
         logging.debug('Deleting neutron network...')
         self.neutron_client.delete_network(network['id'])
+
+        self._test_400_additional_validation([])
 
 
 class NeutronApiTest(NeutronCreateNetworkTest):
--- a/zaza/openstack/charm_tests/neutron/tests.py
+++ b/zaza/openstack/charm_tests/neutron/tests.py
@@ -292,55 +292,44 @@ class NeutronCreateNetworkTest(test_utils.OpenStackBaseTest):
         # set up clients
         cls.neutron_client = (
             openstack_utils.get_neutron_session_client(cls.keystone_session))
+        cls.neutron_client.format = 'json'
 
-    def _test_400_additional_validation(self, expected_network_names):
-        """Additional assertions for test_400_create_network.
-
-        Can be overridden in derived classes.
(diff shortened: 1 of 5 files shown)


zaza/openstack/charm_tests/neutron/tests.py
```diff
@@ -293,8 +293,23 @@ def setUpClass(cls):
...
+
+        Can be overridden in derived classes.
+
+        :type expected_network_names: List[str]
+        """
+        pass
+
     def test_400_create_network(self):
-        """Create a network, verify that it exists, and then delete it."""
+        """Create a network, verify that it exists, and then delete it.
```
ajkavanagh, line 306: This feels a bit awkward, but I can't quite put my finger on it.  What I'm referring to is the introduction of `self._test_400_additional_validation()` at three points in this test, as a form of extending the test and allowing overrides.

I'll try to explain:

 * It seems to be specializing the `_test_400_additional_validation()` using parameter, but it's not really clear how it's supposed to be used.
 * With this kind of adding functionality would normally be done with a `super()` call, but obviously there are 3 points of change.

It's not very helpful for me to just say: "Well, I don't like that for 'reasons'"!  So I'll try and explain what I might do differently: I think there are two options:

1. I'd split the contents of `test_400_create_network()` into 3 sub function helpers, and call those from the test.  Then either those functions can be 'wrapped' (overridden) or `test_400_create_network` can be overridden, and insert the helper validation as well as calling the 3 sub functions.
2. As `test_400_create_network()` is small, I'd probably just override the whole function in the using test, and just add in the bits that are needed; yes, there's code duplication, but there's not a lot, and it's obvious.  However, if bugs are found, they have to be fixed in two places.

zaza/openstack/charm_tests/neutron_arista/setup.py
```diff
@@ -0,0 +1,39 @@
...
+import zaza
+import zaza.openstack.charm_tests.neutron_arista.utils as arista_utils
+
+
+def test_fixture():
+    """Pass arista-virt-test-fixture's IP address to neutron-arista."""
+    fixture_ip_addr = arista_utils.fixture_ip_addr()
+    logging.info(
+        "{}'s IP address is '{}'. Passing it to neutron-arista...".format(
+            arista_utils.FIXTURE_APP_NAME, fixture_ip_addr))
```
ajkavanagh, line 28: Could I suggest a slightly different formatting option for consideration?

```python
    logging.info(
        "{}'s IP address is '{}'. Passing it to neutron-arista..."
        .format(arista_utils.FIXTURE_APP_NAME, fixture_ip_addr))
```

It makes the `format` a little easier to read (for me?), but may just be my preference.

zaza/openstack/charm_tests/neutron_arista/tests.py
```diff
@@ -0,0 +1,53 @@
...
+
+        logging.info('Waiting for Neutron to become ready...')
+        for attempt in tenacity.Retrying(
+                wait=tenacity.wait_fixed(5),  # seconds
+                stop=tenacity.stop_after_attempt(12),
+                reraise=True):
+            with attempt:
+                cls.neutron_client.list_networks()
+
+    def _test_400_additional_validation(self, expected_network_names):
```
ajkavanagh, line 41: See earlier comment, but note that you need to now go and look at the other class to find out what the test_400 is for?

zaza/openstack/charm_tests/neutron_arista/utils.py
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    response = session.post(
+        'https://{}/command-api/'.format(ip_addr),
+        data=json.dumps(data),
+        timeout=10  # seconds
+    )
+
+    result = []
+    for _, region in response.json()['result'][0]['regions'].items():
```
ajkavanagh, line 63: You can use `.values()` instead of `.items()`:

```python
for region in response.json()[].values():
...
```

zaza/openstack/charm_tests/neutron_arista/setup.py
```diff
@@ -0,0 +1,71 @@
...
+            logging.info('Downloading Arista image from {}'
+                         .format(os.environ['TEST_ARISTA_IMAGE_REMOTE']))
+            openstack_utils.download_image(
+                os.environ['TEST_ARISTA_IMAGE_REMOTE'],
+                os.environ['TEST_ARISTA_IMAGE_LOCAL'])
+    except KeyError:
+        pass
+
+    logging.info('Arista image can be found at {}'
+                 .format(os.environ['TEST_ARISTA_IMAGE_LOCAL']))
```
ajkavanagh, line 52: This isn't true if the above try/except failed?  (or i've misunderstood something here?)

PR #343, opened 2020-07-02
Cinder SecurityTests: support keystone over TLS

auria:


auria:
This change is part of https://github.com/openstack-charmers/zaza-openstack-tests/pull/372

I'll go ahead and close this change.

PR #355, opened 2020-07-09
Fix paramiko.ssh_exception.BadAuthenticationType

AurelienLourot:
when SSHing to a new Ubuntu instance.

Note that paramiko still has a few issues around authentication:
https://github.com/paramiko/paramiko/pull/1106/files

This paramiko PR also shows that password='' isn't the same as
password=None


AurelienLourot:
CI runs: https://review.opendev.org/732652

PR #362, opened 2020-07-14
Add Groovy Victoria support

gnuoy:


zaza/openstack/utilities/os_versions.py
```diff
@@ -106,6 +109,8 @@
         ['2.22.0']),
     ('ussuri',
         ['2.24.0']),
```
coreycb, line 111: Not related to this change but ussuri should have: ['2.24.0', '2.25.0']

zaza/openstack/utilities/os_versions.py
```diff
@@ -169,6 +178,7 @@
         ('15', 'stein'),
         ('16', 'train'),
         ('17', 'ussuri'),
```
coreycb, line 180: Also not related but ussuri should be '18'

PR #364, opened 2020-07-14
Image architecture settings

andrewdmcleod:
Use env var image_arch to specify architecture for the image we are pulling down by default. Also set custom image properties for arm64 and ppc64el

Todo: tests and documentation of usage


zaza/openstack/charm_tests/glance/setup.py
```diff
@@ -87,11 +93,20 @@ def add_lts_image(glance_client=None, image_name=None, release=None):
     :param release: Name of ubuntu release.
     :type release: str
     """
+    image_arch = os.environ.get('image_arch')
```
gnuoy, line 96: Tox will only allow env variable with a certain perfix to be passed through so image_arch won't be present even if you set it before running the test. People have the choice of setting deployment context in a config file or in environment variables. To maintain this support please use `get_deployment_context` rather than accessing os.environ directly.

Something like:

```
import zaza.utilities.deployment_env as deployment_env


deploy_ctxt = deployment_env.get_deployment_context()
image_arch = deploy_ctxt.get('TEST_IMAGE_ARCH', 'amd64'))
```

zaza/openstack/charm_tests/glance/setup.py
```diff
@@ -87,11 +93,20 @@ def add_lts_image(glance_client=None, image_name=None, release=None):
...
+    if not image_arch:
+        image_arch = "amd64"
+    logging.info("Image architecture set to {}".format(image_arch))
+    if image_arch == "arm64":
+        properties = {'hw_firmware_type': 'uefi'}
+    if image_arch == "ppc64el":
+        properties = {'architecture': 'ppc64'}
```
gnuoy, line 103: imo it's neater to use a dict here rather than multiple `if` statements:

```
arch_image_properties = {
    'arm64': {'hw_firmware_type': 'uefi'},
    'ppc64el' {'architecture': 'ppc64'}}
properties = properties or arch_image_properties.get(image_arch)
```

PR #365, opened 2020-07-15
Add OVS to OVN migration tests

fnordahl:


AurelienLourot (changes requested):
The second commit says "Make tearDown a class instance method." but it doesn't seem to be doing that anymore?

fnordahl:
> The second commit says "Make tearDown a class instance method." but it doesn't seem to be doing that anymore?

That part got merged already, I'll update the commit message.
https://github.com/openstack-charmers/zaza-openstack-tests/blob/334304e05b35bf632040faf688ded03e613cace5/zaza/openstack/charm_tests/test_utils.py#L113-L119

zaza/openstack/charm_tests/ovn/setup.py
```diff
@@ -0,0 +1,130 @@
...
+        }
+        apps = ('neutron-gateway', 'neutron-openvswitch')
+        for app in apps:
+            try:
+                zaza.model.get_application(app)
+                for k, v in cfg.items():
+                    logging.info('Setting `{}` to "{}" on "{}"...'
+                                 .format(k, v, app))
+                with self.config_change(cfg, cfg, app):
+                    continue
```
AurelienLourot, line 69: I don't understand the intent here. `self.config_change()` is a context manager that yields after having changed the config and before having restored it. So in between we will now execute `continue`. Will that skip the restoring? Or will that be like saying `pass`? 

I decided to try it out:

```
@contextlib.contextmanager
def config_changed():
    print('changing config')
    yield
    print('restoring config back')

for i in (1, 2):
    with config_changed():
        continue
```

results in

```
changing config
restoring config back
changing config
restoring config back
```

So in fact here `continue` behaves like `pass`. I myself learned something new :) But this wasn't your intent either, correct?

AurelienLourot (changes requested):
Not done reviewing yet, but getting there :)

zaza/openstack/charm_tests/ovn/setup.py
fnordahl: The intent is to change the configuration and not change it back and that is accomplished by passing in the same Dict to both the default and alternate parameters. And the code does indeed work as you see on the results.

The reason for using the `config_change` helper to do this is that it already deals with the situation where the config already being set etc. which makes this work regardless of what configuration is passed in in the test bundle.

AurelienLourot: sorry I had something else wrong in mind (some mix-up), so the point I raised didn't make much sense I realize now. Thanks for your patience :)

fnordahl: No worries I welcome the scrutiny, and the fact that it is not obvious to the reviewer may call for me adding a comment. Your assumption will probably be shared by anyone else looking at that code as the normal use for the helper is to change and change back.

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -71,3 +76,231 @@ def setUpClass(cls):
...
+
+    def test_ovs_ovn_migration(self):
+        """Test migration of existing Neutron ML2+OVS deployment to OVN.
+
+        The test should be run after deployment and validation of a legacy
+        deployment combined with subsequent run of a network connectivity test
+        on instances created prior to the migration.
+        """
+        # The setUp method of this test class will perform the migration steps.
+        # The tests.yaml is programmed to to further validation after the
```
AurelienLourot, line 291: `s/to to/to do/`

AurelienLourot (changes requested):
Apart from a typo, this all looks good to me, thanks!

AurelienLourot (approved):

zaza/openstack/charm_tests/ovn/setup.py
```diff
@@ -0,0 +1,139 @@
...
+                                 .format(k, v, app))
+                with self.config_change(cfg, cfg, app):
+                    # The intent here is to change the config and not restore
+                    # it. We accomplish that by passing in the same value for
+                    # default and alternate.
+                    #
+                    # The reason for using the `config_change` helper for this
+                    # is that it already deals with all the permutations of
+                    # config already being set etc and does not get into
+                    # trouble if the test bundle already have the values we try
```
thedac, line 76: s/have/has

zaza/openstack/charm_tests/ovn/setup.py
```diff
@@ -0,0 +1,139 @@
...
+            try:
+                app_cfg = zaza.model.get_application_config(app)
+                ovn_cfg['bridge-interface-mappings'] = app_cfg[
+                    'data-port']['value']
+                ovn_cfg['ovn-bridge-mappings'] = app_cfg[
+                    'bridge-mappings']['value']
+                # Use values from neutron-gateway when present, otherwise use
+                # values from neutron-openvswitch
+                break
+            except KeyError:
```
thedac, line 97: You have used this pattern multiple times now. It might make sense to create a method that takes in a list of applications and a list of config parameters and returns a dictionary of config data.

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -71,3 +76,231 @@ def setUpClass(cls):
...
         ]
+
+
+class OVSOVNMigrationTest(test_utils.BaseCharmTest):
+    """OVS to OVN migration tests."""
+
+    def setUp(self):
+        """Perform migration steps prior to validation."""
+        super(OVSOVNMigrationTest, self).setUp()
+        # These steps here due to them having to be executed once and in a
```
thedac, line 87: These steps *are* here

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -71,3 +76,231 @@ def setUpClass(cls):
...
+                action_params={
+                    'i-really-mean-it': True},
+                raise_on_failure=True,
+            )
+        )
+
+    # Charm readiness is no guarantee for API being ready to serve requests.
+    # https://bugs.launchpad.net/charm-neutron-api/+bug/1854518
+    @tenacity.retry(wait=tenacity.wait_exponential(min=5, max=60),
+                    reraise=True, stop=tenacity.stop_after_attempt(3))
```
thedac, line 231: Do we really want to re-run the action multiple times? Shouldn't we block and wait on some other indicator?

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -71,3 +76,231 @@ def setUpClass(cls):
...
+    def test_ovs_ovn_migration(self):
+        """Test migration of existing Neutron ML2+OVS deployment to OVN.
+
+        The test should be run after deployment and validation of a legacy
+        deployment combined with subsequent run of a network connectivity test
+        on instances created prior to the migration.
+        """
+        # The setUp method of this test class will perform the migration steps.
+        # The tests.yaml is programmed to do further validation after the
+        # migration.
```
thedac, line 292: Interesting. I had not considered the implications of this. We might want to execute from a method other than setUp.

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -71,3 +76,231 @@ def setUpClass(cls):
...
+
+        # Reset the n-gw and n-ovs instance-mtu configuration option so it does
+        # not influence how further tests are executed.
```
thedac, line 295: If all the work is done prior to this, is this step necessary?

zaza/openstack/charm_tests/test_utils.py
```diff
@@ -426,6 +427,50 @@ def pause_resume(self, services, pgrep_full=False):
...
+        method or function.
+
+        Example:
+           # In tests.yaml:
+           tests_options:
+             zaza.charm_tests.noop.tests.NoopTest.test_foo.key: true
+           # called from zaza.charm_tests.noop.tests.NoopTest.test_foo()
+           >>> get_my_tests_options('key')
+           True
+
```
thedac, line 443: I *think* this is a new introduction of tests_options. I'd like a second opinion on this. Let's discuss in the stand-up.

thedac (changes requested):
My apologies for taking so long to review.  But as you might imaging I have several requests.

There are a handful of missing docstrings.

I would also like to get confirmation that regular OVS testing has been run against this PR as some of the common tests are affected.

I would like to get a discussion on the use of test_options.

Thank you for all the comments which made reading the code much easier and made a complex task understandable.

zaza/openstack/charm_tests/ovn/tests.py
fnordahl: The action is idempotent so it is safe to run multiple times. To avoid the tenacity we would have to fix the issue we have in all our charms where when the charm is ready the API/service it deploys often is not, I would love for us to fix that but I think that is out of scope for this work.

zaza/openstack/charm_tests/ovn/tests.py
fnordahl: Since the steps have to occur in a fixed order before we can do anything else we can either do it in setUp which gives us the ability to add more tests if we want to, or we run it in a test method and the test class cannot be used for anything else. I chose to have it in setUp so that we could potentially add more test methods if we want to.

zaza/openstack/charm_tests/ovn/tests.py
fnordahl: Yes, after the migration is done we call [other tests](https://review.opendev.org/#/c/740974/18/tests/tests.yaml@63) to confirm that instances deployed prior to migration still have connectivity. That test would be influenced by the `instance-mtu` configuration option, and we don't want that.

zaza/openstack/charm_tests/test_utils.py
fnordahl: This uses the existing tests_options functionality, the problem it solves is that the tests_options provides a flat namespace for configuration keys. I want to use the [module path of the test to avoid collisions](https://review.opendev.org/#/c/740974/18/tests/tests.yaml@68). Instead of hard coding that into the test I add a re-usable helper to do it for me.

zaza/openstack/charm_tests/ovn/setup.py
fnordahl: Encapsulating patterns in helper functions is a good thing, but while they look similar all three uses here are different.

This section blindly sets configuration on all listed apps whether they are present or not:
https://github.com/openstack-charmers/zaza-openstack-tests/blob/5a10779e4533cb38d42a15f01923789769aeb683/zaza/openstack/charm_tests/ovn/setup.py#L61-L80

This section consciously sets configuration on _one_ specific app from a set of alternatives and fails if none of them are there:
https://github.com/openstack-charmers/zaza-openstack-tests/blob/5a10779e4533cb38d42a15f01923789769aeb683/zaza/openstack/charm_tests/ovn/setup.py#L105-L121

This section _retrieves_ and stores configuration from the listed apps:
https://github.com/openstack-charmers/zaza-openstack-tests/blob/5a10779e4533cb38d42a15f01923789769aeb683/zaza/openstack/charm_tests/ovn/setup.py#L87-L103

fnordahl: I collapsed the two first use cases into a helper method that takes an argument to choose operation mode.

fnordahl:
> I would also like to get confirmation that regular OVS testing has been run against this PR as some of the common tests are affected.

Yes, the [neutron-gateway](https://review.opendev.org/#/c/750287/), [neutron-openvswitch](https://review.opendev.org/#/c/740974/) and [neutron-api-plugin-ovn](https://review.opendev.org/#/c/750607/) have or will be tested with this PR both for regular and OVN migration jobs.

zaza/openstack/charm_tests/ovn/setup.py
```diff
@@ -0,0 +1,151 @@
...
+                            continue
+                else:
+                    if first_match_raise_if_none_found:
+                        raise RuntimeError(
+                            'None of the expected apps ({}) are present in '
+                            'the model.'
+                            .format(apps)
+                        )
+            except KeyError:
+                pass
```
fnordahl, line 86: This caused a regression, putting back to draft for now

zaza/openstack/charm_tests/ovn/setup.py
thedac: Got it. I had a superficial view. No need to collapse them if that is not beneficial.

zaza/openstack/charm_tests/test_utils.py
thedac: As discussed the developers agree this works.

zaza/openstack/charm_tests/ovn/tests.py
thedac: ACK

zaza/openstack/charm_tests/ovn/tests.py
thedac: OK

zaza/openstack/charm_tests/ovn/tests.py
thedac: Glad we have a bug for neutron-api that is a bug. But agree we can work around it here.

zaza/openstack/charm_tests/ovn/setup.py
fnordahl: This actually worked, it was me forgetting to put the configure job into tests.yaml for the neutron-gateway charm ;)

thedac (approved):
Landing.

PR #369, opened 2020-07-21
HaclusterScaleBackAndForthTest

auria:


zaza/openstack/charm_tests/hacluster/tests.py
```diff
@@ -69,3 +70,71 @@ def test_920_put_in_maintenance(self):
...
+    def test_930_scaleback_bionic(self):
+        """Remove a unit, recalculate quorum and add a new one."""
+        principle_app = 'keystone'
+        principle_units = zaza.model.get_status().applications[
+            principle_app]['units']
+        self.assertEqual(len(principle_units), 3)
+        doomed_principle = sorted(principle_units.keys())[0]
+        series = juju_utils.get_machine_series(
+            principle_units[doomed_principle].machine)
+        if series != 'bionic':
```
dshcherb, line 83: My gut feel is that it shouldn't be hard-coded here and instead this should be handled at the tests.yaml level for charm-hacluster.

So what you could do is:

* create a separate test case class deriving from `test_utils.OpenStackBaseTest` which would only contain test_930_scaleback_bionic (such as `class HaclusterScalebackTest(test_utils.OpenStackBaseTest)`);
* Instead of using a default model alias, you could use a separate one and use model-specific configure and test steps (see the links below).
  * the model-specific test step for the bionic bundle would contain HaclusterScalebackTest, while the test steps for the default alias would not.

https://github.com/openstack-charmers/zaza/blob/d341c246ebdf8fe6405f643ee9ccc66f5e5fe6ab/zaza/charm_lifecycle/utils.py#L36-L48
https://github.com/openstack-charmers/zaza/blob/d341c246ebdf8fe6405f643ee9ccc66f5e5fe6ab/zaza/charm_lifecycle/utils.py#L257-L261
https://github.com/openstack-charmers/zaza/blob/d341c246ebdf8fe6405f643ee9ccc66f5e5fe6ab/zaza/charm_lifecycle/utils.py#L208-L215
https://github.com/openstack-charmers/zaza/blob/d341c246ebdf8fe6405f643ee9ccc66f5e5fe6ab/zaza/charm_lifecycle/utils.py#L311-L315

AurelienLourot:
CI runs: https://review.opendev.org/#/c/741592

AurelienLourot:
This is now a superset of #399. Let's land #399 first and rebase on top of it.

AurelienLourot:
Actually it would be better to create a new test class so that this change remains backward-compatible with former charm releases

thedac:
Warning: Running mysql-router tests with this PR I saw it "pass" when the keystone units were in a blocked state: 
keystone/2*                  blocked   idle   3        192.168.254.89  5000/tcp           Datab

Due to
https://github.com/openstack/charm-keystone/blob/master/hooks/keystone_hooks.py#L422
and https://github.com/openstack/charm-keystone/commit/a87b1011066650dbfc924501202d7048f6f022bc

Just a warning that we need to test for that condition and definitely not pass when it does. I did not get the opportunity to test this version.

AurelienLourot:
thedac's last comment is still to be addressed. All other previous comments have been addressed.

AurelienLourot:
@thedac validated locally: I ran the charm-mysql-router functional tests with the current state of this func-test-pr (just omitting running the `update-ring` action as it hasn't landed yet) and indeed the keystone units remain blocked with `Database not initialised` and the test then fails on this as this state isn't whitelisted. 

So in other words, the test is "good" at failing when something is wrong. So I think your comment has been addressed meanwhile and as far as I'm concerned, this PR is safe to land (but I'm co-author and can't land it myself).

AurelienLourot:
Travis has failed on python 3.6 at installing `cryptography` because it now requires a Rust compiler. 
https://mail.python.org/pipermail/cryptography-dev/2021-January/001003.html
Looking for a solution

lourot (approved):

zaza/openstack/charm_tests/hacluster/tests.py
```diff
@@ -130,3 +134,67 @@ def test_930_scaleback(self):
...
+        logging.info('Removing {}'.format(doomed_principle_unit))
+        zaza.model.destroy_unit(
+            self._principle_app_name,
+            doomed_principle_unit,
+            wait_disappear=True)
+        logging.info('OK')
+
+        logging.info('Waiting for model to settle')
+        zaza.model.block_until_unit_wl_status(other_hacluster_unit, 'blocked')
+        zaza.model.block_until_unit_wl_status(other_principle_unit, 'blocked')
```
freyes, line 178: This line is expecting that the principle unit will go into blocked state when there is no quorum, although that's not happening when the tests are run:

This is using a juju status from the crashdump captured by this run https://openstack-ci-reports.ubuntu.com/artifacts/test_charm_pipeline_func_full/openstack/charm-hacluster/741592/22/8108/index.html in this patch https://review.opendev.org/c/openstack/charm-hacluster/+/741592

```
$ cat juju_status.yaml | yq --tojson eval | jq '.applications.keystone.units."keystone/1"."workload-status".current' 
"active"
$ cat juju_status.yaml | yq --tojson eval | jq '.applications.keystone.units."keystone/2"."workload-status".current' 
"active"
$ cat juju_status.yaml | yq --tojson eval | jq '.applications.keystone.units."keystone/1".subordinates."hacluster/2"."workload-status".current'  
"blocked"
$ cat juju_status.yaml | yq --tojson eval | jq '.applications.keystone.units."keystone/2".subordinates."hacluster/0"."workload-status".current'  
"blocked"
```
In percona-cluster charm there is `min-cluster-size` that will make the principle unit to go into blocked state when there are not enough units, but for keystone there doesn't seem to something similar, so this line could be dropped or change 'blocked' to 'active'.

freyes (changes requested):

zaza/openstack/charm_tests/hacluster/tests.py
AurelienLourot: This must be a change of behavior in the charm in the linked Gerrit review then, because the older test class `HaclusterScalebackTest` contains also this check and the current charm (before the review) is passing this check. Digging

AurelienLourot: Ah no I understand: `HaclusterScalebackTest` isn't part of the gate of the hacluster charm yet, good catch!

AurelienLourot: But then this contradicts https://github.com/openstack-charmers/zaza-openstack-tests/pull/369#issuecomment-775866792, where I noticed that the `keystone` units would become blocked and the whole point of this test is to prove that with the new feature, the `keystone` units will afterwards exit the blocked state.

If `keystone` doesn't go into blocked state in the first place, then this entire test doesn't prove anything. I'm now wondering if `keystone` may end up going sometimes in blocked state, sometimes not.

freyes:
On Wed, 2021-02-17 at 05:32 -0800, Aurelien Lourot wrote:
> @AurelienLourot commented on this pull request.
> 
> In zaza/openstack/charm_tests/hacluster/tests.py:
> 
> > +        zaza.model.run_action(
> +            doomed_hacluster_unit,
> +            'pause',
> +            raise_on_failure=True)
> +        logging.info('OK')
> +
> +        logging.info('Removing {}'.format(doomed_principle_unit))
> +        zaza.model.destroy_unit(
> +            self._principle_app_name,
> +            doomed_principle_unit,
> +            wait_disappear=True)
> +        logging.info('OK')
> +
> +        logging.info('Waiting for model to settle')
> +        zaza.model.block_until_unit_wl_status(other_hacluster_unit,
> 'blocked')
> +        zaza.model.block_until_unit_wl_status(other_principle_unit,
> 'blocked')
> But then this contradicts #369 (comment), where I noticed that the
> keystone units would become blocked and the whole point of this test
> is to prove that with the new feature, the keystone units will
> afterwards exit the blocked state.
> 
> If keystone doesn't go into blocked state in the first place, then
> this entire test doesn't prove anything. I'm now wondering if
> keystone may end up going sometimes in blocked state, sometimes not.

The only code that I could find that suggests keystone could go into
blocked state due to hacluster is when the vip* configuration keys
weren't set.
https://github.com/openstack/charm-keystone/blob/master/hooks/keystone_utils.py#L2373-L2379

AurelienLourot:
What I witnessed is that keystone would go into blocked state with `Database not initialised`, so it's not going directly into blocked state due to hacluster but due to the other subordinate mysql-router I think.

freyes:
On Wed, 2021-02-17 at 06:42 -0800, Aurelien Lourot wrote:
> What I witnessed is that keystone would go into blocked state with
> Database not initialised, so it's not going directly into blocked
> state due to hacluster but due to the other subordinate mysql-router
> I think.

so we could drop this line from this patch then(?)

AurelienLourot:
Clarified offline, summary: what we need is in fact a test that doesn't care about the keystone unit but instead removes **two** units (losing quorum: 1 unit out of 3) and re-adds **two** units (before the fix we still have no quorum because we end up wrongly with 3/5 units, with the fix we have properly 3/3 units)

AurelienLourot:
Validated locally, I think we're good now. I'm also validating now that without the fix in the charm, the charm wouldn't reach quorum at the end and the test would fail. I.e. I'm validating that this test would also catch a regression.

AurelienLourot:
When run against a charm that doesn't have the fix, and without running the action, the `hacluster` units at the end are all happily back `active/idle`, so removing and re-adding two units seems still not to be enough to lose quorum. `sudo crm status` reports quorum:

```
Cluster Summary:
  * Stack: corosync
  * Current DC: juju-9dc2a6-zaza-8f162ab10868-3 (version 2.0.3-4b1f869f0f) - partition with quorum
  * Last updated: Thu Feb 18 13:24:33 2021
  * Last change:  Thu Feb 18 12:53:26 2021 by hacluster via crmd on juju-9dc2a6-zaza-8f162ab10868-3
  * 6 nodes configured
  * 7 resource instances configured

Node List:
  * Online: [ juju-9dc2a6-zaza-8f162ab10868-3 juju-9dc2a6-zaza-8f162ab10868-19 juju-9dc2a6-zaza-8f162ab10868-20 ]
  * OFFLINE: [ juju-9dc2a6-zaza-8f162ab10868-1 juju-9dc2a6-zaza-8f162ab10868-2 node1 ]

Full List of Resources:
  * Resource Group: grp_ks_vips:
    * res_ks_a26d1a0_vip        (ocf::heartbeat:IPaddr2):        Started juju-9dc2a6-zaza-8f162ab10868-19
  * Clone Set: cl_ks_haproxy [res_ks_haproxy]:
    * Started: [ juju-9dc2a6-zaza-8f162ab10868-3 juju-9dc2a6-zaza-8f162ab10868-19 juju-9dc2a6-zaza-8f162ab10868-20 ]
    * Stopped: [ juju-9dc2a6-zaza-8f162ab10868-1 juju-9dc2a6-zaza-8f162ab10868-2 node1 ]
```

Still digging in order to figure out how we can write a test that is really validating the new feature.

AurelienLourot:
Added an assertion on `crm status`. Validated locally that the test still passes against a fixed charm. Now validating that the test would fail against a non-fixed charm.

AurelienLourot:
Validated that the new assertion would fail against a non-fixed charm. I think we're good now.

dshcherb (approved):

dshcherb (approved):

PR #371, opened 2020-07-23
Make neutron/arista tests more robust

AurelienLourot:
When creating an Arista-backed Neutron network, the API call from Neutron
to Arista sometimes spuriously fails. This is not fatal and Neutron
simply retries a few seconds later. The tests now account for that.


AurelienLourot:
CI runs: https://review.opendev.org/742460

PR #383, opened 2020-08-13
Use juju model as workspace and store in home

coreycb:
The current juju model will now be used as the tempest workspace name.
Additionally, all workspaces will be stored in ~/.tempest/<workspace>.

This patch also introduces a new option 'keep-workspace' that can be
specified along with other tempest options to keep the workspace after
tempest test execution. It defaults to False.

Also minor adjustment to smoke option to test boolean value.

### DIFF
--- a/zaza/openstack/charm_tests/tempest/setup.py
+++ b/zaza/openstack/charm_tests/tempest/setup.py
@@ -17,7 +17,6 @@
 import jinja2
 import urllib.parse
 import os
-import subprocess
 
 import zaza.utilities.deployment_env as deployment_env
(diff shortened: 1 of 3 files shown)


gnuoy:
Not sure if this does what you need but zaza already supplies a model specific temp dir https://github.com/openstack-charmers/zaza/blob/master/zaza/utilities/deployment_env.py#L199

coreycb:
I don't think so unless I'm missing the context. The ~/.tempest directory already gets created by tempest and seems to be the standard location for workspaces. Part of the problem was that tempest was sometimes getting confused with the default workspace being in ~/.tempest and our workspaces being in the current working directory.

PR #384, opened 2020-08-17
Additional test for Gnocchi S3 feature : adding an external cert

camille-rodriguez:
When configuring Gnocchi to use an external endpoint for an S3 backend, it can be necessary to install a root CA on the gnocchi unit to communicate to this endpoint when it is encrypted. The PR https://review.opendev.org/#/c/746562/ adds automation for this option, and these tests confirm its functionality.


zaza/openstack/charm_tests/gnocchi/tests.py
```diff
@@ -89,6 +92,41 @@ def setUpClass(cls):
...
+        config = {
+            ca_cert_option: b64_cert,
+        }
+        model.set_application_config(
+            'gnocchi',
+            config
+        )
+        model.block_until_all_units_idle()
+
+    def test_validate_cert_location(self):
```
AurelienLourot, line 110: We don't want to rely on the order of execution of the tests because we have noticed in the past that the order is not 100% guaranteed. This leads to a flaky gate that fails every now and then and it's quite hard to figure out why.

I guess what you want here is to merge all 3 test functions into one which is covering the entire scenario, but I'm also fine if you have a better idea.

AurelienLourot (changes requested):

zaza/openstack/charm_tests/gnocchi/tests.py
camille-rodriguez: Ah I did not know that, I was just striving for smaller functions. Sure, I will put them all together.

camille-rodriguez: Update pushed

AurelienLourot (approved):
Thanks! let's land this after a `charm-recheck`

PR #388, opened 2020-08-20
Fix protocol rendering for keystone URL

### LINKED ISSUE #387: Tempest test cannot run against a v2 cloud because it renders '{proto}' in the URL for Keystone

ChrisMacNaughton:
Closes #387


ChrisMacNaughton:
If we look at the v3 version of this line, we see that it matches this change: https://github.com/openstack-charmers/zaza-openstack-tests/blob/a67b4906eef7e199b2a0949f6723a6d64097f670/zaza/openstack/charm_tests/tempest/templates/tempest_v3.j2#L36

gnuoy (approved):

PR #393, opened 2020-08-20
New test for Gnocchi

camille-rodriguez:
When configuring Gnocchi to use an external endpoint for an S3 backend, it can be necessary to install a root CA on the gnocchi unit to communicate to this endpoint when it is encrypted. The PR https://review.opendev.org/#/c/746562/ adds automation for this option, and these tests confirm its functionality.

A new charm class is added for these tests, to avoid breaking any earlier version of the charm.


zaza/openstack/charm_tests/gnocchi/tests.py
```diff
@@ -110,3 +113,38 @@ def test_s3_list_gnocchi_buckets(self):
...
+        }
+        model.set_application_config(
+            'gnocchi',
+            config
+        )
+        model.block_until_all_units_idle()
+
+        cert_location = '/usr/local/share/ca-certificates'
+        cert_name = 'gnocchi-external.crt'
+        cmd = 'ls ' + cert_location + '/' + cert_name
```
thedac, line 138: Let's use block_until_file_has_contents or block_until_file_ready rather than a simple ls.

https://github.com/openstack-charmers/zaza/blob/master/zaza/model.py#L1246
https://github.com/openstack-charmers/zaza/blob/master/zaza/model.py#L1288

zaza/openstack/charm_tests/gnocchi/tests.py
```diff
@@ -110,3 +113,38 @@ def test_s3_list_gnocchi_buckets(self):
...
+        logging.info("Validating that the file {} is created in \
+                     {}".format(cert_name, cert_location))
+        result = model.run_on_unit('gnocchi/0', cmd)
+        self.assertEqual(result['Code'], '0')
+
+        linked_cert_location = '/etc/ssl/certs'
+        linked_cert_name = 'gnocchi-external.pem'
+        cmd = 'ls ' + linked_cert_location + '/' + linked_cert_name
```
thedac, line 146: Same

zaza/openstack/charm_tests/gnocchi/tests.py
camille-rodriguez: I updated the PR using block_until_file_has_contents.

PR #395, opened 2020-08-27
Add tests to check the existence of Nagios/NRPE checks

exsdev0:
This change complements the following functionality being added:
https://review.opendev.org/#/c/747105/


exsdev0:
As per the discussions on the MR in OpenDev (link in OP), the test is being going to be moved to mojo (pre focal) and charmed-openstack-tester (focal+). Closing this MR.

PR #396, opened 2020-08-28
Improve ceph proxy permissions check

javacruft:
Update the cinder-ceph permissions check to look for permissions
specific to the cinder-ceph usage of ceph; this ensures that if
we add glance and nova to the bundle (which creates additional
permissions) the existing test will continue to pass.


gnuoy (approved):
LGTM, thanks

PR #397, opened 2020-08-28
Ensure that octavia sets up application_name

ChrisMacNaughton:
When running in a deployment that is focused on bundles rather
than a specific charm, the Octavia setup can fail because there
is no configured charm_name in the tests.yaml. This change
ensures that the Octavia setup can continue.


fnordahl (approved):
LGTM, thanks!

PR #399, opened 2020-08-29
Add HaclusterScalebackTest

AurelienLourot:
This is a subset of #369.
CI runs: https://review.opendev.org/#/c/748714

### DIFF
--- a/zaza/openstack/charm_tests/hacluster/tests.py
+++ b/zaza/openstack/charm_tests/hacluster/tests.py
@@ -79,93 +79,54 @@ class HaclusterTest(HaclusterBaseTest):
 class HaclusterScalebackTest(HaclusterBaseTest):
     """hacluster scaleback tests."""
 
-    _PRINCIPLE_APP_NAME = 'keystone'
-    _HACLUSTER_APP_NAME = 'hacluster'
-    _HACLUSTER_CHARM_NAME = 'hacluster'
+    @classmethod
+    def setUpClass(cls):
+        """Run class setup for running hacluster scaleback tests."""
+        super(HaclusterScalebackTest, cls).setUpClass()
+        test_config = cls.test_config['tests_options']['hacluster']
+        cls._principle_app_name = test_config['principle-app-name']
+        cls._hacluster_charm_name = test_config['hacluster-charm-name']
 
     def test_930_scaleback(self):
-        """Remove a unit, recalculate quorum and add a new one."""
-        principle_units = zaza.model.get_status().applications[
-            self._PRINCIPLE_APP_NAME]['units']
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/hacluster/tests.py
```diff
@@ -69,3 +74,98 @@ def test_920_put_in_maintenance(self):
 
         self._toggle_maintenance_and_wait('true')
         self._toggle_maintenance_and_wait('false')
+
+
+class HaclusterScalebackTest(HaclusterBaseTest):
+    """hacluster scaleback tests."""
+
+    _PRINCIPLE_APP_NAME = 'keystone'
+    _HACLUSTER_APP_NAME = 'hacluster'
```
ChrisMacNaughton, line 83: This should really use [tests_options](https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/charm_tests/policyd/tests.py#L82) to configure the app name

zaza/openstack/charm_tests/hacluster/tests.py
```diff
@@ -69,3 +74,98 @@ def test_920_put_in_maintenance(self):
...
+                "workload-status": "blocked",
+                "workload-status-message":
+                    "Insufficient peer units for ha cluster (require 3)"
+            },
+            self._PRINCIPLE_APP_NAME: {
+                "workload-status": "blocked",
+                "workload-status-message": "Database not initialised",
+            },
+
+            # NOTE(lourot): these applications are present in the zaza bundles
```
ChrisMacNaughton, line 120: I really don't want us to have tests depend on the structure of the bundle they're deploying into as this test will mysteriously fail if we ever change that bundle (or add it to a new test bundle that doesn't exactly match this example).

zaza/openstack/charm_tests/hacluster/tests.py
```diff
@@ -69,3 +74,98 @@ def test_920_put_in_maintenance(self):
...
+
+        logging.info('Adding a hacluster unit')
+        zaza.model.add_unit(self._PRINCIPLE_APP_NAME, wait_appear=True)
+        expected_states = {
+            self._HACLUSTER_APP_NAME: {
+                "workload-status": "active",
+                "workload-status-message": "Unit is ready and clustered"
+            },
+
+            # NOTE(lourot): these applications remain waiting/blocked after
```
ChrisMacNaughton, line 147: Same as above, we should handle partial `wait_for_application_states` better if this is necessary.

ChrisMacNaughton (changes requested):

ChrisMacNaughton (approved):

PR #401, opened 2020-08-29
Retry on failures to download Arista image

AurelienLourot:
In the pre-deploy step of our neutron-arista tests we download an
Arista image from our swift bucket. It sometimes fails when our CI
system is loaded. This patch adds a tenacity retry.


AurelienLourot:
CI runs: https://review.opendev.org/#/c/748662/

fnordahl (approved):

PR #403, opened 2020-08-31
Add new keystone-ldap charm options in LDAP tests

hemanthnakkina:
New charm options for keystone-ldap are getting introduced
as part of bug [1]. The modifications to keystone-ldap are still
in-progress [2]

This patch modifies the keystone ldap configuration used in ldap
tests to use the new charm options.

[1] https://bugs.launchpad.net/charm-keystone-ldap/+bug/1832765
[2] https://review.opendev.org/#/c/748364/

### DIFF
--- a/zaza/openstack/charm_tests/keystone/tests.py
+++ b/zaza/openstack/charm_tests/keystone/tests.py
@@ -455,3 +455,38 @@ class LdapTests(BaseKeystoneTest):
                 self.assertIsNotNone(johndoe, "user 'john doe' was unknown")
                 janedoe = self._find_keystone_v3_user('jane doe', 'userdomain')
                 self.assertIsNotNone(janedoe, "user 'jane doe' was unknown")
+
+
+class LdapExplicitCharmConfigTests(LdapTests):
+    """Keystone ldap tests tests."""
+
+    def _get_ldap_config(self):
+        """Generate ldap config for current model.
+
+        :return: tuple of whether ldap-server is running and if so, config
+            for the keystone-ldap application.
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -404,6 +404,14 @@ def _get_ldap_config(self):
...
             'ldap-suffix': 'dc=test,dc=com',
             'domain-name': 'userdomain',
+            'ldap-user-tree-dn': 'ou=users,dc=test,dc=com',
+            'ldap-user-filter': '',
+            'ldap-user-objectclass': 'inetOrgPerson',
+            'ldap-user-id-attribute': 'cn',
+            'ldap-user-name-attribute': 'sn',
+            'ldap-group-tree-dn': 'ou=UserGroups',
+            'ldap-group-objectclass': 'groupOfNames',
+            'ldap-group-id-attribute': 'cn',
```
ajkavanagh, line 414: Will this break on previous revision charms if this test is run?  If so, then this will need to be used in a new test class and only called from the new version of the charm (in the tests.yaml).

hemanthnakkina: Yes, this change breaks the existing charm revisions. Thanks for the suggestion. I will update a new patch.

hemanthnakkina: Submitted a new patch with the changes

AurelienLourot (approved):
LGTM, thanks! Let's land this after code freeze.

wolsen (approved):
LGTM

PR #404, opened 2020-09-02
Adapt list of swift services on focal-victoria

AurelienLourot:
CI runs: https://review.opendev.org/#/c/748659/

### DIFF
--- a/zaza/openstack/charm_tests/swift/tests.py
+++ b/zaza/openstack/charm_tests/swift/tests.py
@@ -104,19 +104,17 @@ class SwiftStorageTests(test_utils.OpenStackBaseTest):
         services = ['swift-account-server',
                     'swift-account-auditor',
                     'swift-account-reaper',
-                    'swift-account-replicator',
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/swift/tests.py
```diff
@@ -114,6 +114,17 @@ def test_901_pause_resume(self):
                     'swift-object-replicator',
                     'swift-object-updater',
                     'swift-container-sync']
+
+        focal_victoria = openstack_utils.get_os_release('focal_victoria')
+        if self.current_os_release < focal_victoria:
+            services += ['swift-account-replicator',
+                         'swift-container-replicator',
+                         'swift-object-replicator']
```
ajkavanagh, line 122: Shouldn't these be removed from the list above?

AurelienLourot: Good catch :facepalm:

gnuoy (approved):
LGTM

PR #406, opened 2020-09-04
Fix series-upgrade tests

ajkavanagh:
A couple of changes here:

 * Ensure that the post-upgrade-hook runs BEFORE the config-changed to
   set the openstack-origin/source back to distro.  The former behaviour
   breaks keystone quite badly.
 * Ensure that the charm name is used, as discovered from the model, for
   rabbitmq-server and percona-cluster to cope with different bames for
   the application in the model vs the charm name from the charm-store.
 * Check whether the machine needs to be rebooted after the dist-upgrade
   (before the do-release-upgrade), and reboot the machine first.
   Otherwise, do-release-upgrade will fail.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 aiounittest
 async_generator
 boto3
-juju
+juju!=2.8.3  # blacklist 2.8.3 as it appears to have a connection bug
 juju_wait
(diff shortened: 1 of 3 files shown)


ChrisMacNaughton (approved):
Looks reasonable

PR #410, opened 2020-09-09
trilio: add test for ghost-share action

javacruft:
Restructure test classes a little to support addition of testing
of the ghost-share action to charms that support it.

Execute the action - if any operation fails the action will fail
so an OK return code on the action indicates success.


gnuoy (approved):
LGTM, thanks

PR #416, opened 2020-09-15
Fix network setup for mojo

ajkavanagh:
A recent change introduced a configuration option in the
./tests/test.yaml file (default location) which allows juju wait to be
used rather than waiting on various workload messages.  This,
unfortunately, breaks mojo tests as they don't use a tests.yaml.  This
change refactors that code, and enables a 'use_juju_wait' to be passed
into the relevant functions, and a new command line option (default
true) to disable using juju wait.


fnordahl (approved):
LGTM, thanks!

PR #429, opened 2020-09-25
Nova use common launch instance helper

fnordahl:


zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -38,32 +38,35 @@ def launch_instance(self, instance_key):
...
 
     def test_launch_small_instance(self):
         """Launch a Bionic instance and test connectivity."""
-        zaza.openstack.configure.guest.launch_instance(
-            glance_setup.LTS_IMAGE_NAME,
-            use_boot_volume=True)
+        self.RESOURCE_PREFIX = 'zaza-nova'
+        self.launch_guest(
+            'volume-backed-ubuntu', instance_key=glance_setup.LTS_IMAGE_NAME,
+            use_boot_volmue=True)
```
AurelienLourot, line 69: typo `s/volmue/volume/`

AurelienLourot (changes requested):
Do you have a Gerrit review making use of this as a func-test-pr? Thanks!

zaza/openstack/charm_tests/nova/tests.py
fnordahl: Thank you for catching that :+1:

fnordahl:
> Do you have a Gerrit review making use of this as a func-test-pr? Thanks!

Yes, it was used in https://review.opendev.org/#/c/752071/ which exercised both `CirrosGuestCreateTest` and `LTSGuestCreateTest` from this branch.

AurelienLourot:
@fnordahl does it make sense that it passed although there was this typo? Is that argument really necessary then?

fnordahl:
> @fnordahl does it make sense that it passed although there was this typo? Is that argument (use_boot_volume) really necessary then?

There are three tests and the referenced reviewed used two of them and not the one with the typo. That test is used somewhere else and still needed, yes.

AurelienLourot:
OK thanks, I think we should at least run it once somewhere, to be on the safe side. It doesn't need to be in a Gerrit review: I'm also fine if you tell me you ran it locally.

fnordahl:
> OK thanks, I think we should at least run it once somewhere, to be on the safe side. It doesn't need to be in a Gerrit review: I'm also fine if you tell me you ran it locally.

Took it for a spin here: https://openstack-ci-reports.ubuntu.com/artifacts/test_charm_pipeline_func_smoke/openstack/charm-nova-compute/755762/1/19451/index.html

The result was unsuccessful, but that is probably due to issues with the bundle, at least it shows the test code executing without exploding.

AurelienLourot:
> at least it shows the test code executing without exploding.

from what I can see it didn't run a single test? so it didn't your new code, right?

AurelienLourot:
ok now I see the first run. It's failing on "LTSGuestCreateVolumeBackedTest" although there is tenacity/retry, so it's hard for me to press "merge" right now :D

fnordahl:
> ok now I see the first run. It's failing on "LTSGuestCreateVolumeBackedTest" although there is tenacity/retry, so it's hard for me to press "merge" right now :D

Well as I said that is because the instance goes into ERROR state because bundle is wrong. If we in fact don't use the test anywhere we could remove it, otherwise I guess we can just merge this the next time the nova-compute gate test fails and the retry on the two jobs we actually use would have helped :)

AurelienLourot:
OK sorry, the thing I didn't know in the first place is that this test class is used nowhere. I just wanted to make sure we are not breaking the gate somewhere.

AurelienLourot (approved):

PR #431, opened 2020-09-25
Only configure private Neutron CIDR if it is configured

ChrisMacNaughton:
This change also updates the TEST_NAMESERVER to use the correct environment variable to match what is set in Ubuntu's OpenStack CI


gnuoy (approved):
LGTM

PR #432, opened 2020-09-25
Update vault tests with new status message

hemanthnakkina:
New status message is introduced for vault charm as part of
LP bug #1895577 [1]. The new status message has been added to the
list of workload status messages the test cases will wait for.

Vault charm changes: https://review.opendev.org/#/c/754267/

[1] https://bugs.launchpad.net/vault-charm/+bug/1895577

### DIFF
--- a/zaza/openstack/charm_tests/vault/tests.py
+++ b/zaza/openstack/charm_tests/vault/tests.py
@@ -153,7 +153,8 @@ class VaultTest(BaseVaultTest):
             allowed_domains='openstack.local')
 
         test_config = lifecycle_utils.get_charm_config()
-        del test_config['target_deploy_status']['vault']
+        test_config['target_deploy_status']['vault'][
+            'workload-status-message'] = 'New version of vault installed'
         zaza.model.block_until_file_has_contents(
             'keystone',
             zaza.openstack.utilities.openstack.KEYSTONE_REMOTE_CACERT,


zaza/openstack/charm_tests/vault/tests.py
```diff
@@ -153,7 +153,8 @@ def test_csr(self):
             allowed_domains='openstack.local')
 
         test_config = lifecycle_utils.get_charm_config()
-        del test_config['target_deploy_status']['vault']
+        test_config['target_deploy_status']['vault'][
+            'workload-status-message'] = 'New version of vault installed'
```
AurelienLourot, line 157: Two things:
1. this message isn't guaranteed to happen, is it? 
2. this message isn't going to happen with older versions of the charm (we want these tests to remain backwards compatible with older released versions of the charms)

Not sure how to proceed here.

AurelienLourot (changes requested):

zaza/openstack/charm_tests/vault/tests.py
hemanthnakkina: Hi

The test is backward compatible. The new status message will be added to pre-defined list of approved status messages [1].
I also verified by running this PR against charm vault master branch.

[1] https://github.com/openstack-charmers/zaza/blob/master/zaza/model.py#L1051

fnordahl: Thank you for the patch, however this is not how the list of approved status messages is used.

The authoritative source of information for expected deploy-time status messages is the `tests/tests.yaml` in the charm repository (when they diverge from the default 'active', 'Unit is ready' pattern).

During the course of executing functional tests the test code may need to adjust the expected status message to match what the functional test is doing.

In this case I don't see you adding a snap upgrade test, so the change is redundant.

hemanthnakkina: Thanks for the review comments.

The status message that gets changed in this patch is specific to execution of functional test.
Currently the vault versions are broken due to bugs in vault. Vault installs version 1.5.3 but reports as 1.5.0.
This is fixed in vault 1.5.4 but awaiting snap release for vault 1.5.4 [1].

Until vault snap  for 1.5.4 is released, the vault tests will be failed for [2] without this PR. 
Do you suggest to wait until [1] is released so that we dont need this PR (as you mentioned this change is redundant and only solving to overcome the temporary vault versions issue)

[1] https://github.com/canonical/snap-vault/issues/26
[2] https://review.opendev.org/#/c/754267/

AurelienLourot: Thanks for the explanation and I think this whole explanation should be present as a note in the code. Also please provide this information upfront when creating a pull request next time as we have no chance to guess it as reviewers. This PR gave us the illusion that it was solving https://review.opendev.org/#/c/754267/ where in reality this is a workaround for canonical/snap-vault#26 . Thanks!

And also I agree with @fnordahl I still think this should be in `tests/tests.yaml` instead.

AurelienLourot (changes requested):

hemanthnakkina:
Hi 

Sorry for not giving proper information in the commit message. 
With the snap vault released with 1.5.4 with vault mismatch fix, these changes are not required.
Will discard this PR

PR #433, opened 2020-09-28
Add tests for OVN charm NRPE checks

xavpaice:
The ovn-* charms have added NRPE service checks, this ensures that the
service check files have been added to the units as expected.

### DIFF
--- a/zaza/openstack/charm_tests/ovn/tests.py
+++ b/zaza/openstack/charm_tests/ovn/tests.py
@@ -72,7 +72,8 @@ class BaseCharmOperationTest(test_utils.BaseCharmTest):
                 'check_{}.cfg'.format(check_name)
             )
         ret = self._retry_check_commands_on_units(cmds, units)
-        logging.info(ret)
+        if ret:
+            logging.info(ret)
         self.assertIsNone(ret, msg=ret)
 
 
--- a/zaza/openstack/charm_tests/ovn/tests.py
+++ b/zaza/openstack/charm_tests/ovn/tests.py
@@ -15,11 +15,14 @@
 """Encapsulate OVN testing."""
 
 import logging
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -50,6 +62,19 @@ def test_pause_resume(self):
...
+        """Confirm that the NRPE service check files are created."""
+        units = zaza.model.get_units(self.application_name)
+        cmds = []
+        for check_name in self.nrpe_checks:
+            cmds.append(
+                'egrep -oh /usr/local.* /etc/nagios/nrpe.d/'
+                'check_{}.cfg'.format(check_name)
+            )
+        ret = self._retry_check_commands_on_units(cmds, units)
+        logging.info(ret)
```
fnordahl, line 75: This produces a test output like this on success:
```
06:38:29 2020-10-01 06:38:29 [INFO] Confirm that the NRPE service check files are created.
06:38:29 2020-10-01 06:38:29 [INFO]  ... 
06:38:31 2020-10-01 06:38:31 [INFO] None
06:38:31 2020-10-01 06:38:31 [INFO] ok
```
And that may be a bit terse and/or confusing for someone not familiar with how the test works.

Would logging only when `ret` is not None be an alternative?

fnordahl (commented):
This looks pretty good, have a comment in-line and awaiting a test result from the charms under test prior to approve/merge.

zaza/openstack/charm_tests/ovn/tests.py
xavpaice: I think that's an excellent plan.

fnordahl (approved):
LGTM, thanks!

PR #439, opened 2020-10-08
Clean up newly created instances if faulty

AurelienLourot:
CI results: https://review.opendev.org/#/c/754533

### DIFF
--- a/zaza/openstack/charm_tests/test_utils.py
+++ b/zaza/openstack/charm_tests/test_utils.py
@@ -550,21 +550,23 @@ class OpenStackBaseTest(BaseCharmTest):
         instance_key = instance_key or glance_setup.LTS_IMAGE_NAME
         instance_name = '{}-{}'.format(self.RESOURCE_PREFIX, guest_name)
 
-        instance = self.retrieve_guest(instance_name)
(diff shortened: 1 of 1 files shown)


fnordahl (commented):
The higher level launch_guest helper already removes instances before retrying creation, so my only question would be, would you be able to use that helper or is this for a different issue?

https://github.com/openstack-charmers/zaza-openstack-tests/blob/20263394c91d3be4328223b674ad9dbeab834361/zaza/openstack/charm_tests/test_utils.py#L553-L561

AurelienLourot:
@fnordahl true, good idea

AurelienLourot:
@fnordahl we're already using `launch_guest()` actually: https://openstack-ci-reports.ubuntu.com/artifacts/test_charm_pipeline_func_full/openstack/charm-manila/754533/1/7100/consoleText.test_charm_func_full_10195.txt

So my root cause analysis must be wrong. I'll dig more.

AurelienLourot:
Ah no I see, the cleanup you're mentioning is not at the right place so it did not happen. Will fix.

ChrisMacNaughton (approved):

PR #440, opened 2020-10-09
Make determination of app under test more robust

AurelienLourot:
CI runs: https://review.opendev.org/#/c/754546/


thedac (approved):
Looks good. The nova-cell-controller charm has passed its first test proving this works.

PR #441, opened 2020-10-09
Remove unused security tests

AurelienLourot:
No charm uses them anymore. Also they must be broken anyway because they don't pass a model name as argument to `model.sync_deployed()`.


AurelienLourot:
We're about to enter code freeze for the 20.10 release. Holding off.

ChrisMacNaughton (approved):

PR #445, opened 2020-10-13
Add tests for NRPE checks in aodh

xavpaice:
Adds a test to ensure that Aodh units related to nrpe get service check
files created.


ChrisMacNaughton (approved):
LGTM, thanks!

PR #446, opened 2020-10-18
octavia: Disable check for member operational status

fnordahl:
Temporarily disable this check until we figure out why
operational_status sometimes does not become 'ONLINE'
while the member does indeed work and the subsequent
retrieval of payload through loadbalancer is successful
ref LP: #1896729.

### DIFF
--- a/zaza/openstack/charm_tests/nova/setup.py
+++ b/zaza/openstack/charm_tests/nova/setup.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


AurelienLourot (approved):
I'd like to land this but the linked bug ( https://bugs.launchpad.net/charm-octavia/+bug/1896729 ) has been marked as invalid by the author himself meanwhile. What's the current status?

fnordahl:
@AurelienLourot The bug is invalid for the charm, but the upstream issue which prohibits the gate from passing is real, so I would very much like this to land too.

PR #453, opened 2020-10-22
octavia: Confirm charm status on operational failure during setup

fnordahl:


fnordahl:
Turns out the port operational status retrieved from Neutron is not accurate, so the only way to rig a failure would to have the binding operation fail. Staging that is quite complicated so I'll drop this test, but can inform that the behavior has been validated.

PR #454, opened 2020-10-22
ovn-chassis func-test to configure an invalid port

r-f-g:
- test whether the unit will end up in a blocked state after configuring
  an invalid port


ajkavanagh:
@r-f-g is this still current?  If so, please drop the stale label and resolve the conflict and we'll get it reviewed and merged.  Otherwise, feel free to ignore and we'll clean it up in a few weeks.  Thanks.

PR #456, opened 2020-11-06
Stop assuming there are two iscsi gateways

gnuoy:
The tests assume there are two iscsi gateways which is not a
safe assumption.


AurelienLourot (approved):
LGTM, thanks, happy to land on successful CI validation in https://review.opendev.org/c/openstack/charm-ceph-iscsi/+/761753

PR #458, opened 2020-11-12
Update Keystone LDAP tests

rodrigogansobarbieri:
- Reversed ldap-config-flags vs charm configs precedence
  for https://review.opendev.org/748364
- Added new functional test to validate the precedence
- Added zaza helper function to assist handling charm configs
  that do not accept "" (like booleans and ints), and reverting
  to unset values through config_change()
- Fixed zaza reset helper function that apparently never
  worked ({}.keys() returns dict_keys which is not a sequence)

### DIFF
--- a/unit_tests/charm_tests/test_utils.py
+++ b/unit_tests/charm_tests/test_utils.py
@@ -97,7 +97,7 @@ class TestBaseCharmTest(ut_utils.BaseTestCase):
             self.set_application_config.reset_mock()
         self.assertFalse(self.set_application_config.called)
         self.reset_application_config.assert_called_once_with(
-            'anApp', alterna_config.keys(), model_name='aModel')
+            'anApp', list(alterna_config.keys()), model_name='aModel')
         self.wait_for_application_states.assert_has_calls([
             mock.call(model_name='aModel', states={}),
             mock.call(model_name='aModel', states={}),
@@ -124,6 +124,66 @@ class TestBaseCharmTest(ut_utils.BaseTestCase):
         self.assertFalse(self.reset_application_config.called)
         self.assertFalse(self.wait_for_agent_status.called)
         self.assertFalse(self.wait_for_application_states.called)
+
+    def test_separate_non_string_config(self):
+        intended_cfg_keys = ['foo2', 'foo3', 'foo4', 'foo5']
+        current_config_mock = {
+            'foo2': None,
+            'foo3': 'old_bar3',
+            'foo4': None,
+            'foo5': 'old_bar5',
+        }
+        self.patch_target('config_current')
+        self.config_current.return_value = current_config_mock
+        non_string_type_keys = ['foo2', 'foo3', 'foo4']
+        expected_result_filtered = {
+            'foo3': 'old_bar3',
+            'foo5': 'old_bar5',
+        }
+        expected_result_special = {
+            'foo2': None,
+            'foo4': None,
+        }
+        current, non_string = (
+            self.target.config_current_separate_non_string_type_keys(
+                non_string_type_keys, intended_cfg_keys, 'application_name')
+        )
+
+        self.assertEqual(expected_result_filtered, current)
+        self.assertEqual(expected_result_special, non_string)
+
+        self.config_current.assert_called_once_with(
+            'application_name', intended_cfg_keys)
+
+    def test_separate_special_config_None_params(self):
+        current_config_mock = {
+            'foo1': 'old_bar1',
+            'foo2': None,
+            'foo3': 'old_bar3',
+            'foo4': None,
+            'foo5': 'old_bar5',
+        }
+        self.patch_target('config_current')
+        self.config_current.return_value = current_config_mock
+        non_string_type_keys = ['foo2', 'foo3', 'foo4']
+        expected_result_filtered = {
+            'foo1': 'old_bar1',
+            'foo3': 'old_bar3',
+            'foo5': 'old_bar5',
+        }
+        expected_result_special = {
+            'foo2': None,
+            'foo4': None,
+        }
+        current, non_string = (
+            self.target.config_current_separate_non_string_type_keys(
+                non_string_type_keys)
+        )
+
+        self.assertEqual(expected_result_filtered, current)
+        self.assertEqual(expected_result_special, non_string)
+
+        self.config_current.assert_called_once_with(None, None)
 
 
 class TestOpenStackBaseTest(ut_utils.BaseTestCase):
--- a/zaza/openstack/charm_tests/keystone/tests.py
+++ b/zaza/openstack/charm_tests/keystone/tests.py
@@ -380,7 +380,12 @@ class SecurityTests(BaseKeystoneTest):
 
 
 class LdapTests(BaseKeystoneTest):
-    """Keystone ldap tests tests."""
+    """Keystone ldap tests."""
+
+    non_string_type_keys = ('ldap-user-enabled-mask',
+                            'ldap-user-enabled-invert',
+                            'ldap-group-members-are-ids',
+                            'ldap-use-pool')
 
     @classmethod
     def setUpClass(cls):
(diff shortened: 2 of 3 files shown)


AurelienLourot (changes requested):
This PR seems to not be backward-compatible, i.e. the 20.10 charm now would fail these new tests, as illustrated here: https://review.opendev.org/#/c/763439/

And the charm change https://review.opendev.org/748364 has been merged without this func-test-pr being reviewed. Now we're a bit in the limbo as probably the charm gate is failing. I'm tempted to revert the charm change. Raising to the team for wider discussion @wolsen @rodrigogansobarbieri @freyes

AurelienLourot:
This is probably simple to solve by the way as the charm change moves from LdapTests to LdapExplicitCharmConfigTests as test class but somehow the func-test-pr seems to make slight adaptions to the former LdapTests that break it for 20.10.

"emergency" charm-keystone-ldap revert: https://review.opendev.org/763541

rodrigogansobarbieri:
I just pushed a new update. It is probably fixed now. It was a simple test logic mistake.

AurelienLourot:
Validating against:
- 20.10: https://review.opendev.org/c/openstack/charm-keystone-ldap/+/763439
- not-yet-reverted master: https://review.opendev.org/c/openstack/charm-keystone-ldap/+/763756

zaza/openstack/charm_tests/test_utils.py
```diff
@@ -154,6 +154,36 @@ def setUpClass(cls, application_name=None, model_alias=None):
...
+        :param config_keys: iterable of strs to index into the current config.
+                            If None, return all keys from the config
+        :type config_keys:  Optional[Iterable[str]]
+        :param special_type_keys: list of keys to be separated out
+        :type special_type_keys: list
+        :param application_name: String application name for use when called
+                                 by a charm under test other than the object's
+                                 application.
+        :type application_name:  Optional[str]
+        :return: Dictionary of previous charm configs applied without the
```
AurelienLourot, line 171: I also don't understand this (I swear I read it five times and I was very focused). What does it mean to "apply charm configs" and what are "previous" configs and what does it mean to apply them "without the special type keys"?

AurelienLourot (changes requested):
Aborting the review at this point because this might be me lacking knowledge. I'll ping someone else from the team, thanks!

zaza/openstack/charm_tests/test_utils.py
AurelienLourot: I mean if it was a "config_changed" context manager I would understand concepts like "applying" or "previous" at least but isn't it just a config getter?

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -380,7 +380,12 @@ def test_security_checklist(self):
 
 
 class LdapTests(BaseKeystoneTest):
-    """Keystone ldap tests tests."""
+    """Keystone ldap tests."""
+
+    special_type_keys = ('ldap-user-enabled-mask',
+                         'ldap-user-enabled-invert',
+                         'ldap-group-members-are-ids',
+                         'ldap-use-pool')
```
ajkavanagh, line 388: What makes a key 'special'?  I'm assuming it's something to do with a config value being 'None' or '' which means it doesn't override something in the ldap config key (i.e. there are two ways of setting things).  But it's not clear what 'special' means here.

zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -434,31 +439,42 @@ def _find_keystone_v3_user(self, username, domain):
     def test_100_keystone_ldap_users(self):
         """Validate basic functionality of keystone API with ldap."""
         application_name = 'keystone-ldap'
-        config = self._get_ldap_config()
+        intended_cfg = self._get_ldap_config()
+        filtered_cfg, special_cfg = self.config_current_separate_special_keys(
```
ajkavanagh, line 443: What is being 'filtered'.  Non 'special' keys?  There needs to be a better word to define what these things are.

zaza/openstack/charm_tests/test_utils.py
```diff
@@ -154,6 +154,36 @@ def setUpClass(cls, application_name=None, model_alias=None):
...
 
+    def config_current_separate_special_keys(
+            self, special_type_keys, config_keys=None,
+            application_name=None):
+        """Obtain current config and the special type config separately.
+
+        :param config_keys: iterable of strs to index into the current config.
+                            If None, return all keys from the config
+        :type config_keys:  Optional[Iterable[str]]
+        :param special_type_keys: list of keys to be separated out
```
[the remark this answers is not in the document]
ajkavanagh: I agree with @AurelienLourot here.  I suspect "special" means "keys that, when they have a default of None and are set to '' or None, means that they aren't used."  But it's not clear to me.  This is on the basis of reading the associated Gerrit merged (PR).

zaza/openstack/charm_tests/test_utils.py
```diff
@@ -154,6 +154,36 @@ def setUpClass(cls, application_name=None, model_alias=None):
...
+        :type special_type_keys: list
+        :param application_name: String application name for use when called
+                                 by a charm under test other than the object's
+                                 application.
+        :type application_name:  Optional[str]
+        :return: Dictionary of previous charm configs applied without the
+                 special type keys provided, and dictionary of the special keys
+                 found in the intended config dictionary.
```
ajkavanagh, line 173: I kind of get your meaning (except in the terminology of 'special'.   Please define special in terms of 'things that they do or are'.).

I think it's going to require an example to explain what is going on in the docstring.  i.e. Why should the keys be removed; what's the purpose behind it.

ajkavanagh (changes requested):
Please see inline.

unit_tests/charm_tests/test_utils.py
```diff
@@ -125,6 +125,65 @@ def test_config_change(self):
         self.assertFalse(self.wait_for_agent_status.called)
         self.assertFalse(self.wait_for_application_states.called)
 
+    def test_separate_non_string_config(self):
+        intended_cfg_keys = ['foo2', 'foo3', 'foo4', 'foo5']
+        previous_config_mock = {
```
AurelienLourot, line 130: Let's not use the wording "previous" here either, thanks :)

zaza/openstack/charm_tests/test_utils.py
```diff
@@ -154,6 +154,44 @@ def setUpClass(cls, application_name=None, model_alias=None):
...
+            self, non_string_type_keys, config_keys=None,
+            application_name=None):
+        """Obtain current config and the non-string type config separately.
+
+        If the charm config option is not string, it will not accept being
+        reverted back in "config_change()" method if the current value is None.
+        Therefore, obtain the current config and separate those out, so they
+        can be used for a separate invocation of "config_change()" with
+        reset_to_charm_default set to True.
+
```
AurelienLourot, line 167: We should make 100% clear here that things will get only separated out if they are currently None, otherwise they won't. Because now it just says something like "if you you don't separate them and they are None, you will get troubles" but it doesn't say that if they aren't None, they won't be separated. Maybe this could be explicitly stated in `:param non_string_type_keys:`

unit_tests/charm_tests/test_utils.py
```diff
@@ -125,6 +125,65 @@ def test_config_change(self):
...
+            'foo3': 'old_bar3',
+            'foo5': 'old_bar5',
+        }
+        expected_result_special = {
+            'foo2': None,
+            'foo4': None,
+        }
+        current, non_string = (
+            self.target.config_current_separate_non_string_type_keys(
+                non_string_type_keys, intended_cfg_keys, 'name')
```
AurelienLourot, line 149: nitpick: it'd be easier to read with 'application_name' instead of 'name'

AurelienLourot (changes requested):
This looks much better, thanks a lot for doing this! We're almost there as far as I'm concerned.

AurelienLourot (approved):
Thanks! Happy to land it on successful validation here:
- 20.10: https://review.opendev.org/c/openstack/charm-keystone-ldap/+/763439
- not-yet-reverted master: https://review.opendev.org/c/openstack/charm-keystone-ldap/+/763756

AurelienLourot:
CI has passed, @ajkavanagh 's comments have been addressed, landing, thanks!

PR #461, opened 2020-11-17
Log stderr when manila-ganesha verify() fails

coreycb:


ChrisMacNaughton (approved):
LGTM, I'm always a fan of adding more logging when things go wrong :+1:

PR #462, opened 2020-11-18
Support nova client micro versions

gnuoy:
Support micro versions when requesting a nova client.

### DIFF
--- a/zaza/openstack/utilities/openstack.py
+++ b/zaza/openstack/utilities/openstack.py
@@ -253,7 +253,7 @@ def get_designate_session_client(**kwargs):
                            **kwargs)
 
 
-def get_nova_session_client(session, version=None):
+def get_nova_session_client(session, version=2):
     """Return novaclient authenticated by keystone session.
 
     :param session: Keystone session object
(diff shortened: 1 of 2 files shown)


zaza/openstack/utilities/openstack.py
```diff
@@ -253,15 +253,18 @@ def get_designate_session_client(**kwargs):
...
 
     :param session: Keystone session object
     :type session: keystoneauth1.session.Session object
+    :param version: Version of client to request.
+    :type version: float
     :returns: Authenticated novaclient
     :rtype: novaclient.Client object
     """
-    return novaclient_client.Client(2, session=session)
+    version = version or 2
```
ajkavanagh, line 266: I'm kind of on the fence about whether this should be an `if version is None: ` rather than the `or` shortcut.  I tend to use the `or` shortcut for more complicated types (e.g. [] or {}) but it feels like None is reified and this is a numeric literal.  Hmm.

(Note. Not that this is not approving / just a comment, and I was thinking about it).

ajkavanagh (approved):

PR #469, opened 2020-11-26
Skip 'service' action tests on systems without systemd

mkalcok:
juju action 'service' is not supported on systems without systemd


ajkavanagh (approved):
LGTM

PR #470, opened 2020-12-08
Functional tests for nova-compute 'cloud' actions

mkalcok:
New actions were added to the nova-compute unit:

*  Actions that can `enable` or `disable` scheduling of new VMs on the nova-compute.
* Actions that can register/unregister nova-compute from nova-cloud controller.

This PR contains functional test that tests these four actions.

nova-compute code review: https://review.opendev.org/c/openstack/charm-nova-compute/+/763795

### DIFF
--- a/zaza/openstack/charm_tests/nova/tests.py
+++ b/zaza/openstack/charm_tests/nova/tests.py
@@ -49,6 +49,10 @@ class CirrosGuestCreateTest(test_utils.OpenStackBaseTest):
         self.launch_guest(
             'cirros', instance_key=glance_setup.CIRROS_IMAGE_NAME)
 
+    def tearDown(self):
+        """Cleanup of VM guests."""
+        self.resource_cleanup()
+
 
 class LTSGuestCreateTest(test_utils.OpenStackBaseTest):
     """Tests to launch a LTS image."""
@@ -58,6 +62,10 @@ class LTSGuestCreateTest(test_utils.OpenStackBaseTest):
         self.RESOURCE_PREFIX = 'zaza-nova'
         self.launch_guest(
             'ubuntu', instance_key=glance_setup.LTS_IMAGE_NAME)
+
+    def tearDown(self):
+        """Cleanup of VM guests."""
+        self.resource_cleanup()
 
 
 class LTSGuestCreateVolumeBackedTest(test_utils.OpenStackBaseTest):
@@ -69,6 +77,10 @@ class LTSGuestCreateVolumeBackedTest(test_utils.OpenStackBaseTest):
         self.launch_guest(
             'volume-backed-ubuntu', instance_key=glance_setup.LTS_IMAGE_NAME,
             use_boot_volume=True)
+
+    def tearDown(self):
+        """Cleanup of VM guests."""
+        self.resource_cleanup()
 
 
 class CloudActions(test_utils.OpenStackBaseTest):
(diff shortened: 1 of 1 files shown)


auria:
LGTM.

ajkavanagh (changes requested):
Please could you provide a link to the gerrit review for the change, so we can see the matching code in the nova-compute charm?  Also, please could you describe the change in the PR header for future travelers.  Thanks.  (Please also see inline comment).

mkalcok:
@ajkavanagh Sorry about that description. It should be ok now. However I don't see the inline comments you mention

ajkavanagh:
> @ajkavanagh Sorry about that description. It should be ok now. However I don't see the inline comments you mention

It looks like my comment went missing (probably due to an error I made!).  I'll re-do it.

zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -154,6 +154,31 @@ def test_930_check_virsh_default_network(self):
...
+            self.assertEqual(service.status, 'disabled')
+
+        # Run 'enable' action on units
+        zaza.model.run_action_on_units([unit.name for unit in nova_units],
+                                       'enable')
+
+        # Check action results via nova API
+        for service in self.nova_client.services.list(binary='nova-compute'):
+            self.assertEqual(service.status, 'enabled')
+
```
ajkavanagh, line 181: So if this is a test for a new feature, this will break on old versions of the charm, specifically the stable series.  As we don't (yet) cut stable branches of these tests, it's normal practice to add a new test class with this test (for a new feature), and add the test to the tests/tests.yaml for the master branch of the charm.  Then, the new version gets the new tests, and the stable version of the charm does.  This means that we will continue to be able to backport changes to the stable series.

mkalcok: Good point, thanks. I forgot that this was the procedure for new func tests.

mkalcok: done

ajkavanagh (commented):
Looks good; let's see the associated gerrit review pass prior to landing and it'll be good to merge.

mkalcok:
I added functional tests for actions `remove-from-cloud` and `register-to-cloud`. With their addition I also renamed class that encapsulate theses tests to `CloudActions` as it now covers all actions from nova-compute's `actions/cloud.py`

ajkavanagh (changes requested):
Holding until charm-freeze is over; don't want to disturb the tests during the testing week.

mkalcok:
@ajkavanagh Hi Alex, the associated gerrit review went through. https://review.opendev.org/c/openstack/charm-nova-compute/+/763795

zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -68,6 +78,139 @@ def test_launch_small_instance(self):
...
+        for _ in range(31):
+            sleep(sleep_timeout)
+            service_list = self.nova_client.services.list(
+                host=service_name, binary='nova-compute')
+            if len(service_list) == 0:
+                break
+            sleep_timeout = 10
+        else:
+            self.fail("nova-compute service was not unregistered from the "
+                      "nova-cloud-controller as expected.")
```
ajkavanagh, line 191: This code is sadly repeated below, which violates DRY.  There are several ways to make it better; I suspect tenacity would be a good approach.  However, sadly the actual review in the charm has been landed so this PR needs to be landed.  We'll have to fix it later.

ajkavanagh (approved):
Approving despite reservation on repeated code; the associated gerrit change the charm has already landed, which is generally the wrong order.

mkalcok:
Thanks for merge @ajkavanagh . You are right, the repeated code could've been handled better. I opened PR #534 that should solve it.

ajkavanagh:
> Thanks for the merge @ajkavanagh . You are right, the repeated code could've been handled better. I opened PR #534 that should solve it.

Yes, thank you; seen that and merged it.  Incidentally, it wasn't your 'problem'; sorry if it seemed like that.  I was more sad that the gerrit review had been landed before the associated tests had been fully reviewed and merged.  It kind of forces the issue! :)

PR #471, opened 2020-12-10
Add functional testcase test_virsh_audit_action for bug/1907409

mullumaus:
Bug #1907409 requested to add a new action 'virsh-audit' to list instances as virsh sees them on the compute node (virsh list --all) (sometimes this disagrees with what nova thinks). This PR added testcase test_virsh_audit_action in nova-compute to run action 'virsh-audit'

[ajkavanagh:edit] gerrit review link from comment below: https://review.opendev.org/c/openstack/charm-nova-compute/+/766412

### DIFF
--- a/unit_tests/utilities/test_zaza_utilities_parallel_series_upgrade.py
+++ b/unit_tests/utilities/test_zaza_utilities_parallel_series_upgrade.py
@@ -182,6 +182,8 @@ class TestParallelSeriesUpgrade(AioTestCase):
         self.model.async_wait_for_unit_idle = mock.AsyncMock()
         self.async_run_on_machine = mock.AsyncMock()
         self.model.async_run_on_machine = self.async_run_on_machine
+        self.model.async_block_until_units_on_machine_are_idle = \
+            mock.AsyncMock()
 
     @mock.patch.object(upgrade_utils.cl_utils, 'get_class')
     async def test_run_post_application_upgrade_functions(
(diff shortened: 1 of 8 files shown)


zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -154,6 +154,21 @@ def test_930_check_virsh_default_network(self):
...
+                         ' on  unit {}'.format(unit.entity_id))
+            action = zaza.model.run_action(
+                unit.entity_id,
+                'virsh-audit',
+                model_name=self.model_name,
+                action_params={})
+            if "failed" in action.data["status"]:
+                raise Exception(
+                    "The action failed: {}".format(action.data["message"]))
+
```
ajkavanagh, line 171: If this is a new action on the charm, then this will break on older versions, meaning we can't land stable updates.  If this is the case, then this should be a new test class and that new class added to the revised version of the charm

ajkavanagh (changes requested):
Please see review comment.  Also, please could you link the gerrit review that this test matches up against.  Also, because of issues with the new gerrit, the bug mentioned doesn't have the gerrit review either.  Thanks.

zaza/openstack/charm_tests/nova/tests.py
mullumaus: Hi Alex,  
Thanks for the comments. Yes, this is a new action,  I will add a new test class for it. And here is gerrit review link
https://review.opendev.org/c/openstack/charm-nova-compute/+/766412

PR #475, opened 2020-12-18
Fix BlueStoreCompressionCharmOperation against charm-ceph-iscsi

AurelienLourot:


zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -881,19 +881,31 @@ class BlueStoreCompressionCharmOperation(test_utils.BaseCharmTest):
...
         super(BlueStoreCompressionCharmOperation, self).setUp()
 
-        # determine if the tests should be run or not
-        logging.debug('os_release: {} >= {} = {}'
-                      .format(self.current_release,
-                              self.bionic_rocky,
-                              self.current_release >= self.bionic_rocky))
-        self.mimic_or_newer = self.current_release >= self.bionic_rocky
+        # Determine if the tests should be run or not. If we couldn't determine
+        # the current OpenStack release, assume yes.
```
thedac, line 900: This seems like an unsafe assumption.

thedac (changes requested):
I do not think this is the right approach. We cannot assume that because keystone is not there we are >= mimic. It does not follow. We need to find a way to inspect the actual ceph release. Even if that takes a new utility.

ChrisMacNaughton:
The associated gerrit review has been abandoned (no need to add groovy support) and the iscsi charm has had a lot of work done recently to improve it's test stance elsewhere. Closing this change as we can re-open it if we decide that the change is needed later

PR #476, opened 2021-01-04
Update openstack upgrade tests for focal

ajkavanagh:
This patch modifies the existing openstack upgrade tests so that they
work with focal (by explicitly supporting mysql-innodb-cluster), and are
also interruptable and resumable (at a charm level).  It also makes them
work with the udpated 'get_upgrade_groups()' that ultimately gets a List
of Tuples rather than a dictionary.

### DIFF
--- a/zaza/openstack/utilities/openstack_upgrade.py
+++ b/zaza/openstack/utilities/openstack_upgrade.py
@@ -95,8 +95,8 @@ async def async_action_unit_upgrade(units, model_name=None):
 action_unit_upgrade = sync_wrapper(async_action_unit_upgrade)
 
 
-def action_upgrade_group(applications, model_name=None):
-    """Upgrade units using action managed upgrades.
+def action_upgrade_apps(applications, model_name=None):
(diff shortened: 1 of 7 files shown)


zaza/openstack/utilities/openstack_upgrade.py
```diff
@@ -95,8 +95,8 @@ async def async_action_unit_upgrade(units, model_name=None):
 action_unit_upgrade = sync_wrapper(async_action_unit_upgrade)
 
 
-def action_upgrade_group(applications, model_name=None):
-    """Upgrade units using action managed upgrades.
+def action_upgrade_apps(applications, model_name=None):
+    """Upgrade units in the applciations using action managed upgrades.
```
ChrisMacNaughton, line 99: `in the applications`

ChrisMacNaughton (commented):
Just one typo that I noticed, otherwise looks good

ChrisMacNaughton (approved):

PR #478, opened 2021-01-06
Use ceph-mon to check ceph version not keystone

gnuoy:
The test class BlueStoreCompressionCharmOperation gates tests on
whether the ceph release is mimic or newer but it uses the
keystone application to calculate the currently deployed version.
This PR switches the test class to ceck the version of ceph-mon
instead which makes more sense and the keystone application may
not always present in a ceph deployment.


AurelienLourot (changes requested):
Won't work on victoria as it will produce a pair "groovy_ussuri" which doesn't exist and raise

AurelienLourot (approved):
groovy-victoria is actually not part of the gate yet and I'll fix that case by rebasing #475 on top of this, thanks!

PR #480, opened 2021-01-10
Add multi-backend testing for Keystone SAML Mellon

ionutbalutoiu:
The new updated tests rely on new testing bundles deployed with
two local IdPs via the Juju charm https://jaas.ai/u/ionutbalutoiu/test-saml-idp.

### DIFF
--- a/zaza/openstack/charm_tests/saml_mellon/setup.py
+++ b/zaza/openstack/charm_tests/saml_mellon/setup.py
@@ -25,6 +25,7 @@ from zaza.openstack.utilities import (
     cert as cert_utils,
     cli as cli_utils,
     openstack as openstack_utils,
+    generic as generic_utils,
 )
 
 
@@ -34,8 +35,8 @@ FEDERATED_DOMAIN = "federated_domain"
 FEDERATED_GROUP = "federated_users"
 MEMBER = "Member"
 IDP = "samltest"
+LOCAL_IDP_REMOTE_ID = "http://{}/simplesaml/saml2/idp/metadata.php"
 REMOTE_ID = "https://samltest.id/saml/idp"
-MAP_NAME = "{}_mapping".format(IDP)
 PROTOCOL_NAME = "mapped"
 MAP_TEMPLATE = '''
     [{{
@@ -45,7 +46,7 @@ MAP_TEMPLATE = '''
                         "name": "{{0}}"
                     }},
                     "group": {{
-                        "name": "federated_users",
+                        "name": "{group_id}",
                         "domain": {{
                             "id": "{domain_id}"
                         }}
@@ -55,7 +56,7 @@ MAP_TEMPLATE = '''
                         "name": "{{0}}_project",
                         "roles": [
                                      {{
-                                         "name": "Member"
+                                         "name": "{role_name}"
                                      }}
                                  ]
                     }}
@@ -81,7 +82,10 @@ SP_SIGNING_KEY_INFO_XML_TEMPLATE = '''
 '''
 
 
-def keystone_federation_setup():
+def keystone_federation_setup(federated_domain=FEDERATED_DOMAIN,
+                              federated_group=FEDERATED_GROUP,
+                              idp_name=IDP,
+                              idp_remote_id=REMOTE_ID):
     """Configure Keystone Federation."""
     cli_utils.setup_logging()
     keystone_session = openstack_utils.get_overcloud_keystone_session()
@@ -89,19 +93,19 @@ def keystone_federation_setup():
         keystone_session)
 
     try:
-        domain = keystone_client.domains.find(name=FEDERATED_DOMAIN)
+        domain = keystone_client.domains.find(name=federated_domain)
     except keystoneauth1.exceptions.http.NotFound:
         domain = keystone_client.domains.create(
-            FEDERATED_DOMAIN,
+            federated_domain,
             description="Federated Domain",
             enabled=True)
 
     try:
         group = keystone_client.groups.find(
-            name=FEDERATED_GROUP, domain=domain)
+            name=federated_group, domain=domain)
     except keystoneauth1.exceptions.http.NotFound:
         group = keystone_client.groups.create(
-            FEDERATED_GROUP,
+            federated_group,
             domain=domain,
             enabled=True)
 
@@ -109,30 +113,33 @@ def keystone_federation_setup():
     keystone_client.roles.grant(role, group=group, domain=domain)
 
     try:
-        idp = keystone_client.federation.identity_providers.find(
-            name=IDP, domain_id=domain.id)
+        idp = keystone_client.federation.identity_providers.get(idp_name)
     except keystoneauth1.exceptions.http.NotFound:
         idp = keystone_client.federation.identity_providers.create(
-            IDP,
-            remote_ids=[REMOTE_ID],
+            idp_name,
+            remote_ids=[idp_remote_id],
             domain_id=domain.id,
             enabled=True)
 
-    JSON_RULES = json.loads(MAP_TEMPLATE.format(domain_id=domain.id))
+    JSON_RULES = json.loads(MAP_TEMPLATE.format(
+        domain_id=domain.id, group_id=group.id, role_name=MEMBER))
 
+    map_name = "{}_mapping".format(idp_name)
     try:
-        keystone_client.federation.mappings.find(name=MAP_NAME)
+        keystone_client.federation.mappings.get(map_name)
     except keystoneauth1.exceptions.http.NotFound:
         keystone_client.federation.mappings.create(
-            MAP_NAME, rules=JSON_RULES)
+            map_name, rules=JSON_RULES)
 
     try:
-        keystone_client.federation.protocols.get(IDP, PROTOCOL_NAME)
+        keystone_client.federation.protocols.get(idp_name, PROTOCOL_NAME)
     except keystoneauth1.exceptions.http.NotFound:
         keystone_client.federation.protocols.create(
-            PROTOCOL_NAME, mapping=MAP_NAME, identity_provider=idp)
+            PROTOCOL_NAME, mapping=map_name, identity_provider=idp)
 
 
+# This setup method is deprecated. It will be removed once we fully drop the
+# `samltest.id` dependency.
 def attach_saml_resources(application="keystone-saml-mellon"):
     """Attach resource to the Keystone SAML Mellon charm."""
     test_idp_metadata_xml = "samltest.xml"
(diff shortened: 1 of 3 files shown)


zaza/openstack/charm_tests/saml_mellon/tests.py
```diff
@@ -20,28 +20,28 @@
...
 
 
 class FailedToReachIDP(Exception):
     """Custom Exception for failing to reach the IDP."""
 
-    pass
 
-
-class CharmKeystoneSAMLMellonTest(BaseKeystoneTest):
+class BaseCharmKeystoneSAMLMellonTest(BaseKeystoneTest):
```
ajkavanagh, line 30: Changing the name from `CharmKeystoneSAMLMellonTest` to `CharmKeystoneSAMLMellonIDP1Test` will, I think, break stable versions of the charm tests.  We tend to have to leave the old names alone and add new ones only so as to not break existing stable charms.

ajkavanagh (changes requested):
Looks good apart from the test name change for stable charms / existing tests (e.g. from a tests.yaml).  As the Test names are essentially the API to the testing library they tend to be "forever".

zaza/openstack/charm_tests/saml_mellon/tests.py
ionutbalutoiu: @ajkavanagh

Makes sense. I'll update the PR, and make sure to keep the backwards compatibility.

ionutbalutoiu:
> Looks good apart from the test name change for stable charms / existing tests (e.g. from a tests.yaml). As the Test names are essentially the API to the testing library they tend to be "forever".

I updated the PR and took care of the backwards compatibility. I also validated these changes with a `func-smoke` against:
* Latest master with my bundles changes from https://review.opendev.org/c/openstack/charm-keystone-saml-mellon/+/770056
* The `stable/20.10` charm release.

Everything went fine on my local Zaza environment.

There's some duplicate code in the `zaza/openstack/charm_tests/saml_mellon/tests.py` between the old `CharmKeystoneSAMLMellonTest` class (which uses upstream `samltest.id`), and the new `BaseCharmKeystoneSAMLMellonTest` class (which uses the local `test-saml-idp`).

I left it like this on purpose, since we can remove the old testing class, once we fully drop the `samltest.id` dependency (even on stable charm releases).

AurelienLourot (changes requested):
I haven't reviewed yet, just posting this comment first, but I'll circle back to it soon, thanks!

zaza/openstack/charm_tests/saml_mellon/tests.py
```diff
@@ -156,3 +156,177 @@ def _do_redirect_check(url, region, idp_expect, horizon_expect):
         # We may need to try/except to allow horizon to build its pages
         _do_redirect_check(url, region, idp_expect, horizon_expect)
         logging.info("SUCCESS")
+
```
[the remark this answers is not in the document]
ionutbalutoiu: @AurelienLourot Sounds good. I'll wait for your review, and address all the comments at once.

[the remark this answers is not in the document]
ionutbalutoiu: Added a comment about the deprecation of the old testing class using `samltest.id` dependency.

zaza/openstack/charm_tests/saml_mellon/tests.py
```diff
@@ -156,3 +158,177 @@ def _do_redirect_check(url, region, idp_expect, horizon_expect):
...
+    @classmethod
+    def setUpClass(cls,
+                   application_name="keystone-saml-mellon",
+                   test_saml_idp_app_name="test-saml-idp",
+                   horizon_idp_option_name="myidp_mapped",
+                   horizon_idp_display_name="myidp via mapped"):
+        """Run class setup for running Keystone SAML Mellon charm tests."""
+        super(BaseCharmKeystoneSAMLMellonTest, cls).setUpClass()
+        cls.application_name = application_name
```
AurelienLourot, line 174: This note was useful, can we keep it?

```
        # Note: The BaseKeystoneTest class sets the application_name to
        # "keystone" which breaks keystone-saml-mellon actions. Explicitly set
        # application name here.
```

AurelienLourot (changes requested):
LGTM in general, thanks!

zaza/openstack/charm_tests/saml_mellon/setup.py
```diff
@@ -161,3 +165,88 @@ def attach_saml_resources(application="keystone-saml-mellon"):
...
+        test_saml_idp_app_name).pop()
+    action_result = zaza.model.run_action(
+        test_saml_idp_unit.name, 'get-idp-metadata')
+    idp_metadata = action_result.data['results']['output']
+
+    with tempfile.NamedTemporaryFile(mode='w', suffix='.xml') as fp:
+        fp.write(idp_metadata)
+        fp.flush()
+        zaza.model.attach_resource(
+            keystone_saml_mellon_app_name, "idp-metadata", fp.name)
```
ajkavanagh, line 183: There's a pattern developing here (this is repeated 3 times).  Would breaking out a helper that does this and attaches the resource (or even add a new helper), help with the clarity and intention of the code.  e.g.

```python
zaza.utils.attach_resource(app, "idp-metadata', idp_metadata, ".xml")
```

zaza/openstack/charm_tests/saml_mellon/tests.py
```diff
@@ -156,3 +158,177 @@ def _do_redirect_check(url, region, idp_expect, horizon_expect):
...
+        cls.current_release = openstack_utils.get_os_release()
+        cls.FOCAL_USSURI = openstack_utils.get_os_release("focal_ussuri")
+
+    def test_run_get_sp_metadata_action(self):
+        """Validate the get-sp-metadata action."""
+        unit = zaza.model.get_units(self.application_name)[0]
+        if self.vip:
+            ip = self.vip
+        else:
+            ip = unit.public_address
```
ajkavanagh, line 188: I tend to prefer (but don't change it if you prefer the current format):

```python
ip = self.vip if self.vip else unit.public_address
```

zaza/openstack/charm_tests/saml_mellon/tests.py
```diff
@@ -156,3 +158,177 @@ def _do_redirect_check(url, region, idp_expect, horizon_expect):
...
+
+        action = zaza.model.run_action(unit.entity_id, self.action)
+        if "failed" in action.data["status"]:
+            raise Exception(
+                "The action failed: {}".format(action.data["message"]))
```
ajkavanagh, line 193: This is test code.  Why not:

```python
self.assertNotIn("failed", action.data["status"], msg="The action failed ....")
```

zaza/openstack/charm_tests/saml_mellon/tests.py
```diff
@@ -156,3 +158,177 @@ def _do_redirect_check(url, region, idp_expect, horizon_expect):
...
+
+        output = action.data["results"]["output"]
+        root = etree.fromstring(output)
+        for item in root.items():
+            if "entityID" in item[0]:
+                assert ip in item[1]
```
ajkavanagh, line 199:
```python
self.assertIn(ip, item[1])
```

zaza/openstack/charm_tests/saml_mellon/tests.py
```diff
@@ -156,3 +158,177 @@ def _do_redirect_check(url, region, idp_expect, horizon_expect):
...
+
+        for appt in root.getchildren():
+            for elem in appt.getchildren():
+                for item in elem.items():
+                    if "Location" in item[0]:
+                        assert ip in item[1]
```
ajkavanagh, line 205: As above.

zaza/openstack/charm_tests/saml_mellon/tests.py
```diff
@@ -156,3 +158,177 @@ def _do_redirect_check(url, region, idp_expect, horizon_expect):
...
+        if horizon_vip:
+            horizon_ip = horizon_vip
+        else:
+            unit = zaza.model.get_units("openstack-dashboard")[0]
+            horizon_ip = unit.public_address
+
+        if self.tls_rid:
+            proto = "https"
+        else:
+            proto = "http"
```
ajkavanagh, line 229: One line if-else possible if preferred.

zaza/openstack/charm_tests/saml_mellon/tests.py
```diff
@@ -156,3 +158,177 @@ def _do_redirect_check(url, region, idp_expect, horizon_expect):
...
+                msg = "FAILURE code={} text={}".format(response, response.text)
+                # Raise a custom exception.
+                raise FailedToReachIDP(msg)
+
+            idp_url = ("http://{0}/simplesaml/"
+                       "module.php/core/loginuserpass.php").format(idp_address)
+
+            # Validate that we were redirected to the proper IdP
+            assert response.url.startswith(idp_url)
+            assert idp_url in response.text
```
ajkavanagh, line 296: This is a big function to be embedded in another.  Please break it out and add `@staticmethod` to it, plus docstring with purpose, parameters and types.

ajkavanagh (changes requested):
Please see the single comments; apologies for not starting a review.   I'd already proceeding with the single comments and it was too late by then!

zaza/openstack/charm_tests/saml_mellon/setup.py
```diff
@@ -81,56 +81,60 @@
...
     try:
-        keystone_client.federation.protocols.get(IDP, PROTOCOL_NAME)
+        keystone_client.federation.protocols.get(idp_name, PROTOCOL_NAME)
     except keystoneauth1.exceptions.http.NotFound:
         keystone_client.federation.protocols.create(
-            PROTOCOL_NAME, mapping=MAP_NAME, identity_provider=idp)
+            PROTOCOL_NAME, mapping=map_name, identity_provider=idp)
 
 
 def attach_saml_resources(application="keystone-saml-mellon"):
```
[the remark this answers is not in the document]
ionutbalutoiu: Yes. I'll add the commend.

zaza/openstack/charm_tests/saml_mellon/setup.py
```diff
@@ -161,3 +165,88 @@ def attach_saml_resources(application="keystone-saml-mellon"):
         fp.flush()
         zaza.model.attach_resource(
             application, sp_signing_keyinfo_name, fp.name)
+
+
+def _attach_saml_resources_local_idp(keystone_saml_mellon_app_name=None,
+                                     test_saml_idp_app_name=None):
+    """Attach resources to the Keystone SAML Mellon and the local IdP."""
+    test_saml_idp_unit = zaza.model.get_units(
+        test_saml_idp_app_name).pop()
```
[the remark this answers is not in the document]
ionutbalutoiu: In fact, I only need to run the `get-idp-metadata` action, and I noticed that we have `run_action_on_leader` and I don't even need to fetch an unit. I'll use that.

zaza/openstack/charm_tests/saml_mellon/setup.py
ionutbalutoiu: Good suggestion. Will implement it.

zaza/openstack/charm_tests/saml_mellon/setup.py
```diff
@@ -161,3 +165,88 @@ def attach_saml_resources(application="keystone-saml-mellon"):
...
+            keystone_saml_mellon_app_name, 'sp-private-key', fp.name)
+
+    with tempfile.NamedTemporaryFile(mode='w', suffix='.xml') as fp:
+        fp.write(SP_SIGNING_KEY_INFO_XML_TEMPLATE.format(cert))
+        fp.flush()
+        zaza.model.attach_resource(
+            keystone_saml_mellon_app_name, 'sp-signing-keyinfo', fp.name)
+
+    keystone_saml_mellon_unit = zaza.model.get_units(
+        keystone_saml_mellon_app_name).pop()
```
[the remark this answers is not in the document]
ionutbalutoiu: Same solution as above. I can use `runa_action_on_leader`.

zaza/openstack/charm_tests/saml_mellon/tests.py
ionutbalutoiu: Will add it back.

zaza/openstack/charm_tests/saml_mellon/tests.py
ionutbalutoiu: I like this as well. I'll change it.

zaza/openstack/charm_tests/saml_mellon/tests.py
ionutbalutoiu: Will change it.

zaza/openstack/charm_tests/saml_mellon/tests.py
ionutbalutoiu: This function was carried on from the previous implementation.

I agree with you. I'll clean it up.

ionutbalutoiu:
@ajkavanagh @AurelienLourot

Thank-you for the reviews. I addressed all the commends. Please check the updated PR when you have a chance.

ajkavanagh (approved):
LGTM - thanks for making the  changes, particularly the code-gardening around the embedded function; much clearer now.

AurelienLourot (approved):
Perfect, thanks! Happy to land this when the corresponding gerrit review turns green. ( https://review.opendev.org/c/openstack/charm-keystone-saml-mellon/+/770056 )

AurelienLourot:
Note: the `.decode()` call has been moved, this is what has changed in the last push-force. Still LGTM. @ionutbalutoiu can you please not push-force? Because we can't easily spot here in the browser what changed and this requires special steps to figure it out. Thanks!

ionutbalutoiu:
> Note: the `.decode()` call has been moved, this is what has changed in the last push-force. Still LGTM. @ionutbalutoiu can you please not push-force? Because we can't easily spot here in the browser what changed and this requires special steps to figure it out. Thanks!

Oh, sorry about `force-push`. I tend to have a single clean commit. But I realize that it can get confusing when the changes are reviewed already.

The `decode()` had to be removed from there, because the helper function is meant to get a `str` (not an encoded `str`).

After I got your approvals, I quickly ran a local Zaza run and noticed that the function was failing when it got `str`, because it tried to `decode()` it. The `decode()` needs to be called before using the helper (if needed).

That's the full story of the most recent change. Our local Zaza run went through without any problems after fixing this.

AurelienLourot (approved):

PR #483, opened 2021-01-12
Support configuring networknig charms on MAAS

fnordahl:
The function does three separate things today, and two of its
tasks are useful for other provider types such as MAAS.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,22 @@
 [tox]
 envlist = pep8, py3
 skipsdist = True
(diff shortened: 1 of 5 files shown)


fnordahl:
A successful test run confirming the code works just as before with OpenStack provider in OSCI can be viewed here: https://review.opendev.org/c/openstack/charm-neutron-openvswitch/+/770653

I have verified this on MAAS locally by hand.

ajkavanagh (approved):
LGTM

PR #485, opened 2021-01-19
Fix policyd call to _login

gnuoy:
8681b023 changed the signature of _login but did not update the
policyd test. This was not immediately picked up because the
policyd test is skipped before groovy due to Bug #1880959

Closes-Bug: #1911923


AurelienLourot:
CI validation: https://review.opendev.org/c/openstack/charm-openstack-dashboard/+/761582

AurelienLourot (approved):

PR #486, opened 2021-01-22
Update Ceph RBD Mirror tests

ionutbalutoiu:
The updated tests add the possibility of testing deployments
with `image` RBD mirroring mode implemented as part of the
[Cinder Ceph Replication](https://opendev.org/openstack/charm-specs/src/commit/8cbfdbc9081d95d11e3db047fae4d4a98b0ef2d5/specs/wallaby/backlog/cinder-ceph-replication.rst) charm spec.

### DIFF
--- a/zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
+++ b/zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
@@ -16,6 +16,9 @@
 import json
 import logging
 import re
+import time
+
+import cinderclient.exceptions as cinder_exceptions
 
 import zaza.openstack.charm_tests.test_utils as test_utils
 
@@ -39,16 +42,20 @@ class CephRBDMirrorBase(test_utils.OpenStackBaseTest):
         cls.site_a_model = cls.site_b_model = zaza.model.get_juju_model()
         cls.site_b_app_suffix = '-b'
 
-    def run_status_action(self, application_name=None, model_name=None):
+    def run_status_action(self, application_name=None, model_name=None,
+                          pools=[]):
         """Run status action, decode and return response."""
+        action_params = {
+            'verbose': True,
+            'format': 'json',
+        }
+        if len(pools) > 0:
+            action_params['pools'] = ','.join(pools)
         result = zaza.model.run_action_on_leader(
             application_name or self.application_name,
             'status',
             model_name=model_name,
-            action_params={
-                'verbose': True,
-                'format': 'json',
-            })
+            action_params=action_params)
         return json.loads(result.results['output'])
 
     def get_pools(self):
@@ -71,7 +78,8 @@ class CephRBDMirrorBase(test_utils.OpenStackBaseTest):
     def wait_for_mirror_state(self, state, application_name=None,
                               model_name=None,
                               check_entries_behind_master=False,
-                              require_images_in=[]):
+                              require_images_in=[],
+                              pools=[]):
         """Wait until all images reach requested state.
 
         This function runs the ``status`` action and examines the data it
@@ -90,6 +98,9 @@ class CephRBDMirrorBase(test_utils.OpenStackBaseTest):
         :type check_entries_behind_master: bool
         :param require_images_in: List of pools to require images in
         :type require_images_in: list of str
+        :param pools: List of pools to run status on. If this is empty, the
+                      status action will run on all the pools.
+        :type pools: list of str
         :returns: True on success, never returns on failure
         """
         rep = re.compile(r'.*entries_behind_master=(\d+)')
@@ -97,7 +108,8 @@ class CephRBDMirrorBase(test_utils.OpenStackBaseTest):
             try:
                 # encapsulate in try except to work around LP: #1820976
                 pool_status = self.run_status_action(
-                    application_name=application_name, model_name=model_name)
+                    application_name=application_name, model_name=model_name,
+                    pools=pools)
             except KeyError:
                 continue
             for pool, status in pool_status.items():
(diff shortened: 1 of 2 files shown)


AurelienLourot:
Note for future travelers: here is a charm change making use of this PR: https://review.opendev.org/c/openstack/charm-ceph-rbd-mirror/+/772001

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -124,6 +136,119 @@ def wait_for_mirror_state(self, state, application_name=None,
...
+            glance = openstack.get_glance_session_client(session)
+            images = openstack.get_images_by_name(glance, CIRROS_IMAGE_NAME)
+            if images:
+                return images[0]
+            logging.info("Failed to find {} image, falling back to {}".format(
+                CIRROS_IMAGE_NAME,
+                LTS_IMAGE_NAME))
+            return openstack.get_images_by_name(glance, LTS_IMAGE_NAME)[0]
+
+        def create_volume_type(cinder):
```
AurelienLourot, line 173: I think it should be `create_repl_volume_type_if_doesnt_exist` or so.

Also I'd rather have such generic reusable functions as private functions one level above (not as functions in functions) so that we could easily use them from somewhere else some day without having again to move the code (reviews where code gets moved are really time consuming for reviewers and require a lot of focus to try to spot tiny differences between a big red wall of code and a big green wall of code :D )

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -124,6 +136,119 @@ def wait_for_mirror_state(self, state, application_name=None,
...
+            try:
+                vol_type = cinder.volume_types.find(name='repl')
+            except cinder_exceptions.NotFound:
+                vol_type = cinder.volume_types.create('repl')
+                vol_type.set_keys(metadata={
+                    'volume_backend_name': 'cinder-ceph',
```
AurelienLourot, line 179: Shouldn't `'cinder-ceph'` actually come from an argument instead of being hard-coded, like for `get_cinder_rbd_mirroring_mode()`?

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -124,6 +136,119 @@ def wait_for_mirror_state(self, state, application_name=None,
...
+                                    timeout=300):
+        """Failover Cinder volume host."""
+        host = 'cinder@{}'.format(backend_name)
+        logging.info(
+            'Failover Cinder host %s to backend_id %s',
+            host, target_backend_id)
+        cinder_client.services.failover_host(
+            host=host,
+            backend_id=target_backend_id)
+        start = time.time()
```
AurelienLourot, line 234: I think we should use tenacity.Retry() here, except if there is a reason not to (like on line 184).

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -258,85 +347,170 @@ def create_volume_from_image(cinder, image, retry=20):
...
+        self.assertEqual(cinder.volumes.get(volume.id).status, 'available')
+        self.failover_cinder_volume_host(
+            cinder_client=cinder,
+            target_backend_id='default',
+            target_status='enabled',
+            target_replication_status='enabled')
+        self.assertEqual(cinder.volumes.get(volume.id).status, 'available')
+
     def test_fail_over_fall_back(self):
         """Validate controlled fail over and fall back."""
```
AurelienLourot, line 372: This is hard to understand now why this function is doing way more than just calling `cinder_fail_over_fall_back()`, although both functions have pretty much the same description. Is there a way to make things crystal clear in the description of the functions? Thanks!

EDIT: Having now read it fully, it feels like we demote, promote, resync stuff, and then finally we do a failover and then fall back. Can it be that this is a giant test validating many things? Would it be wise to split it? I'm sorry I'm having a really hard time to follow that scenario.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -258,85 +347,170 @@ def create_volume_from_image(cinder, image, retry=20):
...
         site_a_pools, site_b_pools = self.get_pools()
+        site_a_action_params = {}
+        site_b_action_params = {}
+        if self.get_cinder_rbd_mirroring_mode() == 'image':
+            site_a_pools.remove('cinder-ceph')
```
AurelienLourot, line 377: I'm getting lost here. It feels like this test requires expert knowledge to be understood. What I read is "if the mirroring mode is image, remove a pool called cinder-ceph". I'm lost and I don't understand what's going on here. I think we should add inline comments to make it possible for someone who isn't an expert here to vaguely understand what is going on :) Why is it necessary to remove pools in order to validate a failover? Are these pools a leftover from previous tests? Should they be cleaned up in these previous tests then?

AurelienLourot (changes requested):
Thanks for working on this! I have reviewed everything so far but `CephRBDMirrorDisasterFailoverTest` (keeping it for the next iteration). Please do not push-force from now on on this particularly big review as this would then be really hard for me to spot what has changed, thanks!

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
ionutbalutoiu: Sounds good. I'll make the change.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
ionutbalutoiu: Yes, you're right. I'll change it.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
ionutbalutoiu: I'll look to use the `tenacity` helper.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
ionutbalutoiu: I believe you're right. The function was already hard to understand as it has a lot of business logic. Adding more on top on it is not a good idea.

I'll refactor the function a bit.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
ionutbalutoiu: No.

I'm sorry for the opacity. The idea goes like this:

* When you have an environment with Ceph image-mirroring enabled, you want to do the fail-over / fail-back via Cinder API, **only** for the Cinder pool (because that's the reason image-mirroring makes sense in the first place).
* However, in the same environment, there might be other pools replicated (like Glance, for example). These pools need to be fail-overed / fail-back as usual, via the charm actions.

  This is exactly the main reason I added another optional parameter to the `ceph-rbd-mirror` actions, with the pools list. Because we want to exclude the `cinder-ceph` pool when running the Juju actions, since Cinder is enabled to do the `demote`, `promote` stuff.

That's the extra functionality here. The test is doing everything as before, but excluding the `cinder-ceph` pool, *only if* Cinder has image mirroring enabled (and we use the Cinder API in that case).

I'll make sure I add comments / docstring to clearly reflect this.

ionutbalutoiu:
@AurelienLourot 

Please see the latest PR update. The commit https://github.com/openstack-charmers/zaza-openstack-tests/pull/486/commits/2fefca5a406647a1efcd91e5f67441766f6976b7 has all the changes since you reviewed the PR.

I refactored the code to have it cleaner, and added tons of comments / docstrings to explain the tests workflow.

Also, when we are ready to merge, I would rebase against current master, and squash the changes into a single commit.

AurelienLourot (approved):
Awesome, thanks a lot!

Issue #487, opened 2021-01-23
Tests assume the certificate is called keystone_juju_ca_cert.crt


gnuoy:
After a recent change to charm helpers the certificate that comes from vault is now called `keystone_juju_ca_cert.crt` not `keystone_juju_ca_cert.crt`. This breaks some setup and tests which assume the old name.

PR #488, opened 2021-01-23
Handle change of CA cert.

gnuoy:
Update code to stop assuming that a charms CA file is called keystone_juju_ca_cert.crt. 

Closes issue #487

### DIFF
--- a/zaza/openstack/utilities/openstack.py
+++ b/zaza/openstack/utilities/openstack.py
@@ -2053,7 +2053,7 @@ def get_overcloud_auth(address=None, model_name=None):
             'API_VERSION': 3,
         }
     local_ca_cert = get_remote_ca_cert_file('keystone', model_name=model_name)
-    if ca_cert:
+    if local_ca_cert:
         auth_settings['OS_CACERT'] = local_ca_cert
 
     return auth_settings
--- a/zaza/openstack/utilities/openstack.py
+++ b/zaza/openstack/utilities/openstack.py
@@ -69,6 +69,7 @@ from neutronclient.common import exceptions as neutronexceptions
 from octaviaclient.api.v2 import octavia as octaviaclient
 from swiftclient import client as swiftclient
 
+from juju.errors import JujuError
 
 import zaza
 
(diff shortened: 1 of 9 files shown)


AurelienLourot:
CI validation: https://review.opendev.org/c/openstack/charm-masakari-monitors/+/761569

zaza/openstack/utilities/openstack.py
```diff
@@ -181,18 +183,70 @@
...
+                    output = await unit.run('cat {}'.format(ca_file))
+                    contents = output.data.get('results').get('Stdout', '')
+                    if ca_cert not in contents:
+                        break
+                # libjuju throws a generic error for connection failure. So we
+                # cannot differentiate between a connectivity issue and a
+                # target file not existing error. For now just assume the
+                # latter.
+                except JujuError:
+                    continue
```
AurelienLourot, line 223: This should be `break`, as `continue` stands for "we have found the cert", right?

zaza/openstack/utilities/openstack.py
```diff
@@ -1950,29 +2004,83 @@ def get_overcloud_auth(address=None, model_name=None):
...
+            except JujuError:
+                continue
+            # ensure that the path to put the local cacert in actually exists.
+            # The assumption that 'tests/' exists for, say, mojo is false.
+            # Needed due to:
+            # commit: 537473ad3addeaa3d1e4e2d0fd556aeaa4018eb2
+            _dir = os.path.dirname(_local_cert_file)
+            if not os.path.exists(_dir):
+                os.makedirs(_dir)
+            shutil.move(_tmp_ca.name, _local_cert_file)
```
AurelienLourot, line 2077: I think the idea is that "move" is atomic, so your intention is good, but keep in mind that it's not atomic if done across two different filesystems, and I think here you're moving from a tmpfs to `/usr` so it won't be atomic

AurelienLourot (changes requested):
I haven't reviewed the unit tests yet. Looks good to me in general, thanks!

zaza/openstack/utilities/openstack.py
```diff
@@ -181,18 +183,70 @@
...
+KEYSTONE_LOCAL_CACERT = ("{}/{}".format(LOCAL_CERT_DIR, KEYSTONE_CACERT))
+
+
+async def async_block_until_ca_exists(application_name, ca_cert,
+                                      model_name=None, timeout=2700):
+    """Block until a CA cert is on all units of application_name.
+
+    :param application_name: Name of application to check
+    :type application_name: str
+    :param ca_cert: The certificate to look for.
```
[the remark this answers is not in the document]
gnuoy: Updated

zaza/openstack/utilities/openstack.py
gnuoy: Good spot, thanks!

zaza/openstack/utilities/openstack.py
gnuoy: In this case its the fact that shutil.move works across filesystem boundaries that made me opt for it rather than os.rename. Whether the operation is atomic or not it not too much of an issue here.

AurelienLourot (approved):

thedac (approved):
Looks good.
I confrim this resolved my problems with the keystone-saml-mellon charm [0]

[0] https://review.opendev.org/c/openstack/charm-keystone-saml-mellon/+/771024

PR #490, opened 2021-01-29
Add a check before scp'ing ca cert

gnuoy:
This check before scp is not strictly needed but without it
scp errors are outputed and it wrongly appears to be an
error.


AurelienLourot (approved):

AurelienLourot (changes requested):
Travis failure looks legit: I think something needs to be mocked

PR #493, opened 2021-02-01
Use per-model tmp-dir to store local copy of CA cert

fnordahl:
The current approach of storing the deployment CA certificate in
the 'test/' relative path does not allow for executing tests for
multiple targets from the same environment.

We have previously moved (7a90110) the local copy of the SSH
private key for similar reasons.

Remove the global constants as we cannot build them without making
function calls, and we'd rather avoid doing that at module import
time. Code using the location of the local CA certificate has
already been changed to use helper functions.


fnordahl:
Excerpts from a set of successful runs as in Zaza can talk to the deployd cloud APIs and have put the files in the expected location:
https://gist.github.com/fnordahl/f7c0e57808b736d8542a8b4903eea27d

ChrisMacNaughton (approved):

PR #494, opened 2021-02-01
Replace get_relation_from_unit for designate test

Gleland:
In the current state, deprecation warnings will show for the `get_relation_from_unit` method when running tox:

```
[WARNING] get_relation_from_unit from zaza.openstack.utilities.juju is deprecated. Please use the equivalent from zaza.utilities.juju
```

This will remove the error. There are other files using this method as well, but to keep the change atomic, they will be handled separately. I changed the library alias to be `juju_utils` to be consistent with [other cases that already exist in the repo](https://github.com/openstack-charmers/zaza-openstack-tests/blob/91924c5caf40ac59039ea52fb13ab83325ac469f/zaza/openstack/utilities/openstack.py#L2934).

### DIFF
--- a/zaza/openstack/charm_tests/designate/tests.py
+++ b/zaza/openstack/charm_tests/designate/tests.py
@@ -23,7 +23,7 @@ import designateclient.v1.records as records
 import designateclient.v1.servers as servers
 
 import zaza.model
-import zaza.openstack.utilities.juju as zaza_juju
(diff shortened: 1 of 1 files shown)


Gleland:
[No-op change](https://review.opendev.org/c/openstack/charm-designate/+/773547) to test designate

ajkavanagh (approved):
LGTM

PR #497, opened 2021-02-03
Zaza_juju rename

Gleland:
Renaming `zaza_juju` in a few import statements to be consistent with the `juju_utils` name across the rest of the repository.


ajkavanagh (approved):
LGTM - this is great 'code gardening' ... it's lovely to see some technical debt being burned down.  Thank you for looking into this.

PR #500, opened 2021-02-07
Add test for the new nova_cc charm action

ionutbalutoiu:
The new action is meant to sync the availability zones from the
underlying Juju provider with the OpenStack availability zones.

It is useful within MAAS environments to map the MAAZ AZs to
OpenStack AZs.

### DIFF
--- a/zaza/openstack/charm_tests/nova/tests.py
+++ b/zaza/openstack/charm_tests/nova/tests.py
@@ -178,6 +178,52 @@ class NovaComputeActionTest(test_utils.OpenStackBaseTest):
                     "The action failed: {}".format(action.data["message"]))
 
 
+class NovaCloudControllerActionTest(test_utils.OpenStackBaseTest):
+    """Run nova-cloud-controller specific tests.
+
+    Add this test class for new nova-cloud-controller action
+    to avoid breaking older version.
+    """
+
+    def test_sync_compute_az_action(self):
+        """Test sync-compute-availability-zones action."""
+        juju_units_az_map = {}
+        for unit in zaza.model.get_units('nova-compute',
+                                         model_name=self.model_name):
+            result = zaza.model.run_on_unit(unit.name,
+                                            'echo $JUJU_AVAILABILITY_ZONE',
+                                            model_name=self.model_name,
+                                            timeout=60)
+            self.assertEqual(int(result['Code']), 0)
+            juju_units_az_map[unit.public_address] = result['Stdout'].strip()
+            continue
+
+        session = openstack_utils.get_overcloud_keystone_session()
+        nova = openstack_utils.get_nova_session_client(session)
+        expected_output = ''
+        for hypervisor in nova.hypervisors.list():
+            az = juju_units_az_map.get(hypervisor.host_ip, None)
+            expected_output += ('Hypervisor {} added to '
+                                'availability zone {}\n'.format(
+                                    hypervisor.hypervisor_hostname, az))
+
+        result = zaza.model.run_action_on_leader(
+            'nova-cloud-controller',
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -178,6 +178,52 @@ def test_virsh_audit_action(self):
...
+
+        session = openstack_utils.get_overcloud_keystone_session()
+        nova = openstack_utils.get_nova_session_client(session)
+        expected_output = ''
+        for hypervisor in nova.hypervisors.list():
+            az = juju_units_az_map.get(hypervisor.host_ip, None)
+            expected_output += ('Hypervisor {} added to '
+                                'availability zone {}\n'.format(
+                                    hypervisor.hypervisor_hostname, az))
+
```
AurelienLourot, line 209: should we assert before running the action that the AZs are different? So that we're sure that the action is needed and doing something?

zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -178,6 +178,52 @@ def test_virsh_audit_action(self):
...
+                                            timeout=60)
+            self.assertEqual(int(result['Code']), 0)
+            juju_units_az_map[unit.public_address] = result['Stdout'].strip()
+            continue
+
+        session = openstack_utils.get_overcloud_keystone_session()
+        nova = openstack_utils.get_nova_session_client(session)
+        expected_output = ''
+        for hypervisor in nova.hypervisors.list():
+            az = juju_units_az_map.get(hypervisor.host_ip, None)
```
AurelienLourot, line 205: Should we assert that it's not `None` everywhere? Because if have no AZ at all then I think the entire test is passing but not covering as much as we want.

AurelienLourot (changes requested):
LGTM in general, thanks! Note that this test is currently failing in https://review.opendev.org/c/openstack/charm-nova-cloud-controller/+/773929 though

zaza/openstack/charm_tests/nova/tests.py
wolsen: I think checking the output itself is somewhat brittle. Discussing this with David, we reached the conclusion that we should just check that the action is successful since the subsequent code tests to make sure the aggregates are setup correctly.

zaza/openstack/charm_tests/nova/tests.py
wolsen: No, it shouldn't be checked. This actually highlighted for me that we're using the wrong setting in the availability zone in the nova-compute charm. I've raised bug https://bugs.launchpad.net/charm-nova-compute/+bug/1925412 and subsequently submitted https://review.opendev.org/c/openstack/charm-nova-compute/+/787470 to fix that.

To fix this, we need to be able to honor both maas and non-maas environments. I've got a branch with commits to handle this at https://github.com/wolsen/zaza-openstack-tests/tree/update-nova-tests/ - which are updates based on this branch/MR. I think it handles the scenario.

ionutbalutoiu: @wolsen

Thank-you for raising (and solving) the issue with non-MAAS environments as well!

If you think it's fine, I'm more than happy to cherry-pick your fixes from:
https://github.com/wolsen/zaza-openstack-tests/tree/update-nova-tests/ 

And test this PR with those changes included.

zaza/openstack/charm_tests/nova/tests.py
ionutbalutoiu: I agree with @wolsen on this one.

AurelienLourot:
Makes sense, thanks! Closing this one in favor of #563

PR #505, opened 2021-02-11
Add test for Glance with Ceph RGW backend

ionutbalutoiu:
This test shall be used for Glance testing bundles with Ceph
deployed, and multi-backend enabled.

It validates that images can be stored using the Ceph RGW (via
the OpenStack Swift API).


wolsen (approved):
LGTM and functional tests pass in related change https://review.opendev.org/c/openstack/charm-glance/+/775197

PR #510, opened 2021-03-03
Reduce duplication in filter functions and enable custom filters

ajkavanagh:
Another set of changes at rationalising and improving the series upgrade tests:

* Reduce duplication of code for filters
* Consolidate/remove tests.py and parallel_tests.py for series upgrade.  There should only be one; make them do the same.


ajkavanagh:
Closing as there's probably not much appetite in doing the merging of parallel and series tests now.

PR #511, opened 2021-03-03
Pin cryptography < 3.4 due to rust compiler requirement

ajkavanagh:
We need to pin the module due to requiring a rust compiler to build it
(or setting an ENV var).  At the moment we are pinning, but we will need
to review this once we've worked out what to do about the rust
requirements.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+cryptography<3.4
 aiounittest
 async_generator
 boto3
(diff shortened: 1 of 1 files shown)


AurelienLourot (changes requested):
I think the comment was useful and should be kept. What do you think?

ajkavanagh:
> I think the comment was useful and should be kept. What do you think?

I think it's very useful!  I'm not sure what happened with this commit.  I was breaking up a series of commits that I had in one branch I was running, and I split this off to another PR (this one).  I think the pin of cryptography had already been done independently and thus, this is a duplicate.  I'm going to drop this PR as it's a duplicate.  Thanks for the spot!

PR #512, opened 2021-03-03
add reload and restart action tests for vault

mastier:
Hi,
these are tests for new feature in the vault charm the actions : reload, restart

LP: https://bugs.launchpad.net/vault-charm/+bug/1916921
CR: https://review.opendev.org/c/openstack/charm-vault/+/778118

### DIFF
--- a/zaza/openstack/charm_tests/vault/tests.py
+++ b/zaza/openstack/charm_tests/vault/tests.py
@@ -244,8 +244,6 @@ class VaultTest(BaseVaultTest):
         Pause service and check services are stopped, then resume and check
         they are started.
         """
-        # Restarting vault process will set it as sealed so it's
-        # important to have the test executed at the end.
         vault_actions = zaza.model.get_actions(
             'vault')
         if 'pause' not in vault_actions or 'resume' not in vault_actions:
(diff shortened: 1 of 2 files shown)


zaza/openstack/charm_tests/vault/tests.py
```diff
@@ -257,6 +257,60 @@ def test_zzz_pause_resume(self):
...
+    def test_vault_reload(self):
+        """Run reload tests.
+
+        Reload service and check services were restarted
+        by doing simple change in the running config by API.
+        Then confirm that service is not sealed
+        """
+        vault_actions = zaza.model.get_actions(
+            'vault')
+        if 'reload' not in vault_actions:
```
coreycb, line 269: Since these are new tests that will only get run by the master charm-vault charm I don't think we need to test if the restart action exists.

zaza/openstack/charm_tests/vault/tests.py
```diff
@@ -257,6 +257,60 @@ def test_zzz_pause_resume(self):
...
+    def test_vault_restart(self):
+        """Run pause and resume tests.
+
+        Restart service and check services are started.
+        """
+        # Restarting vault process will set it as sealed so it's
+        # important to have the test executed at the end.
+        vault_actions = zaza.model.get_actions(
+            'vault')
+        if 'restart' not in vault_actions:
```
coreycb, line 302: Since these are new tests that will only get run by the master charm-vault charm I don't think we need to test if the restart action exists.

zaza/openstack/charm_tests/vault/tests.py
mastier: I have followed the same approach as for the rest of the tests

zaza/openstack/charm_tests/vault/tests.py
```diff
@@ -257,6 +257,60 @@ def test_zzz_pause_resume(self):
...
+            vault_utils.get_running_config(lead_client)[
+                'data']['disable_mlock'])
+        self.assertFalse(lead_client.hvac_client.seal_status['sealed'])
+
+    def test_vault_restart(self):
+        """Run pause and resume tests.
+
+        Restart service and check services are started.
+        """
+        # Restarting vault process will set it as sealed so it's
```
ChrisMacNaughton, line 298: Rather than hoping that this test is run last, it would be better to clean up after itself by unsealing Vault at the end of the test

mastier: Thanks for pointing that out. That was stuping copy paste from pause resume. Actually the vault gets unselead after every test, what you spotted, in the BaseVaultTest class by tearDown(). 
So I will just remove these misleading comments and we are good.

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -137,6 +137,20 @@ def get_vip_client(cacert=None):
...
+    """Get Vault running config.
+
+    :param client: Client to use for initiliasation
+    :type client: CharmVaultClient
+    The hvac library does not support getting info
+    from endpoint /v1/sys/config/state/sanitized
+    Therefore we implement it here
+    """
+    return requests.get(
+        client.hvac_client.adapter.base_uri + '/v1/sys/config/state/sanitized',
```
ChrisMacNaughton, line 150: I'm sad that this isn't available in the hvac library :-/

ChrisMacNaughton (approved):
Looks good to me

zaza/openstack/charm_tests/vault/utils.py
mastier: me too :crying_cat_face:

Issue #348, opened 2020-07-06
octavia: Add more resilience/retry handling in the lb-creation code


fnordahl:
In one example we can see that just after successful creation of a load balancer using one provider driver, the next attempt using a different provider driver failed due to connection dropped while talking to the API.

Nothing in the crash dump bears indication of any services being restarted, VIPs moved etc, so this appears like a fluke and reality of interconnected clustered systems.

The test code should be more resilient and retry under such circumstances.

```
2020-07-04 17:15:52 [INFO] Found "This is the default welcome page" in page retrieved through load balancer  (provider="ovn") at "http://172.17.110.224/"
2020-07-04 17:15:52 [INFO] Creating loadbalancer with provider amphora
2020-07-04 17:16:02 [INFO] Awaiting loadbalancer to reach provisioning_status "ACTIVE"
2020-07-04 17:17:33 [INFO] ERROR
2020-07-04 17:17:33 [INFO] ======================================================================
2020-07-04 17:17:33 [INFO] ERROR: test_create_loadbalancer (zaza.openstack.charm_tests.octavia.tests.LBAASv2Test)
2020-07-04 17:17:33 [INFO] Create load balancer.
2020-07-04 17:17:33 [INFO] ----------------------------------------------------------------------
2020-07-04 17:17:33 [INFO] Traceback (most recent call last):
2020-07-04 17:17:33 [INFO]   File "/tmp/tmp.K630hNXERG/func/lib/python3.5/site-packages/urllib3/connectionpool.py", line 677, in urlopen
2020-07-04 17:17:33 [INFO]     chunked=chunked,
2020-07-04 17:17:33 [INFO]   File "/tmp/tmp.K630hNXERG/func/lib/python3.5/site-packages/urllib3/connectionpool.py", line 426, in _make_request
2020-07-04 17:17:33 [INFO]     six.raise_from(e, None)
2020-07-04 17:17:33 [INFO]   File "<string>", line 3, in raise_from
2020-07-04 17:17:33 [INFO]   File "/tmp/tmp.K630hNXERG/func/lib/python3.5/site-packages/urllib3/connectionpool.py", line 421, in _make_request
2020-07-04 17:17:33 [INFO]     httplib_response = conn.getresponse()
2020-07-04 17:17:33 [INFO]   File "/usr/lib/python3.5/http/client.py", line 1225, in getresponse
2020-07-04 17:17:33 [INFO]     response.begin()
[...]

PR #514, opened 2021-03-03
Object retrier

ajkavanagh:
This adds a wrapper class that detects if a callable object in any of
the descendent objects raises an Exception.  If so, then it retries that
exception.

This is to attempt to make the zaza tests a little more robust in the
face of small network failures or strange restarts.  This is a test, and
robust logging a reporting should be used to determine whether it is
covering up actual bugs rather than CI system issues.

Related Bug: (zot repo)#348

### DIFF
--- a/zaza/openstack/charm_tests/octavia/tests.py
+++ b/zaza/openstack/charm_tests/octavia/tests.py
@@ -25,7 +25,7 @@ import osc_lib.exceptions
 import zaza.openstack.charm_tests.test_utils as test_utils
 import zaza.openstack.utilities.openstack as openstack_utils
 
-from zaza.openstack import ObjectRetrier
+from zaza.openstack.utilities import ObjectRetrier
 
 
 class CharmOperationTest(test_utils.OpenStackBaseTest):
--- a/zaza/openstack/charm_tests/octavia/tests.py
+++ b/zaza/openstack/charm_tests/octavia/tests.py
@@ -25,7 +25,7 @@ import osc_lib.exceptions
 import zaza.openstack.charm_tests.test_utils as test_utils
 import zaza.openstack.utilities.openstack as openstack_utils
 
-from zaza.openstack.utilities import ObjectRetrier
+from zaza.openstack.utilities import ObjectRetrierWraps
 
 
 class CharmOperationTest(test_utils.OpenStackBaseTest):
@@ -65,11 +65,11 @@ class LBAASv2Test(test_utils.OpenStackBaseTest):
     def setUpClass(cls):
         """Run class setup for running LBaaSv2 service tests."""
         super(LBAASv2Test, cls).setUpClass()
-        cls.keystone_client = ObjectRetrier(
+        cls.keystone_client = ObjectRetrierWraps(
             openstack_utils.get_keystone_session_client(cls.keystone_session))
-        cls.neutron_client = ObjectRetrier(
+        cls.neutron_client = ObjectRetrierWraps(
             openstack_utils.get_neutron_session_client(cls.keystone_session))
-        cls.octavia_client = ObjectRetrier(
+        cls.octavia_client = ObjectRetrierWraps(
             openstack_utils.get_octavia_session_client(cls.keystone_session))
         cls.RESOURCE_PREFIX = 'zaza-octavia'
 
--- a/zaza/openstack/charm_tests/octavia/tests.py
+++ b/zaza/openstack/charm_tests/octavia/tests.py
@@ -25,6 +25,8 @@ import osc_lib.exceptions
 import zaza.openstack.charm_tests.test_utils as test_utils
 import zaza.openstack.utilities.openstack as openstack_utils
 
+from zaza.openstack import ObjectRetrier
+
 
 class CharmOperationTest(test_utils.OpenStackBaseTest):
     """Charm operation tests."""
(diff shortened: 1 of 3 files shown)


ajkavanagh:
Tested with https://review.opendev.org/c/openstack/charm-octavia/+/778003 (which has now merged)

zaza/openstack/utilities/__init__.py
```diff
@@ -13,3 +13,154 @@
...
+        backoff = self.__kwargs['backoff']
+        total_wait = self.__kwargs['total_wait']
+        num_retries = self.__kwargs['num_retries']
+        retry_exceptions = self.__kwargs['retry_exceptions']
+        wait_so_far = 0
+        while True:
+            try:
+                return obj(*args, **kwargs)
+            except Exception as e:
+                # if retry_exceptions is None, or the type of the exception is
```
thedac, line 125: is not None correct?

zaza/openstack/charm_tests/octavia/tests.py
```diff
@@ -63,12 +65,12 @@ class LBAASv2Test(test_utils.OpenStackBaseTest):
...
-        cls.neutron_client = openstack_utils.get_neutron_session_client(
-            cls.keystone_session)
-        cls.octavia_client = openstack_utils.get_octavia_session_client(
-            cls.keystone_session)
+        cls.keystone_client = ObjectRetrier(
+            openstack_utils.get_keystone_session_client(cls.keystone_session))
+        cls.neutron_client = ObjectRetrier(
+            openstack_utils.get_neutron_session_client(cls.keystone_session))
+        cls.octavia_client = ObjectRetrier(
+            openstack_utils.get_octavia_session_client(cls.keystone_session))
```
thedac, line 73: I am a bit surprised to see these at the setUpClass level. Are we seeing failures here?

thedac (commented):
Not blocking but a few thoughts. The ObjectRetrier call really feels like a re-implementation of tenacity. Is there a reason you are not using tenacity here?

I was a bit horrified to see the first usage was at the setUpClass level. Are we truly seeing failures at this stage? That would seem to indicate the overcloud is not really ready yet. Does it not?

It feels like we will want this kind of thing everywhere. Should we consider wrapping the openstack clients and importing the wrapped versions rather than the direct clients?

zaza/openstack/charm_tests/octavia/tests.py
ajkavanagh: Ah, I may not have explained how it works properly.  The `__init__(...)` method does the wrapping of the object, but it's when then client is *used* that the retry pattern is activated.  e.g.

```python
# wrap a client that is used to make calls.
client = ObjectRetrier(some_object_instance)
...
# later
...
# This is where the retries are done if needed
client.do_call()
```

zaza/openstack/utilities/__init__.py
ajkavanagh: Yup, typo; thanks for picking this up.

ajkavanagh:
> Not blocking but a few thoughts. The ObjectRetrier call really feels like a re-implementation of tenacity. Is there a reason you are not using tenacity here?

Why didn't I use tenacity?  It's a fair question.  It was mostly about having more obvious (in the code) control over how the exceptions from the wrapped code are handled and determine whether a retry will happen and then either eating or propagating that exception; I felt that tenacity was hiding it too much and also it was quite awkward to set it up.

> I was a bit horrified to see the first usage was at the setUpClass level. Are we truly seeing failures at this stage? That would seem to indicate the overcloud is not really ready yet. Does it not?

No, that's just the setup code to wrap the object; no retries happening at that point - please see the inline comment; I can see how the naming of the class might make people think that *that* is when the retries happen.  Maybe the class should be renamed to `SetupObjectRetrier` or something like that that telegraphs that it's a setup phase, rather than when the retries happen.  Alternatively, I can void the `__init__()` to a noop and use a class method (say `_setup_`) to do the wrap.  However, I was trying to be minimally invasive of the namespace of the wrapped object as all 'calls' go to it via the retry mechanism.  Hence renaming the class may be the better idea.

> It feels like we will want this kind of thing everywhere. Should we consider wrapping the openstack clients and importing the wrapped versions rather than the direct clients?

Valid point.  My purpose here (in this PR) was to introduce the code for the `ObjectRetrier`, show how it can be used, and test as fix for the Octavia tests.  I didn't want to introduce it on everything (yet) as I wanted to see how it might impact our 'worst' offender set of tests first! :)  What do you think?

zaza/openstack/charm_tests/octavia/tests.py
thedac: This makes complete sense. I am not sure why I didn't see it at review time.

thedac:
> > Not blocking but a few thoughts. The ObjectRetrier call really feels like a re-implementation of tenacity. Is there a reason you are not using tenacity here?
> 
> Why didn't I use tenacity? It's a fair question. It was mostly about having more obvious (in the code) control over how the exceptions from the wrapped code are handled and determine whether a retry will happen and then either eating or propagating that exception; I felt that tenacity was hiding it too much and also it was quite awkward to set it up.

I am torn on this since we have the pattern of using tenacity everywhere else. But as I as, I am not blocking on this.

> > I was a bit horrified to see the first usage was at the setUpClass level. Are we truly seeing failures at this stage? That would seem to indicate the overcloud is not really ready yet. Does it not?
> 
> No, that's just the setup code to wrap the object; no retries happening at that point - please see the inline comment; I can see how the naming of the class might make people think that _that_ is when the retries happen. Maybe the class should be renamed to `SetupObjectRetrier` or something like that that telegraphs that it's a setup phase, rather than when the retries happen. Alternatively, I can void the `__init__()` to a noop and use a class method (say `_setup_`) to do the wrap. However, I was trying to be minimally invasive of the namespace of the wrapped object as all 'calls' go to it via the retry mechanism. Hence renaming the class may be the better idea.

Makes complete sense now.

> > It feels like we will want this kind of thing everywhere. Should we consider wrapping the openstack clients and importing the wrapped versions rather than the direct clients?
> 
> Valid point. My purpose here (in this PR) was to introduce the code for the `ObjectRetrier`, show how it can be used, and test as fix for the Octavia tests. I didn't want to introduce it on everything (yet) as I wanted to see how it might impact our 'worst' offender set of tests first! :) What do you think?

I appreciate this is the first round and is targeted. It makes sense to land a lighter weight change first.

thedac (changes requested):
To be clear we are just waiting on the typo fix to land this.

thedac (approved):
Looks good. Landing. Thanks for the work.

PR #519, opened 2021-03-08
remove python 3.4 support

ChrisMacNaughton:


.github/workflows/tox.yaml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.4, 3.5, 3.6, 3.7, 3.8, 3.9]
+        python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
```
ajkavanagh, line 12: I'm assuming we can't *get* py3.4 here or something?

ChrisMacNaughton: python3.4 shipped with Trusty, and 3.5 shipped with Xenial. I'm sure there's some way that we could conceivably get 3.4 on here but I can't convince myself that it's worth it to support a developer running Zaza tests on Trusty :-)

PR #521, opened 2021-03-09
Fix the kerberos set-up by loosening ubuntu check

ajkavanagh:
The cs:ubuntu charm's workload-status-message has changed recently (as
the charm was re-written).  This PR loosens the application states check
to ignore the workload-status-message and just wait for the charm to be
active workload-status.  This is almost certainly enough, as it's a very
simple charm.


ChrisMacNaughton (approved):
Looks good to me

PR #532, opened 2021-03-26
[WIP] Functional testing for glance S3 backend

LCVcode:
Currently in a temporary but working state.

Starting this PR now, so it can be included in gerrit tests.

### DIFF
--- a/zaza/openstack/charm_tests/glance/setup.py
+++ b/zaza/openstack/charm_tests/glance/setup.py
@@ -14,9 +14,14 @@
 
 """Code for configuring glance."""
 
+import json
+import boto3
 import logging
+
+import zaza.model as model
 import zaza.openstack.utilities.openstack as openstack_utils
 import zaza.utilities.deployment_env as deployment_env
+import zaza.utilities.juju as juju
 
 CIRROS_IMAGE_NAME = "cirros"
 CIRROS_ALT_IMAGE_NAME = "cirros_alt"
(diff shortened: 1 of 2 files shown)


zaza/openstack/charm_tests/glance/setup.py
```diff
@@ -161,3 +166,85 @@ def add_lts_image(glance_client=None, image_name=None, release=None,
...
+         's3-store-secret-key': secret_key,
+         's3-store-bucket': 'glance-s3-test-bucket'})
+    model.wait_for_agent_status()
+    model.wait_for_application_states(
+        states={
+            'glance': {
+                'workload-status': 'active',
+                'workload-status-message': 'Unit is ready'}})
+
+    # Install python3-boto3
```
coreycb, line 235: I'm a little confused on this part. Does python3-boto3 need to be installed on the glance unit and if so should the charm be doing that?

zaza/openstack/charm_tests/glance/tests.py
```diff
@@ -92,3 +96,45 @@ def test_901_pause_resume(self):
...
+
+        # Create boto client and verify image exists
+        logging.info('Verifying image exists in S3 bucket')
+        s3_client = boto3.client(
+            's3',
+            aws_access_key_id=access_key,
+            aws_secret_access_key=secret_key,
+            endpoint_url=endpoint)
+        objects = s3_client.list_objects(Bucket='glance-s3-test-bucket')
+        self.assertTrue('Contents' in objects.keys())
```
coreycb, line 139: Would it make sense to expand this test just a little bit and add an object to the bucket?

coreycb (changes requested):
Thanks Connor, just a couple of comments.

zaza/openstack/charm_tests/glance/setup.py
nobuto-m: Agreed. I've left a comment in the charm review:
https://review.opendev.org/c/openstack/charm-glance/+/781748/11#message-096873b259f0e6cc049ee0643712132111dfcfd7

zaza/openstack/charm_tests/glance/tests.py
LCVcode: Hi Corey, sorry for slow reply.  I agree with expanded checks here. I just put this in initially to have a check in place.  I need to include checks such as object name, size, etc.

AurelienLourot:
Superseded by #574, thanks!

PR #535, opened 2021-03-31
Functional tests for `instance-count` action on nova-compute charm.

mkalcok:
Launchpad: [#1911011](https://bugs.launchpad.net/charm-nova-compute/+bug/1911011)
Gerrit review: https://review.opendev.org/c/openstack/charm-nova-compute/+/783371


AurelienLourot (changes requested):
This looks great, thanks! I'm just blocking it right now because we're in code freeze but I intend to land it right after freeze.

PR #538, opened 2021-04-08
Disable test_110_force_quorum_using_partition_of

AurelienLourot:
Temporarily disabling this test in stable/21.04 because
the charm feature hasn't landed yet

Making use of it in the 21.04 library freeze batch: https://review.opendev.org/c/openstack/charm-mysql-innodb-cluster/+/784671


ajkavanagh (approved):
LGTM

PR #540, opened 2021-04-08
Deferred restart test fixes

gnuoy:

PR #541, opened 2021-04-08
Deferred restart test fixes

gnuoy:


AurelienLourot (changes requested):
So this is supposed to be a backport of #540, which I may squash-merge, so I'd rather close this pull-request, wait for #540 to be merged, and then re-open a cherry-pick/backport pull-request to `stable/21.04`.

It's the same rule as for charm contributions basically. We don't want to see backport reviews before the actual fix has been landed in master, because of the failures we know arise out of having these two reviews in parallel.

AurelienLourot:
Replaced by #548

PR #543, opened 2021-04-08
Fix Ceph service action tests

### LINKED ISSUE #542: Ceph OSD ServiceTest.test_start_specific is faulty
The test case uses the command 'systemctl stop ceph-osd.target'  in order to stop all of the OSDs and then attempts to start a single OSD service. Unfortunately, the systemd target is a means to group a set of related services that should run together. Thus, when starting a single OSD other OSDs are also started.

The test expects that only the OSD service that was specifically started is running, however they all start due to their relation to the ceph-osd.target. Fix this by stopping all of the ceph-osd services individually.

wolsen:
The ServiceTest.test_start_specific under the ceph osd tests stops all
services using ceph-osd.target. This will stop all the OSDs, but the
restarting of any single OSD will restart all the OSDs. This behavior
causes the charm test to fail.

Unfortunately, this wasn't caught in the initial review because the test
is skipped when there are fewer the 2 OSDs on a unit and the final test
run skipped the test due to this condition.

Fixes #542

Signed-off-by: Billy Olsen <billy.olsen@gmail.com>

### DIFF
--- a/zaza/openstack/charm_tests/ceph/osd/tests.py
+++ b/zaza/openstack/charm_tests/ceph/osd/tests.py
@@ -258,8 +258,11 @@ class ServiceTest(unittest.TestCase):
         to_start = should_stop.pop()
         should_stop = [service.name for service in should_stop]
 
+        # Note: can't stop ceph-osd.target as restarting a single OSD will
+        # cause this to start all of the OSDs when a single one starts.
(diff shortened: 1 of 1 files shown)


AurelienLourot (approved):
LGTM, thanks! Validated in https://review.opendev.org/c/openstack/charm-ceph-osd/+/785555

PR #545, opened 2021-04-09
Fix Ceph service action tests

### LINKED ISSUE #542: Ceph OSD ServiceTest.test_start_specific is faulty

AurelienLourot:
Backport of #543 (validated [here](https://review.opendev.org/c/openstack/charm-ceph-osd/+/785555)) to stable/21.04.
Fixes #542 and unblocks the [libraries-sync-21-04 batch](https://review.opendev.org/q/topic:libraries-sync-21-04)

### DIFF
--- a/zaza/openstack/charm_tests/ceph/osd/tests.py
+++ b/zaza/openstack/charm_tests/ceph/osd/tests.py
@@ -258,8 +258,11 @@ class ServiceTest(unittest.TestCase):
         to_start = should_stop.pop()
         should_stop = [service.name for service in should_stop]
 
+        # Note: can't stop ceph-osd.target as restarting a single OSD will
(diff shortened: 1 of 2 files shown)


AurelienLourot:
Targetted the wrong branch

AurelienLourot:
Replaced by #546

PR #582, opened 2021-05-28
added glance functional test for image-conversion config

hernandanielg:
Functional test for image-conversion feature added in https://review.opendev.org/c/openstack/charm-glance/+/792065

Test does:
* Validates that test runs only on supported versions
* Enables config image-conversion 
* Wait for unit to settle
* Creates, stages and imports an image and waits until it's imported
* Assert that image format is `raw`

### DIFF
--- a/zaza/openstack/charm_tests/glance/tests.py
+++ b/zaza/openstack/charm_tests/glance/tests.py
@@ -67,6 +67,32 @@ class GlanceTest(test_utils.OpenStackBaseTest):
             {'image_format': {'disk_formats': ['qcow2']}},
             ['glance-api'])
 
+    def test_412_image_conversion(self):
+        """Check image-conversion config.
+
+        When image-conversion config is enabled glance will convert images
+        to raw format, this is only performed for interoperable image import
+        docs.openstack.org/glance/train/admin/interoperable-image-import.html
+        image conversion is done at server-side for better image handling
+        """
+        current_release = openstack_utils.get_os_release()
+        bionic_stein = openstack_utils.get_os_release('bionic_stein')
+        if current_release < bionic_stein:
+            self.skipTest('image-conversion config is supported since '
+                          'bionic_stein or newer versions')
+
+        with self.config_change({'image-conversion': 'false'},
+                                {'image-conversion': 'true'}):
+            image_url = openstack_utils.find_cirros_image(arch='x86_64')
+            image = openstack_utils.create_image(
+                self.glance_client,
+                image_url,
+                'cirros-test-import',
+                force_import=True)
+
+            disk_format = self.glance_client.images.get(image.id).disk_format
+            self.assertEqual('raw', disk_format)
+
     def test_900_restart_on_config_change(self):
         """Checking restart happens on config change."""
         # Config file affected by juju set config change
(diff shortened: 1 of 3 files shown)


zaza/openstack/charm_tests/glance/tests.py
```diff
@@ -67,6 +68,38 @@ def test_411_set_disk_format(self):
             {'image_format': {'disk_formats': ['qcow2']}},
             ['glance-api'])
 
+    def test_412_image_conversion(self):
```
AurelienLourot, line 71: Test cases should make an effort to restore the system under test to how it was before the test was run, so that the order in which all the tests are run doesn't really matter. So this test should revert the config change at the end I would say, if possible.

This is something we do a lot in other tests and it looks like:

```
with self.config_change(...):
    ...
```

What do you think?

zaza/openstack/utilities/openstack.py
```diff
@@ -2444,6 +2444,45 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
...
+    """
+    image = glance.images.create(
+        name=image_name,
+        disk_format=disk_format,
+        visibility=visibility,
+        container_format=container_format)
+    glance.images.stage(image.id, open(local_path, 'rb'))
+    glance.images.image_import(image.id, method='glance-direct')
+
+    resource_reaches_status(
```
AurelienLourot, line 2477: As discussed offline, this probably doesn't wait long enough

zaza/openstack/utilities/openstack.py
```diff
@@ -2444,6 +2444,45 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
     return image
 
 
+def import_image_to_glance(glance, local_path, image_name,
+                           disk_format='qcow2', visibility='public',
+                           container_format='bare'):
+    """Creates, stages and import an image.
```
AurelienLourot, line 2450: I'm not an Glance expert: can you explain in the docstring a bit the difference with `upload_image_to_glance()`? What's the difference between "uploading" and "staging/importing" an image?

zaza/openstack/utilities/openstack.py
```diff
@@ -2501,6 +2540,50 @@ def create_image(glance, image_url, image_name, image_cache_dir=None, tags=[],
...
+                  '({})...'.format(image_name))
+
+    img_name = os.path.basename(urllib.parse.urlparse(image_url).path)
+    local_path = os.path.join(image_cache_dir, img_name)
+
+    if not os.path.exists(local_path):
+        logging.info('Downloading {} ...'.format(image_url))
+        download_image(image_url, local_path)
+    else:
+        logging.info('Cached image found at {} - Skipping download'.format(
```
AurelienLourot, line 2576: you're duplicating code from `create_image()`. Can we refactor in order not to duplicate code?

AurelienLourot (changes requested):
Thanks for working on this! See my comments

zaza/openstack/charm_tests/glance/tests.py
hernandanielg: Great advice! thanks :)

zaza/openstack/utilities/openstack.py
hernandanielg: Images are getting cached so no need to extend waiting time ;-)

AurelienLourot (approved):

AurelienLourot:
This needs to be merged at the same time as https://review.opendev.org/c/openstack/charm-glance/+/792065 , when ready

PR #586, opened 2021-06-03
gss: Log Keystone service catalog on failure

fnordahl:
The simplestreams service relies on the Keystone Service catalog.
To help debug any issues occurring during test execution, log the
contents of the Keystone service catalog in the event of failure.

Related-Bug: #1930654


ChrisMacNaughton (approved):
Looks reasonable

PR #588, opened 2021-06-08
Handle Vault running tests twice

thedac:
Vault's runs its tests twice. In some circumstances the removal of the
vault key from the test_config dictionary can throw a KeyError. Do not
fail when this happens.


ajkavanagh (approved):
LGTM

PR #589, opened 2021-06-09
Write private SSH key with appropriate permissions

### LINKED ISSUE #583: Developer ergonomics: id_rsa_zaza created with too loose permissions
Deployments created by Zaza are short lived test artifacts, so this is not a security issue.

However, if I as a developer want to interact with a deployment set up by a Zaza test run I would get into this situation:
```bash
$ ssh -i /tmp/zaza-be6c6eab4a3b/id_rsa_zaza ubuntu@10.78.95.90
The authenticity of host '10.78.95.90 (10.78.95.90)' can't be established.
ECDSA key fingerprint is SHA256:qwsUztWvCWDX9hJuXylFpGwa4E3EQZbXSV6K86V8jgA.
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added '10.78.95.90' (ECDSA) to the list of known hosts.
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0664 for '/tmp/zaza-be6c6eab4a3b/id_rsa_zaza' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
Load key "/tmp/zaza-be6c6eab4a3b/id_rsa_zaza": bad permissions
ubuntu@10.78.95.90: Permission denied (publickey).
```

Creating the file with the mode the ssh client expects would improve this experience.

fnordahl:
Fixes #583

### DIFF
--- a/zaza/openstack/utilities/openstack.py
+++ b/zaza/openstack/utilities/openstack.py
@@ -2768,8 +2768,14 @@ def write_private_key(keypair_name, key):
     :param key: PEM Encoded Private Key
     :type key: str
     """
-    with open(get_private_key_file(keypair_name), 'w') as key_file:
-        key_file.write(key)
+    # Create the key file with mode 0o600 to allow the developer to pass it to
+    # the `ssh` command without getting a "bad permissions" error.
+    stored_umask = os.umask(0o177)
+    try:
+        with open(get_private_key_file(keypair_name), 'w') as key_file:
+            key_file.write(key)
+    finally:
+        os.umask(stored_umask)
 
 
 def get_private_key(keypair_name):
--- a/unit_tests/utilities/test_zaza_utilities_openstack.py
+++ b/unit_tests/utilities/test_zaza_utilities_openstack.py
@@ -633,14 +633,35 @@ class TestOpenStackUtils(ut_utils.BaseTestCase):
     def test_write_private_key(self):
         self.patch_object(openstack_utils.deployment_env, 'get_tmpdir',
                           return_value='/tmp/zaza-model1')
+        self.patch_object(openstack_utils.os, 'umask',
+                          return_value='fakeumask')
         m = mock.mock_open()
         with mock.patch(
             'zaza.openstack.utilities.openstack.open', m, create=False
         ):
             openstack_utils.write_private_key('mykeys', 'keycontents')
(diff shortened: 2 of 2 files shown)


zaza/openstack/utilities/openstack.py
```diff
@@ -2687,8 +2687,10 @@ def write_private_key(keypair_name, key):
     :param key: PEM Encoded Private Key
     :type key: str
     """
+    stored_umask = os.umask(0o600)
```
fnordahl, line 2690: This is of course inverted and should be something like `0277`

zaza/openstack/utilities/openstack.py
```diff
@@ -2740,9 +2740,15 @@ def write_private_key(keypair_name, key):
...
     :type key: str
     """
+    # Create the key file with mode 0o600 to allow the developer to pass it to
+    # the `ssh` command without getting a "bad permissions" error.
+    stored_umask = os.umask(0o177)
+
     with open(get_private_key_file(keypair_name), 'w') as key_file:
         key_file.write(key)
 
+    os.umask(stored_umask)
```
freyes, line 2750: this should be part of a try-finally so we always restore the umask no matter if the writing failed.

```
stored_umask = os.umask(0o177)
try:
    with open(get_private_key_file(keypair_name), 'w') as key_file:
        key_file.write(key)
finally:
    os.umask(stored_umask)
```

freyes (changes requested):

zaza/openstack/utilities/openstack.py
fnordahl: That is definitively more clean, I avoided adding that because the exception raised would have made zaza crash and exit anyway.

But if someone comes in the future to add a except here, this will probably not be thought of and become a latent bug.

I'll add your suggestion. Thx!

freyes (approved):
LGTM.

PR #590, opened 2021-06-13
Skip test_vault_reload with enabling/disabling mlock in LXD

nobuto-m:
When Vault is running in a LXD container, test_vault_reload is expected
to fail because mlock cannot be enabled in LXD containers. Skip it when
running in containers is detected. It's useful for local development
environments with Juju's localhost/LXD provider. The test won't be
skipped in the OpenStack Charm CI since the Vault units will be run in
KVM, not in LXD containers.

### DIFF
--- a/zaza/openstack/charm_tests/vault/tests.py
+++ b/zaza/openstack/charm_tests/vault/tests.py
@@ -18,6 +18,7 @@
 
 import contextlib
 import hvac
+import json
 import logging
 import time
 import unittest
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/vault/tests.py
```diff
@@ -277,6 +278,15 @@ def test_vault_reload(self):
             raise unittest.SkipTest("The version of charm-vault tested does "
                                     "not have reload action")
 
+        container_results = zaza.model.run_on_leader(
+            "vault", "systemd-detect-virt --container"
+        )
+        container_rc = json.loads(container_results["Code"])
```
gnuoy, line 284: Could you use `zaza.utilities._local_utils.get_provider_type` here instead ?

nobuto-m: Thanks, I didn't know about `get_provider_type`.

It returns `None` with my environment, and there is no clouds.yaml neither on the machine using for localhost/LXD provider or another machine with AWS provider. I'm not sure we can rely on get_provider_type without having clouds.yaml here.

```
2021-06-13 09:40:58 [INFO] Using config /home/nobuto/.local/share/juju/clouds.yaml
None
```

```
$ ls -1F ~/.local/share/juju/
accounts.yaml
bootstrap-config.yaml
controllers.yaml
cookies/
credentials.yaml
lxd/
models.yaml
public-clouds.yaml
ssh/
```

gnuoy: ok, lets go ahead with the patch as it is then

gnuoy (approved):
LGTM, thanks

PR #597, opened 2021-06-29
Ensure application name is passed in scenario tests

gnuoy:
When running Trilio tests in a scenario test there is no 'charm_name'
so the application name needs to be explicitly set when calling
the setup class otherwise some config, such as lead_unit, is
missing.


gnuoy:
Tested with trilio-wlm gate tests locally: https://paste.ubuntu.com/p/zPbcs2JTxg/

ChrisMacNaughton (approved):
LGTM, thanks!

PR #598, opened 2021-07-05
Fix neutron-api-plugin-ovn's num-expected-units

AurelienLourot:
After zaza#451 Zaza's application readiness detection
has been improved, which now exposes issues from the
past that were hidden. In other words, we were relying
on a bug on Zaza's side and need to get this straight.

Gerrit review making use of this: https://review.opendev.org/c/openstack/charm-neutron-gateway/+/799443


ChrisMacNaughton (approved):
LGTM, thanks!

ajkavanagh:
FYI, the other way of doing this is just to delete the 'num-expected-units' key if it is present; this way it will mean 'at least one unit of x' rather than '1 unit of x exactly'.  Maybe the key should have been "expect-units" and be a boolean??

PR #604, opened 2021-07-19
Add connection timeout to CirrOS image download

AurelienLourot:
Without this timeout it has been observed that Zaza
may hang forever on open(). It is better to fail
faster and save time and resources.


AurelienLourot:
Validated together with https://github.com/openstack-charmers/charmed-openstack-tester/pull/45

ajkavanagh (approved):
LGTM

PR #605, opened 2021-07-19
Check blocked ovn-chassis units with bad config

gabrielcocenza:
LP bug#1919481 [0] needs to have a func-test to check if units can
handle bad bridge-interface-mappings configuration and set units
workload to block state and return to active state with good config.
    
[0] https://bugs.launchpad.net/charm-ovn-chassis/+bug/1919481

### DIFF
--- a/zaza/openstack/charm_tests/ovn/tests.py
+++ b/zaza/openstack/charm_tests/ovn/tests.py
@@ -170,25 +170,25 @@ class ChassisCharmOperationTest(BaseCharmOperationTest):
 
     def test_wrong_bridge_config(self):
         """Confirm that ovn-chassis units block with wrong bridge config."""
-        alternate_config = {'bridge-interface-mappings': 'incorrect'}
-        logging.info('set wrong bridge-interface-mappings')
-        zaza.model.set_application_config(
-            self.application_name,
-            self._stringed_value_config(alternate_config),
-            model_name=self.model_name
-        )
-        zaza.model.block_until_unit_wl_status('ovn-chassis/0', 'blocked')
-        zaza.model.block_until_unit_wl_status('ovn-chassis/1', 'blocked')
-        logging.info('ovn-chassis units successfully blocked')
+        stored_target_deploy_status = self.test_config.get(
+            'target_deploy_status', {})
+        new_target_deploy_status = stored_target_deploy_status.copy()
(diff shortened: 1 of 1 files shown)


fnordahl (changes requested):
Thank you for your work on this. What the test achieves looks good, but it's implementation is a bit repetitive. I would love to see the same test utilize the common config_changed helper:

https://github.com/openstack-charmers/zaza-openstack-tests/blob/57ca7c68e0c90f64545a3ec6b7d8cea9a1347dbe/zaza/openstack/charm_tests/test_utils.py#L265

There are many examples of its use in this repository.

gabrielcocenza:
> Thank you for your work on this. What the test achieves looks good, but it's implementation is a bit repetitive. I would love to see the same test utilize the common config_changed helper:
> 
> https://github.com/openstack-charmers/zaza-openstack-tests/blob/57ca7c68e0c90f64545a3ec6b7d8cea9a1347dbe/zaza/openstack/charm_tests/test_utils.py#L265
> 
> There are many examples of its use in this repository.

I actually tried to use the `config_change` before doing this way, but I faced a problem that the states that I want to test are not the same as the `target_deploy_status` used in this function and they were hanging forever.

https://github.com/openstack-charmers/zaza-openstack-tests/blob/57ca7c68e0c90f64545a3ec6b7d8cea9a1347dbe/zaza/openstack/charm_tests/test_utils.py#L328-L330

Should I try to change the attribute `test_config` before running this test?

fnordahl:
> Should I try to change the attribute `test_config` before running this test?

Yes, that would work, there is an example of that here:
https://github.com/openstack-charmers/zaza-openstack-tests/blob/57ca7c68e0c90f64545a3ec6b7d8cea9a1347dbe/zaza/openstack/charm_tests/ceph/tests.py#L992-L1016

gabrielcocenza:
Apparently, everything went well with this way you showed.

https://pastebin.canonical.com/p/YnNQrvdvHb/

Thanks for the tip.

fnordahl (approved):
LGTM, thanks!

fnordahl:
Now that #608 has merged the tests should pass once this PR is rebased.

fnordahl (approved):
LGTM, thanks! Validated in https://review.opendev.org/c/x/charm-ovn-chassis/+/801525

PR #607, opened 2021-07-21
Wait between actions on OpenStack upgrade

AurelienLourot:
We're more likely to time out while waiting for the
action's result if we launch an action while the
model is still executing. Thus it's safer to wait
for the model to settle between actions.

Validated locally:

```
$ tox -e openstack-upgrade kitchen-sink-focal-ussuri
[...]
  Deploy Bundle:
    Start: 1626813365.993568
    Finish: 1626813550.7522993
    Elapsed Time: 184.75873136520386
    PCT Of Run Time: 2
  Prepare Environment:
    Start: 1626813363.8050616
    Finish: 1626813365.4636593
    Elapsed Time: 1.658597707748413
    PCT Of Run Time: 1
  Test zaza.openstack.charm_tests.openstack_upgrade.tests.OpenStackUpgradeTestsFocalUssuri:
    Start: 1626818712.0522335
    Finish: 1626822851.7874122
    Elapsed Time: 4139.73517870903
    PCT Of Run Time: 44
  Test zaza.openstack.charm_tests.openstack_upgrade.tests.WaitForMySQL:
    Start: 1626818711.1111934
    Finish: 1626818712.0521748
    Elapsed Time: 0.940981388092041
    PCT Of Run Time: 1
  Wait for Deployment:
    Start: 1626813550.7523944
    Finish: 1626816112.8488872
    Elapsed Time: 2562.096492767334
    PCT Of Run Time: 28
Metadata: {}

_____________________________________________________________________________________________________ summary _____________________________________________________________________________________________________
  openstack-upgrade: commands succeeded
  congratulations :)
```

### DIFF
--- a/zaza/openstack/utilities/openstack_upgrade.py
+++ b/zaza/openstack/utilities/openstack_upgrade.py
@@ -132,12 +132,23 @@ def action_upgrade_apps(applications, model_name=None):
             status=status,
             model_name=model_name)
 
+        # NOTE(lourot): we're more likely to time out while waiting for the
+        # action's result if we launch an action while the model is still
+        # executing. Thus it's safer to wait for the model to settle between
+        # actions.
+        zaza.model.block_until_all_units_idle(model_name)
         pause_units(hacluster_units, model_name=model_name)
+
+        zaza.model.block_until_all_units_idle(model_name)
         pause_units(target, model_name=model_name)
 
+        zaza.model.block_until_all_units_idle(model_name)
         action_unit_upgrade(target, model_name=model_name)
 
+        zaza.model.block_until_all_units_idle(model_name)
         resume_units(target, model_name=model_name)
+
+        zaza.model.block_until_all_units_idle(model_name)
         resume_units(hacluster_units, model_name=model_name)
 
         done.extend(target)


AurelienLourot:
Fails on py3.8 with

```
Collecting distro-info
  Downloading distro-info-0.10.tar.gz (5.7 kB)
WARNING: Discarding https://files.pythonhosted.org/packages/91/e0/ccd731d32238bc18bb8ac80353b4028394f80d5ff7db0f93a5ba488690d3/distro-info-0.10.tar.gz#sha256=3c5c2c800e84f36e56e4c59b09981eae357127494f9eb391712888c3f7ac2916 (from https://pypi.org/simple/distro-info/). Requested distro-info from https://files.pythonhosted.org/packages/91/e0/ccd731d32238bc18bb8ac80353b4028394f80d5ff7db0f93a5ba488690d3/distro-info-0.10.tar.gz#sha256=3c5c2c800e84f36e56e4c59b09981eae357127494f9eb391712888c3f7ac2916 (from -r /home/runner/work/zaza-openstack-tests/zaza-openstack-tests/requirements.txt (line 44)) has inconsistent version: filename has '0.10', but metadata has '0.0.0'
ERROR: Could not find a version that satisfies the requirement distro-info (from versions: 0.10)
ERROR: No matching distribution found for distro-info
```

ajkavanagh (commented):
LGTM - this seems like an entirely reasonable change for testing, although it would also be reasonable to me if the units themselves could cope with actions when other things in the model are changing.  But that way probably leads to deadlocks as we have no really way of deferring an action.

AurelienLourot:
@ajkavanagh the problem isn't races or deadlocks, but the actions don't even get started yet that we already time out waiting for their results. Their seems to be a mechanism in juju that queues an action as long as a unit is executing, but we don't take this into account in our timeout mechanism

ajkavanagh:
> @ajkavanagh the problem isn't races or deadlocks, but the actions don't even get started yet that we already time out waiting for their results. Their seems to be a mechanism in juju that queues an action as long as a unit is executing, but we don't take this into account in our timeout mechanism

I was more wondering about whether the charms could be updated to deal with actions failing due to other parts of the model executing.  The fact that we have to wait until the model is 'calm' again, before running an action means that the charms can't "cope" with the model still updating; i.e. hold the action until it is possible.  But seeing as the "system" is Turing-complete, that way leads to all sorts of interesting issues -- so failing in the action (as in "this is undefined behaviour so I give up") is probably the simplest way of dealing with this level of complexity.

ajkavanagh (approved):
LGTM

PR #611, opened 2021-07-29
Neutron gateway actions

mkalcok:
Functional tests for neutron-gateway actions needed by juju-verify.

Related LP issues (1 for each action):
* https://bugs.launchpad.net/charm-neutron-gateway/+bug/1917401
* https://bugs.launchpad.net/charm-neutron-gateway/+bug/1917403
* https://bugs.launchpad.net/charm-neutron-gateway/+bug/1917405

Neutron gateway charm PR: https://review.opendev.org/c/openstack/charm-neutron-gateway/+/782896

### DIFF
--- a/zaza/openstack/charm_tests/neutron/tests.py
+++ b/zaza/openstack/charm_tests/neutron/tests.py
@@ -249,13 +249,13 @@ class NeutronGatewayTest(NeutronPluginApiSharedTests):
         return services
 
 
-class NeutronGatewayStatusActionsTest(test_utils.OpenStackBaseTest):
-    """Test status actions of Neutron Gateway Charm.
+class NeutronGatewayShowActionsTest(test_utils.OpenStackBaseTest):
+    """Test "show" actions of Neutron Gateway Charm.
 
     actions:
-      * get-status-routers
-      * get-status-dhcp
-      * get-status-lb
+      * show-routers
+      * show-dhcp-networks
+      * show-loadbalancers
     """
 
     SKIP_LBAAS_TESTS = True
@@ -263,7 +263,7 @@ class NeutronGatewayStatusActionsTest(test_utils.OpenStackBaseTest):
     @classmethod
     def setUpClass(cls, application_name='neutron-gateway', model_alias=None):
         """Run class setup for running Neutron Gateway tests."""
-        super(NeutronGatewayStatusActionsTest, cls).setUpClass(
+        super(NeutronGatewayShowActionsTest, cls).setUpClass(
             application_name, model_alias)
         # set up clients
         cls.neutron_client = (
@@ -294,8 +294,8 @@ class NeutronGatewayStatusActionsTest(test_utils.OpenStackBaseTest):
         # assert that juju action returned expected resources
         self.assertEqual(result_resource_ids, expected_resource_ids)
 
-    def test_get_status_routers(self):
-        """Test that get-status-routers reports correct neutron routers."""
+    def test_show_routers(self):
+        """Test that show-routers action reports correct neutron routers."""
         # fetch neutron routers using neutron client
         ngw_unit = zaza.model.get_units(self.application_name,
                                         model_name=self.model_name)[0]
@@ -308,14 +308,14 @@ class NeutronGatewayStatusActionsTest(test_utils.OpenStackBaseTest):
 
         # fetch neutron routers using juju-action
         result = zaza.model.run_action(ngw_unit.entity_id,
-                                       'get-status-routers',
+                                       'show-routers',
                                        model_name=self.model_name)
 
         # assert that data from neutron client match data from juju action
         self._assert_result_match(result, routers_from_client, 'router-list')
 
-    def test_get_status_dhcp(self):
-        """Test that get-status-dhcp reports correct DHCP networks."""
+    def test_show_dhcp_networks(self):
+        """Test that show-dhcp-networks reports correct DHCP networks."""
         # fetch DHCP networks using neutron client
         ngw_unit = zaza.model.get_units(self.application_name,
                                         model_name=self.model_name)[0]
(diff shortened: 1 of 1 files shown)


auria (commented):
* I'd recommend removing the backslashes due to col length.
* Could some comments be added, specially for the `run_action` calls? 
* The 3 tests have a repeating pattern to check if `result.status` is `completed` and compare data from the client and the one returned from the action. Maybe this code could be wrapped into a helper function?

Other than the above, tests lgtm.

auria (commented):
Thank you for the refactor. LGTM, although it needs to be approved by an OpenStack team member :)

zaza/openstack/charm_tests/neutron/tests.py
```diff
@@ -248,6 +249,138 @@ def _get_services(cls):
...
+                                current_release >= bionic_train)
+
+    def tearDown(self):
+        """Cleanup loadbalancers if there are any left over."""
+        super(NeutronGatewayStatusActionsTest, self).tearDown()
+        if not self.SKIP_LBAAS_TESTS:
+            load_balancers = self.neutron_client.list_loadbalancers().get(
+                'loadbalancers', [])
+            for lbaas in load_balancers:
+                self.neutron_client.delete_loadbalancer(lbaas['id'])
```
ajkavanagh, line 287: This gets run after each test in the class?  Is this what you want?  I only see one test that uses load balancers and it might be more "obvious" and less resource hungry to just wrap the majority of a test in a try: except: and put this code in a finally: to ensure it gets run?

zaza/openstack/charm_tests/neutron/tests.py
```diff
@@ -248,6 +249,138 @@ def _get_services(cls):
...
+        """Test that get-status-routers reports correct neutron routers."""
+        # fetch neutron routers using neutron client
+        ngw_unit = zaza.model.get_units(self.application_name,
+                                        model_name=self.model_name)[0]
+        routers_from_client = self.neutron_client.list_routers().get(
+            'routers', [])
+
+        if not routers_from_client:
+            self.fail('At least one router must be configured for this test '
+                      'to pass.')
```
ajkavanagh, line 317: Do you want the test to fail, or just to be skipped?

zaza/openstack/charm_tests/neutron/tests.py
```diff
@@ -248,6 +249,138 @@ def _get_services(cls):
...
+        """Test that get-status-dhcp reports correct DHCP networks."""
+        # fetch DHCP networks using neutron client
+        ngw_unit = zaza.model.get_units(self.application_name,
+                                        model_name=self.model_name)[0]
+        networks_from_client = self.neutron_client.list_networks().get(
+            'networks', [])
+
+        if not networks_from_client:
+            self.fail('At least one network must be configured for this test '
+                      'to pass.')
```
ajkavanagh, line 338: As above; should this test be skipped if there is no resource to test against rather than failing?

zaza/openstack/charm_tests/neutron/tests.py
```diff
@@ -248,6 +249,138 @@ def _get_services(cls):
...
+
+        # create LBaasV2 for the purpose of this test
+        lbaas_name = 'test_lbaas'
+
+        subnet_list = self.neutron_client.list_subnets(
+            name='private_subnet').get('subnets', [])
+
+        if not subnet_list:
+            raise RuntimeError('Expected subnet "private_subnet" is not '
+                               'configured.')
```
ajkavanagh, line 363: As above, skip or fail?  The reason for asking, is whether this can be a generic test or is very specific to this environment.

ajkavanagh (commented):
Thanks for the patch; please take a look at my comments; they may not be accurate as I'm just thinking through the consequences of skip vs failure and I'm not sure of the context.  Thanks.

zaza/openstack/charm_tests/neutron/tests.py
mkalcok: Intention was for tests to fail because the model is not properly configured. Unless there's a reason for some test models to not have routers. But none of the func test models had problem with this so far.

zaza/openstack/charm_tests/neutron/tests.py
mkalcok: This is a good point. Indeed the original purpose of this `tearDown` was to ensure that the loadbalancers are removed even if the test that created them failed. I'll fix it.

zaza/openstack/charm_tests/neutron/tests.py
```diff
@@ -274,18 +274,9 @@ def setUpClass(cls, application_name='neutron-gateway', model_alias=None):
         current_release = openstack_utils.get_os_release()
         bionic_train = openstack_utils.get_os_release('bionic_train')
         xenial_mitaka = openstack_utils.get_os_release('xenial_mitaka')
-        cls.SKIP_LBAAS_TESTS = (xenial_mitaka <= current_release or
+        cls.SKIP_LBAAS_TESTS = (current_release <= xenial_mitaka or
                                 current_release >= bionic_train)
```
ajkavanagh: An alternative option/syntax:

```python
 cls.SKIP_LBAAS_TESTS = not (xenial_mitaka > current_release < bionic_train)
```

zaza/openstack/charm_tests/neutron/tests.py
```diff
@@ -352,33 +343,45 @@ def test_get_status_load_balancers(self):
...
-        self._assert_result_match(result, lbaas_from_client, 'load-balancers')
+            result = zaza.model.run_action(ngw_unit.entity_id,
+                                           'get-status-lb',
+                                           model_name=self.model_name,
+                                           action_params={"format": "json"})
+
+            self._assert_result_match(result, lbaas_from_client,
+                                      'load-balancers')
+        except Exception as exc:
+            raise exc
```
ajkavanagh, line 382: This isn't needed; the exception with happen if no exceptions are caught which is essentially the same thing.

ajkavanagh (commented):
Looking good; just a couple of minor comments.

zaza/openstack/charm_tests/neutron/tests.py
mkalcok: thanks, that does look better.

AurelienLourot (commented):
LGTM, thanks! Not giving +2 yet because when addressing comments in the Gerrit review, the actions will be renamed and thus this PR will need the same renamings. As far as I can tell all comments from other reviewers have been addressed.

AurelienLourot (approved):

PR #618, opened 2021-08-11
Switch tempest config to use include and exclude flags

coreycb:


zaza/openstack/charm_tests/tempest/tests.py
```diff
@@ -33,8 +33,12 @@ def run(self):
         """Run tempest tests as specified in tests/tests.yaml.
 
         Test keys are parsed from ['tests_options']['tempest']['model'], where
-        valid test keys are: smoke (bool), whitelist (list of tests), blacklist
-        (list of tests), regex (list of regex's), and keep-workspace (bool).
+        valid test keys are:
+          - smoke (bool)
+          - include-list (list of tests)
+          - exclude-list (list of tests)
+          - regex (list of regex's)
```
AurelienLourot, line 40: we should also mention `exclude-regex` in this list

AurelienLourot (changes requested):
LGTM, thanks, just one nitpick

coreycb:
@AurelienLourot Good catch, thanks. Apologies as I've force pushed that change.

PR #620, opened 2021-08-13
Maas configure one interface

fnordahl:


coreycb (commented):
This looks good to me, thanks Frode. One other thing that I think might be nice is if the data-port is set in the bundle, we could leave it as is. I think you mentioned that might break some things though?

fnordahl:
I think the main problem is that we have enabled the bundles we ship in the charm store with glue to run Zaza on them, and as they are today they do contain a value for `data-port` / `bridge-interface-mappings`, example:

https://github.com/openstack-charmers/openstack-bundles/blob/3f74a67704ed7654e0f574ccfd724dc3a224c9f8/stable/openstack-base/bundle.yaml#L7

So if you run zaza on that bundle with MAAS and zaza refuses to update the config it will most likely not work, depending on what machine you end up deploying to. OTOH, given that is not a problem we hear about every day we may not be using them in that way today anyway.

I guess a compromise could be to add some sort of command line option or environment variable that must be set before the config will be overridden or something like that?

coreycb:
For the openstack-base bundle, we could potentially comment out the bridge-interface-mapping so that we would have something like this:

  ovn-chassis:
    annotations:
      gui-x: '120'
      gui-y: '1030'
    charm: cs:ovn-chassis-14
    options:
      ovn-bridge-mappings: physnet1:br-ex
      # Please update and un-comment the following `bridge-interface-mappings`
      # option to values suitable for the hardware used in your deployment. See
      # the referenced documentation at the top of this file.
      # bridge-interface-mappings: br-ex:eno2

coreycb:
A CLI option might make sense too. I think in that case it would make sense for what's specified in the bundle to be used by default though, and the flag could override with zaza magic.

fnordahl:
Changing those bundles would also work. The bundles we use in neutron-openvswitch, neutron-gateway, ovn-chassis and neutron-api-plugin-ovn gates appear to have `data-port` / `bridge-interface-mappings` empty, so that would support using this as a general guard.

We could also log a warning / informational message when there is a value there in case someone unintentionally trips over the change.

coreycb:
I can work on those changes. How about we land this part now and we can follow up with what we've discussed.

coreycb:
I've proposed the following PRs based no our discussion above:
https://github.com/openstack-charmers/zaza-openstack-tests/pull/625
https://github.com/openstack-charmers/openstack-bundles/pull/219

PR #622, opened 2021-08-16
Add support for restarting and then retesting manila share services

ChrisMacNaughton:
In this change, a charm test that subclasses the ManilaBaseTest
class can also opt-in to re-validating a share after restarting
the share services by overriding the _restart_share_instance
method.

This stable backport is being tested and run in https://review.opendev.org/c/openstack/charm-manila-ganesha/+/802934


ajkavanagh:
@ChrisMacNaughton marking as state.  The associated gerrit review is abandoned; please could you evaluate whether it's still needed.

ChrisMacNaughton:
It was for a backport to 21.04, seems reasonable to drop :)

PR #624, opened 2021-08-18
Adds tests for NRPE checks in Manila

gustavosr98:
Adds a test to ensure that Manila units related to NRPE get service check files created.

Closes-Bug: #1925977

### DIFF
--- a/zaza/openstack/charm_tests/manila/tests.py
+++ b/zaza/openstack/charm_tests/manila/tests.py
@@ -86,13 +86,6 @@ class ManilaTests(test_utils.OpenStackBaseTest):
     def _list_shares(self):
         return self.manila_client.shares.list()
 
-    @tenacity.retry(
-        retry=tenacity.retry_if_result(lambda ret: ret is not None),
-        wait=tenacity.wait_fixed(120),
-        stop=tenacity.stop_after_attempt(2))
-    def _retry_check_commands_on_units(self, cmds, units):
-        return generic_utils.check_commands_on_units(cmds, units)
-
     def test_902_nrpe_service_checks(self):
         """Confirm that the NRPE service check files are created."""
         units = zaza.model.get_units('manila')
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/manila/tests.py
```diff
@@ -85,6 +86,31 @@ def test_manila_api(self):
     def _list_shares(self):
         return self.manila_client.shares.list()
 
+    @tenacity.retry(
+        retry=tenacity.retry_if_result(lambda ret: ret is not None),
+        wait=tenacity.wait_fixed(120),
+        stop=tenacity.stop_after_attempt(2))
+    def _retry_check_commands_on_units(self, cmds, units):
+        return generic_utils.check_commands_on_units(cmds, units)
```
ajkavanagh, line 94: As this is only being used once, it might be more logical to put it inline in the code using a `with` statement instead?

e.g.

```python
for attempt in tenacity.Retrying(...):
    with attempt:
        ret = generic_utils.check_commands_on_units(cmds, units)
```

gustavosr98: Thanks

zaza/openstack/charm_tests/manila/tests.py
```diff
@@ -85,6 +86,29 @@ def test_manila_api(self):
...
+
+        cmds = []
+        for check_name in services:
+            cmds.append(
+                'egrep -oh /usr/local.* /etc/nagios/nrpe.d/'
+                'check_{}.cfg'.format(check_name)
+            )
+
+        for attempt in tenacity.Retrying(
+            retry=tenacity.retry_if_result(lambda ret: ret is not None),
```
AurelienLourot, line 102: I think this is wrong now that we have an `assertIsNone`. I.e. in case of failure, we won't have a None ret but a raised exception instead. I believe this should now be something like

```
        for attempt in tenacity.Retrying(
                wait=tenacity.wait_fixed(20),  # seconds
                stop=tenacity.stop_after_attempt(12),
                reraise=True):
            with attempt:
```

Note that I changed the `wait_fixed` as well, as waiting 2 minutes between retries was extremely long IMHO.

zaza/openstack/charm_tests/manila/tests.py
```diff
@@ -85,6 +86,29 @@ def test_manila_api(self):
...
+            wait=tenacity.wait_fixed(120),
+            stop=tenacity.stop_after_attempt(2)
+        ):
+            with attempt:
+                ret = generic_utils.check_commands_on_units(cmds, units)
+                if ret:
+                    logging.info(ret)
```
AurelienLourot, line 109: There is no need to log `ret` here as this will be logged already thanks to the `assertIsNone`

AurelienLourot (changes requested):
LGTM in general but see my inline comments, thanks!
To be re-valided after corrections within https://review.opendev.org/c/openstack/charm-manila/+/804932 - one bundle is enough for me, thanks!

AurelienLourot (approved):
LGTM, thanks! I'll land this together with https://review.opendev.org/c/openstack/charm-manila/+/804932

PR #625, opened 2021-08-19
Do not update external network data port if already set

coreycb:


coreycb:
I've tested this successfully:

From bundle:
   data-port: &data-port br-ex:enp3s0f0

2021-08-19 02:25:52 [INFO] Skip update of external network data port config.Config 'bridge-interface-mappings' already set to value: br-ex:enp3s0f0

$ juju config ovn-chassis bridge-interface-mappings                                                                                                                                                                                      
br-ex:enp3s0f0

fnordahl (approved):
LGTM, thanks!

PR #626, opened 2021-08-25
Implement the tests for the cinder-lvm charm

lmlg:
The cinder-lvm charm is a subordinate of the cinder charm. The spec
can be found here:
https://github.com/openstack/charm-specs/blob/master/specs/xena/backlog/cinder-lvm.rst

### DIFF
--- a/zaza/openstack/charm_tests/cinder_lvm/tests.py
+++ b/zaza/openstack/charm_tests/cinder_lvm/tests.py
@@ -25,6 +25,7 @@ import zaza.openstack.utilities.openstack as openstack_utils
 
 
 def with_conf(key, value):
+    """Temporarily set a configuration option for the cinder-lvm unit."""
     def patched(f):
         def inner(*args, **kwargs):
             prev = openstack_utils.get_application_config_option(
@@ -54,6 +55,7 @@ class CinderLVMTest(test_utils.OpenStackBaseTest):
 
     @classmethod
     def tearDown(cls):
+        """Remove tests resources."""
         volumes = cls.cinder_client.volumes
         for volume in volumes.list():
             if volume.name.startswith('zaza'):
@@ -64,6 +66,7 @@ class CinderLVMTest(test_utils.OpenStackBaseTest):
                     pass
 
     def test_cinder_config(self):
+        """Test that configuration options match our expectations."""
         logging.info('cinder-lvm')
         expected_contents = {
             'LVM-zaza-lvm': {
@@ -86,6 +89,7 @@ class CinderLVMTest(test_utils.OpenStackBaseTest):
             timeout=10)
 
     def _tst_create_volume(self):
+        """Create a volume via the LVM backend."""
         test_vol_name = "zaza{}".format(uuid.uuid1().fields[0])
         vol_new = self.cinder_client.volumes.create(
             name=test_vol_name,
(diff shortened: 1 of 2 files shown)


zaza/openstack/charm_tests/cinder_lvm/tests.py
```diff
@@ -0,0 +1,136 @@
...
+
+import logging
+import uuid
+
+import zaza.model
+import zaza.openstack.charm_tests.test_utils as test_utils
+import zaza.openstack.utilities.openstack as openstack_utils
+
+
+def with_conf(application, config, model_name=None):
```
freyes, line 27: you should use `BaseCharmTest.config_changed()` - https://github.com/freyes/zaza-openstack-tests/blob/master/zaza/openstack/charm_tests/test_utils.py#L264

zaza/openstack/charm_tests/cinder_lvm/tests.py
```diff
@@ -0,0 +1,136 @@
...
+                'volumes_dir': ['/var/lib/cinder/volumes'],
+                'volume_name_template': ['volume-%s'],
+                'volume_clear_size': ['0'],
+                'volume_driver': ['cinder.volume.drivers.lvm.LVMVolumeDriver'],
+            }}
+
+        zaza.model.run_on_leader(
+            'cinder',
+            'sudo cp /etc/cinder/cinder.conf /tmp/',
+            model_name=self.model_name)
```
freyes, line 87: there is no need to pass the model_name unless you are doing something special, zaza internally uses `zaza.model.run_in_model()`[0] which will rely on `get_juju_model()` just like you do in `setUpClass()`.

[0] https://github.com/openstack-charmers/zaza/blob/master/zaza/model.py#L186

zaza/openstack/charm_tests/cinder_lvm/tests.py
```diff
@@ -0,0 +1,136 @@
...
+    def test_device_none(self):
+        """Test creating a volume in a dummy device (set as 'none')."""
+        self._create_volume()
+
+    @with_conf('cinder-lvm', {'remove-missing': 'true'})
+    def test_remove_missing_volume(self):
+        """Test creating a volume after remove missing ones in a group."""
+        self._create_volume()
+
+    @with_conf('cinder-lvm', {'remove-missing-force': 'true'})
```
freyes, line 133: IIUC this under the hood should end up calling `vgreduce --removemissing ...`, I think you may need to setup a scenario where there are VGs referencing missing PVs so we are certain that under those circumstances it's possible to create volumes.

freyes (changes requested):

zaza/openstack/charm_tests/cinder_lvm/tests.py
lmlg: @freyes as discussed a couple days ago, cinder makes it hard to actually test this in a clean manner, since only a single volume name can be configured as opposed to a comma-separated list. We could, of course, manually create them in the tests, but it will end up polluting the test environment, since all the LVM stuff are global settings and can't be isolated.

wolsen: I think this is a good test to have. Yes, it will manipulate the test environment - but that can be done as the last test to run so it doesn't inadvertently pollute other tests that are exercised. However, I think I'm comfortable with this being a subsequent improvement to the cinder-lvm tests as long as @lmlg is willing to commit to do those in a follow-on patch.

lmlg: I can absolutely commit them in a follow up PR.

PR #628, opened 2021-08-31
Add certificate check for Ceph dashboard

gnuoy:
Add setup step for the dashboard which will block until the
certificates are present and the model is idle. This is to prevent
the tests from continuing when the certificates are not ready.
Also up CephDashboardTest to use standard tools for collecting
ca cert.


gnuoy:
This branch is being used here: https://review.opendev.org/c/openstack/charm-ceph-dashboard/+/806792

wolsen (approved):

PR #646, opened 2021-10-04
Add retries when accessing Ceph dashboard

gnuoy:


ajkavanagh (approved):
Looks good; I'm resisting the urge to ask for refactoring to reduce the duplication as it currently is easy to understand and see what is going on.  Sometimes my urge for DRY and functional programming approaches obscures code rather than revealing it.

PR #650, opened 2021-10-07
Add a functional test to check ceph-mon descaling

lmlg:
This PR adds a new functional test for the ceph-mon charm to check that descaling (that is, removal of a unit) works correctly by maintaining consistency with the ceph cluster. In other words, when removing a ceph-mon unit (via `juju remove-unit`), the charm must also be deregistered from the set of ceph monitors.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 # This is necessary for Xenial builders
 # BUG: https://github.com/openstack-charmers/zaza-openstack-tests/issues/530
 lxml<4.6.3
-pyparsing<3.0.0
+pyparsing<3.0.0  # pin for aodhclient which is held for py35
 aiounittest
 async_generator
 boto3
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ install_require = [
     'PyYAML',
     'tenacity',
     'oslo.config<6.12.0',
+    'pyparsing<3.0.0',  # pin for aodhclient which is held for py35
     'aodhclient<1.4.0',
     'gnocchiclient>=7.0.5,<8.0.0',
     'pika>=1.1.0,<2.0.0',
--- a/unit_tests/utilities/test_zaza_utilities_openstack.py
+++ b/unit_tests/utilities/test_zaza_utilities_openstack.py
@@ -58,11 +58,12 @@ class TestOpenStackUtils(ut_utils.BaseTestCase):
 
         self.network = {
             "network": {"id": "network_id",
-                              "name": self.ext_net,
-                              "tenant_id": self.project_id,
-                              "router:external": True,
-                              "provider:physical_network": "physnet1",
-                              "provider:network_type": "flat"}}
+                        "name": self.ext_net,
+                        "router:external": True,
+                        "shared": False,
+                        "tenant_id": self.project_id,
+                        "provider:physical_network": "physnet1",
+                        "provider:network_type": "flat"}}
 
         self.networks = {
             "networks": [self.network["network"]]}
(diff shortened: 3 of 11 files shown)


zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -1014,3 +1015,45 @@ def test_invalid_compression_configuration(self):
...
+        logging.info('Monitor count is {}'.format(mon_count))
+        # We add a new unit so that the ceph-mon count remains
+        # above the 'monitor-count' specified in the bundle.
+        zaza_model.add_unit('ceph-mon')
+        logging.info('Waiting for new ceph-mon unit to settle')
+        zaza_model.block_until_unit_count('ceph-mon', mon_count + 1)
+        zaza_model.block_until_all_units_idle()
+
+        # Now remove a ceph-mon unit and test for cluster consistency.
+        zaza_model.destroy_unit('ceph-mon', self.get_mon_unit())
```
ChrisMacNaughton, line 1055: Given that the Juju leader for a deployed ceph-mon doesn't really relate to the Ceph cluster, I'd be tempted to suggest grabbing the first mon, rather than filtering by leadership

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -1032,15 +1032,22 @@ def get_mon_count(self, unit):
         match = rx.search(result['Stdout'])
         return int(match.group(1))
 
+    def get_mon_units(self):
+        """Get a list of the ceph-mon units."""
+        return list(x.entity_id for x in zaza_model.get_units('ceph-mon'))
```
ajkavanagh, line 1037: It would be slightly more efficient to use a list comprehension rather than a generator and then consume that to a list, but I realise it was already like this previously.

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -1032,15 +1032,22 @@ def get_mon_count(self, unit):
...
+        zaza_model.add_unit('ceph-mon', wait_appear=True)
+        try:
+            logging.info('New monitor has been added')
+            new_mon = next(iter(set(self.get_mon_units()) - set(mons)))
+            zaza_model.wait_for_unit_idle(new_mon)
+            self.assertEqual(mon_count + 1, self.get_mon_count(mons[0]))
+            zaza_model.destroy_unit('ceph-mon', mons[0], wait_disappear=True)
+            self.assertEqual(mon_count, self.get_mon_count(mons[1]))
+        except Exception:
+            zaza_model.destroy_unit('ceph-mon', mons[0])
```
ajkavanagh, line 1053: It seems a bit odd to wrap this in an Exception that catches everything?  I don't understandas this is supposed to be a test?  Shouldn't it fail when something goes wrong?

lmlg: Definitely. The idea of that clause is that the unit that was added is still removed regardless. This is done so that any subsequent tests run with the expected number of ceph-mon units. It may seem a bit wasteful since this is in case of failure anyway, but I think it's useful so that other tests run in a non-polluted environment.

ajkavanagh: I don't think I explained myself properly.  The test can't fail as any exceptions are caught, unless destroy_unit() fails.  I think you have missed of a "raise" statement after the destroy_unit() call to re-raise the exception and fail the test.

lmlg: ahhh, yes, you are right. I'll add the code to reraise the exception.

ajkavanagh:
Unfortunately, it looks like pyparsing needs to be pinned for this repo too (for the moment!) ... see the failing tests.

lmlg:
> Unfortunately, it looks like pyparsing needs to be pinned for this repo too (for the moment!) ... see the failing tests.

yep, just saw it.

PR #653, opened 2021-10-14
Favor the V3 volume endpoint for tempest config

coreycb:
If the V3 volume endpoint is available, use it for tempest testing.
The V2 volume endpoint is removed in OpenStack Xena, so this allows
us to use the latest available endpoint.


ChrisMacNaughton (approved):
LGTM, thanks!

PR #654, opened 2021-10-18
Extend OPENSTACK_CODENAMES up to Yoga

lourot:
This dict is used in the OpenStack upgrade tests in
order to automatically determine the next release and
set the openstack-origin accordingly.

Also taking the opportunity to extend neighboring
lists.


ajkavanagh (approved):
LGTM

PR #655, opened 2021-10-20
Support OVN charms in automated OpenStack upgrade

AurelienLourot:
Validated by running it from https://github.com/openstack-charmers/charmed-openstack-tester

### DIFF
--- a/zaza/openstack/utilities/upgrade_utils.py
+++ b/zaza/openstack/utilities/upgrade_utils.py
@@ -41,8 +41,8 @@ SERVICE_GROUPS = (
         'aodh', 'barbican', 'ceilometer', 'ceph-fs',
         'ceph-radosgw', 'cinder', 'designate',
         'designate-bind', 'glance', 'gnocchi', 'heat', 'manila',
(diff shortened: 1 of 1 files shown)


zaza/openstack/utilities/upgrade_utils.py
```diff
@@ -41,8 +41,8 @@
         'aodh', 'barbican', 'ceilometer', 'ceph-fs',
         'ceph-radosgw', 'cinder', 'designate',
         'designate-bind', 'glance', 'gnocchi', 'heat', 'manila',
-        'manila-generic', 'neutron-api', 'neutron-gateway', 'placement',
-        'nova-cloud-controller', 'openstack-dashboard']),
+        'manila-generic', 'neutron-api', 'neutron-gateway', 'ovn-central',
```
fnordahl, line 44: When does the `ovn-chassis` or `ovn-dedicated-chassis` get upgraded?

AurelienLourot: charms not in this list of principle charms get upgraded at the very end of the process. You're right I'll add the dedicated chassis then, thanks!

fnordahl (approved):

PR #656, opened 2021-10-22
Swift proxy remove devices test added

Barteus:
A functional test for multi-zone deployment was added.

The functional test removes the devices in region1 zone3. This test requires a single region with "redundant zone 3", so it can be safely removed.
To check if cluster works after device removal - object upload and retrieval of header data are used.

### DIFF
--- a/zaza/openstack/charm_tests/swift/tests.py
+++ b/zaza/openstack/charm_tests/swift/tests.py
@@ -95,6 +95,54 @@ class SwiftProxyTests(test_utils.OpenStackBaseTest):
 class SwiftProxyMultiZoneTests(test_utils.OpenStackBaseTest):
     """Tests specific to swift proxy in multi zone environment."""
 
+    RESOURCE_PREFIX = 'zaza-swift-proxy-multizone-tests'
+
+    @classmethod
+    def setUpClass(cls):
+        """Run class setup for running tests."""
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/swift/tests.py
```diff
@@ -92,6 +92,84 @@ def test_903_disk_usage_action(self):
...
+        logging.info(action)
+        self.assertEqual(action.status, "completed")
+
+        container_name, obj_name, _ = swift_utils.create_object(
+            self.swift_region1,
+            self.region1_proxy_app,
+            self.storage_topology,
+            self.RESOURCE_PREFIX,
+            model_name=self.region1_model_name)
+        # Check object is accessible from other regions proxy.
```
AurelienLourot, line 168: Noob question: what do you mean by "other regions"? It seems to be always the same region (RegionOne) throughout the test.

AurelienLourot (changes requested):
LGTM, thanks! Just one question.

Also for future travelers, this is being used in https://review.opendev.org/c/openstack/charm-swift-proxy/+/814415

AurelienLourot (approved):
LGTM, thanks! As this is a new test class this is safe to land as is without waiting for the charm review, as soon as the unit tests here turn green. If it turns out there are any mistakes, a follow-up func-test-pr can be done.

PR #658, opened 2021-10-22
Adjust cindet-netapp tests

lmlg:
Much like what happened with cinder-lvm, the host attribute for volumes is now mangled differently, which means we need to make a small adjustment to the check.

### DIFF
--- a/zaza/openstack/charm_tests/cinder_netapp/tests.py
+++ b/zaza/openstack/charm_tests/cinder_netapp/tests.py
@@ -67,18 +67,20 @@ class CinderNetAppTest(test_utils.OpenStackBaseTest):
 
     def _create_volumes(self, n):
         """Create N volumes."""
+        volumes = []
         for _ in range(n):
             name = "zaza{}".format(uuid.uuid1().fields[0])
             vol = self.cinder_client.volumes.create(
                 name=name,
                 size='1')
             vol.reset_state('available')
-            self.zaza_volumes.append(vol)
+            volumes.append(vol)
+        self.zaza_volumes.extend(volumes)
+        return volumes
 
     def test_create_volume(self):
         """Test creating volumes with basic configuration."""
-        self._create_volumes(2)
-        for vol in self.cinder_client.volumes.list():
+        for vol in self._create_volumes(2):
             self.assertTrue(vol)
             openstack_utils.resource_reaches_status(
                 self.cinder_client.volumes,
--- a/zaza/openstack/charm_tests/cinder_netapp/tests.py
+++ b/zaza/openstack/charm_tests/cinder_netapp/tests.py
@@ -22,79 +22,13 @@ from zaza.openstack.charm_tests.cinder_backend.tests import CinderBackendTest
 class CinderNetAppTest(CinderBackendTest):
     """Encapsulate netapp tests."""
 
-    @classmethod
-    def setUpClass(cls):
-        """Run class setup for running tests."""
-        super(CinderNetAppTest, cls).setUpClass()
-        cls.keystone_session = openstack_utils.get_overcloud_keystone_session()
-        cls.model_name = zaza.model.get_juju_model()
-        cls.cinder_client = openstack_utils.get_cinder_session_client(
-            cls.keystone_session)
-        cls.zaza_volumes = []
+    backend_name = 'cinder-netapp'
 
-    @classmethod
-    def tearDown(cls):
-        """Remove test resources."""
-        for volume in cls.zaza_volumes:
-            try:
-                volume.detach()
-                volume.force_delete()
-            except Exception:
-                pass
-
-    def test_cinder_config(self):
-        """Test that configuration options match our expectations."""
-        expected_contents = {
-            'cinder-netapp': {
-                'netapp_storage_family': ['ontap_cluster'],
-                'netapp_storage_protocol': ['iscsi'],
-                'volume_driver':
-                    ['cinder.volume.drivers.netapp.common.NetAppDriver'],
-            }}
-
-        zaza.model.run_on_leader(
-            'cinder',
-            'sudo cp /etc/cinder/cinder.conf /tmp/')
-        zaza.model.block_until_oslo_config_entries_match(
-            'cinder',
-            '/tmp/cinder.conf',
-            expected_contents,
-            timeout=2)
-
-    def _create_volumes(self, n):
-        """Create N volumes."""
-        volumes = []
-        for _ in range(n):
-            name = "zaza{}".format(uuid.uuid1().fields[0])
-            vol = self.cinder_client.volumes.create(
-                name=name,
-                size='1')
-            vol.reset_state('available')
-            volumes.append(vol)
-        self.zaza_volumes.extend(volumes)
-        return volumes
-
-    def _volume_host(self, vol):
-        ret = getattr(vol, 'os-vol-host-attr:host', None)
-        if ret is not None:
-            return ret
-        # A bit more roundabout, but also works when the attribute
-        # hasn't been set yet, for whatever reason.
-        for service in vol.manager.api.services.list():
-            if 'cinder-volume' in service.binary:
-                return service.host
-
-    def test_create_volume(self):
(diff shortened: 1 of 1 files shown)


AurelienLourot (approved):

AurelienLourot:
Thanks, and I think @ajkavanagh 's comment has been addressed. To be landed together with https://review.opendev.org/c/openstack/charm-cinder-netapp/+/815130

zaza/openstack/charm_tests/cinder_netapp/tests.py
```diff
@@ -55,23 +65,28 @@ def test_cinder_config(self):
...
-        """Test creating a volume with basic configuration."""
-        test_vol_name = "zaza{}".format(uuid.uuid1().fields[0])
-        vol_new = self.cinder_client.volumes.create(
-            name=test_vol_name,
-            size='1')
-        try:
+        """Test creating volumes with basic configuration."""
+        self._create_volumes(2)
+        for vol in self.cinder_client.volumes.list():
+            self.assertTrue(vol)
```
ajkavanagh, line 82: It's not `True`; it's just not `None`; shouldn't this assert be for what it actually is, as `True` can't have an `.id` later in the test (on line 85).

zaza/openstack/charm_tests/cinder_netapp/tests.py
```diff
@@ -55,23 +65,28 @@ def test_cinder_config(self):
...
     def test_create_volume(self):
-        """Test creating a volume with basic configuration."""
-        test_vol_name = "zaza{}".format(uuid.uuid1().fields[0])
-        vol_new = self.cinder_client.volumes.create(
-            name=test_vol_name,
-            size='1')
-        try:
+        """Test creating volumes with basic configuration."""
+        self._create_volumes(2)
+        for vol in self.cinder_client.volumes.list():
```
ajkavanagh, line 81: Shouldn't there be a check that the volume is one of the `self.zaza_volumes`?  Otherwise, this test may be polluted by other volumes that the cinder client can see (say due to other tests run, or previous tests not cleaning up properly)?

zaza/openstack/charm_tests/cinder_netapp/tests.py
lmlg: From what I understand, `assertTrue` is basically `bool(x)`, which helps if the volume class implements the `__bool__` protocol (don't know if it does, just wanted to be cautious).

zaza/openstack/charm_tests/cinder_netapp/tests.py
lmlg: Good point. I'll go with `self.zaza_volumes` since it covers that specific case (I actually added it for that purpose, just forgot to use it here).

zaza/openstack/charm_tests/cinder_netapp/tests.py
ajkavanagh: It's not that it *won't* work; I'm absolutely certain it will.  It's that it *says* "assert that vol is True", whereas I think you are actually asserting that `vol` is not None, which is the only alternative to vol not being an object that you can later get an attribute on (`is`).  My issue is that I think code should say what it means: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsNotNone

lmlg: Sounds reasonable. I'll change it.

zaza/openstack/charm_tests/cinder_netapp/tests.py
```diff
@@ -55,23 +65,30 @@ def test_cinder_config(self):
...
                 expected_status='available',
                 msg='Volume status wait')
-            test_vol = self.cinder_client.volumes.find(name=test_vol_name)
-            self.assertEqual(
-                getattr(test_vol, 'os-vol-host-attr:host').split('#')[0],
-                'cinder@cinder-netapp')
-        finally:
-            self.cinder_client.volumes.delete(vol_new)
+            test_vol = self.cinder_client.volumes.find(name=vol.name)
+            host = getattr(test_vol, 'os-vol-host-attr:host').split('#')[0]
```
AurelienLourot, line 93: This seems to fail with `AttributeError: 'NoneType' object has no attribute 'split'`
See https://review.opendev.org/c/openstack/charm-cinder-netapp/+/815130

AurelienLourot (changes requested):

zaza/openstack/charm_tests/cinder_netapp/tests.py
coreycb: Maybe we can add a debug to that line for now and re-run the gate?

ajkavanagh: > Maybe we can add a debug to that line for now and re-run the gate?

It sounds like the attr doesn't exist.  Should it exist on the version it is failing on?  Is it a race? Is there something else wrong?

AurelienLourot: Fine with me with adding more debugging output here, if hard to reproduce

lmlg: Running with the default configuration, I'm getting:
```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/cinder/volume/drivers/netapp/dataontap/client/api.py", line 199, in send_http_request
    response = self._opener.open(request)
  File "/usr/lib/python3.8/urllib/request.py", line 525, in open
    response = self._open(req, data)
  File "/usr/lib/python3.8/urllib/request.py", line 542, in _open
    result = self._call_chain(self.handle_open, protocol, protocol +
  File "/usr/lib/python3.8/urllib/request.py", line 502, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.8/urllib/request.py", line 1383, in http_open
    return self.do_open(http.client.HTTPConnection, req)
  File "/usr/lib/python3.8/urllib/request.py", line 1357, in do_open
    raise URLError(err)
urllib.error.URLError: <urlopen error [Errno 113] EHOSTUNREACH>
```

I'm investigating what's going on with the netapp appliance.

AurelienLourot (approved):
LGTM, thanks, waiting for green results on https://review.opendev.org/c/openstack/charm-cinder-netapp/+/815130

zaza/openstack/charm_tests/cinder_netapp/tests.py
```diff
@@ -55,23 +65,40 @@ def test_cinder_config(self):
...
                 msg='Volume status wait')
-            test_vol = self.cinder_client.volumes.find(name=test_vol_name)
-            self.assertEqual(
-                getattr(test_vol, 'os-vol-host-attr:host').split('#')[0],
-                'cinder@cinder-netapp')
-        finally:
-            self.cinder_client.volumes.delete(vol_new)
+            test_vol = self.cinder_client.volumes.find(name=vol.name)
+            host = self._volume_host(test_vol)
+            self.assertIn('cinder-netapp', host)
```
AurelienLourot, line 104: Failed with `AssertionError: 'cinder-netapp' not found in 'juju-57c895-zaza-b42470362b3b-2@LVM'` in https://review.opendev.org/c/openstack/charm-cinder-netapp/+/815130

AurelienLourot (changes requested):

AurelienLourot:
Blocked by https://review.opendev.org/c/openstack/cinder/+/815560

zaza/openstack/charm_tests/cinder_netapp/tests.py
coreycb: > Blocked by https://review.opendev.org/c/openstack/cinder/+/815560

This patch has landed in jammy/yoga for ubuntu as well as proposed pockets for impish/xena and wallaby.

gnuoy:
Apologies, I missed this PR and I think it has been superseded by df91c1e6cb76

coreycb:
This appears to have been fixed by https://github.com/openstack-charmers/zaza-openstack-tests/commit/ed0bca46fff0ad3f750099f2e110ea90221b2356

PR #662, opened 2021-10-27
Add cinder encryption test

gnuoy:


nobuto-m:
I'm not 100% sure, but this may need to be rebased onto master to include https://github.com/openstack-charmers/zaza-openstack-tests/pull/670

zaza/openstack/charm_tests/cinder/tests.py
```diff
@@ -30,15 +33,15 @@
...
+class CinderBase(test_utils.OpenStackBaseTest):
+    """Base for Cinder tests."""
 
     RESOURCE_PREFIX = 'zaza-cindertests'
 
     @classmethod
     def setUpClass(cls):
         """Run class setup for running tests."""
-        super(CinderTests, cls).setUpClass(application_name='cinder')
+        super(CinderBase, cls).setUpClass(application_name='cinder')
```
ajkavanagh, line 44: Just out of interest, as we are Py3 only in this library we can use the shortcut version of this:

```python
super().setUpClass(application_name='cinder')
```

PR #667, opened 2021-11-10
Add wait for sync to ceph RGW tests

gnuoy:


freyes:
I added the Func-Test-Pr header to this patch https://review.opendev.org/c/openstack/charm-ceph-radosgw/+/817396 so we can exercise it in the CI.

PR #670, opened 2021-11-22
Authenticate with cinder v3 endpoint by default

coreycb:
The v3 endpoint has been supported by the charms since Pike,
and the V2 endpoint was removed in Xena, therefore it makes
to use the v3 endpoint by default at this point.


AurelienLourot (approved):
Makes sense, thanks! Waiting for https://review.opendev.org/c/openstack/charm-cinder/+/814177 to turn green

freyes (approved):
LGTM and https://review.opendev.org/c/openstack/charm-cinder/+/814177  completed successfully :+1:

PR #672, opened 2021-11-22
Authenticate with cinder v3 endpoint by default

coreycb:
The v3 endpoint has been supported by the charms since Pike,
and the V2 endpoint was removed in Xena, therefore it makes
to use the v3 endpoint by default at this point.

(cherry picked from commit 1eaa2233c11f04aebd1535f1a58ceca3e5cf6bab)

### DIFF
--- a/zaza/openstack/utilities/openstack.py
+++ b/zaza/openstack/utilities/openstack.py
@@ -407,7 +407,7 @@ def get_heat_session_client(session, version=1):
     return heatclient.Client(session=session, version=version)
 
 
-def get_cinder_session_client(session, version=2):
+def get_cinder_session_client(session, version=3):
     """Return cinderclient authenticated by keystone session.
 
     :param session: Keystone session object


ajkavanagh:
Good ole pyparsing issue with py36:

```
   File "/home/runner/work/zaza-openstack-tests/zaza-openstack-tests/.tox/py3.6/lib/python3.6/site-packages/aodhclient/utils.py", line 43, in <module>
    expr = pp.operatorPrecedence(condition, [
AttributeError: module 'pyparsing' has no attribute 'operatorPrecedence'
```

coreycb:
I've created a pyparsing cherry-pick at https://github.com/openstack-charmers/zaza-openstack-tests/pull/673

ajkavanagh:
Doesn't changing the default here for stable/21.10 branch mean that we have to verify that branch against all of the tests/charms that might use the client code?

ajkavanagh (changes requested):
Just holding; this is a change in default against the stable/21.10 branch which we did all the verification for.

coreycb:
func-test-pr doesn't work for stable branches so I think we would have to land this first and then run tests for the charm change

ajkavanagh:
Sorry to ask a daft question, @coreycb but do we *have* to backport this in zaza-openstack-tests for the stable/21.01 branch?  i.e. were our tests invalid for stable/21.10, or has something been backported that now invalidates those tests?

coreycb:
In my opinion, yes we have to. The 21.10 release of the cinder charm supports xena, so we need to fix this in the stable cinder branch -> https://review.opendev.org/c/openstack/charm-cinder/+/818739, and it depends on this z-o-t fix to successfully run its gate tests.

ajkavanagh (approved):
Based on our discussion re: low regression issues, I'm very happy to approve.

PR #677, opened 2021-11-29
Add test to masakari to verify transport_url param

drencrom:
The transport_url param should show all rabbitmq servers available. I
configured 3 servers in the test bundles of the masakari charm and check
here that all of them are present.

### DIFF
--- a/zaza/openstack/charm_tests/masakari/tests.py
+++ b/zaza/openstack/charm_tests/masakari/tests.py
@@ -228,3 +228,11 @@ class MasakariTest(test_utils.OpenStackBaseTest):
         logging.info('{} pid is now {}'.format(vm_name, new_guest_pid))
         assert new_guest_pid and new_guest_pid != guest_pid, (
             "Restart failed or never happened")
+
+    def test_transport_url(self):
+        zaza.model.block_until_file_matches_re(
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/masakari/tests.py
```diff
@@ -228,3 +228,11 @@ def test_instance_restart_on_fail(self):
         logging.info('{} pid is now {}'.format(vm_name, new_guest_pid))
         assert new_guest_pid and new_guest_pid != guest_pid, (
             "Restart failed or never happened")
+
+    def test_transport_url(self):
+        zaza.model.block_until_file_matches_re(
+            'masakari',
+            '/etc/masakari/masakari.conf',
+            'transport_url = rabbit://(masakari:\w+@[\d.]+:\d+,?){3}/masakari',
```
freyes, line 236: As I mentioned in the proposed patch for rabbitmq-server charm, I think this issue could be validated with a unit test only since the functional test checks the rendered file contains the 3 servers. That approach would also help to not increase the footprint of CI jobs.

freyes (changes requested):

drencrom:
Removed functional testing for this patch

PR #678, opened 2021-11-29
Continue to use cinder v2 endpoint before Pike

coreycb:


zaza/openstack/utilities/openstack.py
```diff
@@ -417,6 +417,8 @@ def get_cinder_session_client(session, version=3):
     :returns: Authenticated cinderclient
     :rtype: cinderclient.Client object
     """
+    if (get_os_release() < get_os_release('xenial_pike')):
+        version = 2
```
ajkavanagh, line 421: Technically, I think this works, but it does mean that you can't ask for a client 3 in xenial.

How about:

```python
    def get_cinder_session_client(session, version=None):
        if version is None:
            version = 2 if (get_os_release() < get_os_release('xenial_pike')) else 3
        ...
```

This provides the correct default if no version is passed, otherwise the version is selected.
```

coreycb: That's a lot better, thanks for the suggestion. It's handled in the new commit.

ajkavanagh (approved):
LGTM

PR #679, opened 2021-11-30
Ensure that series upgrades ignore percona at focal

ajkavanagh:
When doing a series upgrade from bionic to focal, the percona-cluster
charm needs to be avoided.  This filters the applications for percona if
the target is focal.  Note if percona is placed on the same unit as
something else that needs to be upgraded (bad idea) then it may still
get 'upgraded' and the operation will fail.

### DIFF
--- a/zaza/openstack/charm_tests/rabbitmq_server/tests.py
+++ b/zaza/openstack/charm_tests/rabbitmq_server/tests.py
@@ -26,8 +26,8 @@ import zaza.model
 import zaza.openstack.charm_tests.test_utils as test_utils
 import zaza.openstack.utilities.generic as generic_utils
 
-from charmhelpers.core.host import CompareHostReleases
 from zaza.openstack.utilities.generic import get_series
+from zaza.openstack.utilities.os_versions import CompareHostReleases
 
 from . import utils as rmq_utils
 from .utils import RmqNoMessageException
--- a/zaza/openstack/utilities/os_versions.py
+++ b/zaza/openstack/utilities/os_versions.py
@@ -278,3 +278,105 @@ PACKAGE_CODENAMES = {
         ('4', 'victoria'),
     ]),
 }
+
+
+UBUNTU_RELEASES = (
+    'lucid',
+    'maverick',
+    'natty',
+    'oneiric',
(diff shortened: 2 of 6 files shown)


zaza/openstack/utilities/os_versions.py
```diff
@@ -278,3 +278,105 @@
...
+        s_newton = CompareOpenstack('newton')
+
+        assert s_newton > s_mitaka
+
+        :returns: <string>
+        """
+        return self._list[self.index]
+
+
+class CompareHostReleases(BasicStringComparator):
```
AurelienLourot, line 373: What is the difference with `from charmhelpers.core.host.CompareHostReleases`?

zaza/openstack/charm_tests/rabbitmq_server/tests.py
```diff
@@ -26,8 +26,8 @@
 import zaza.openstack.charm_tests.test_utils as test_utils
 import zaza.openstack.utilities.generic as generic_utils
 
-from charmhelpers.core.host import CompareHostReleases
```
AurelienLourot, line 29: I'm surprised, see my other question, why did we copy this code from charm-helpers in here?

AurelienLourot (changes requested):
LGTM, thanks, just one question about code duplication

zaza/openstack/charm_tests/rabbitmq_server/tests.py
ajkavanagh: Oh that's deliberate, but perhaps a tad conscientious.  I don't think that zaza-openstack-tests should rely on charmhelpers as a library.  charmhelpers is really meant for running on the unit, and this is a test framework.  CompareHostReleases is a relatively small piece of code and the list of releases (as a concept) is already in os_versions.py in various forms.

Therefore, it actually feels better to drop charmhelpers as a dependency for z-o-t just for this one function (which is useful) and just copy the code in.

AurelienLourot: Makes sense, thanks!

AurelienLourot (approved):

PR #682, opened 2021-12-02
Stop retrying if LB provisioning status reached ERROR

freyes:
The method `wait_for_lb_resource()` retries for 15m while the load
balancer could have reached to ERROR during the provisioning in the
first few minutes, this approach makes the testing take longer for no
reason.

This change makes the ERROR state in provisioning_status final and
abort raising a ValueError() exception.

More details of the provisioning_status possible states can be found at:
https://docs.openstack.org/api-ref/load-balancer/v2/#provisioning-status-codes

### DIFF
--- a/zaza/openstack/charm_tests/octavia/tests.py
+++ b/zaza/openstack/charm_tests/octavia/tests.py
@@ -26,6 +26,10 @@ import zaza.openstack.charm_tests.test_utils as test_utils
 import zaza.openstack.utilities.openstack as openstack_utils
 
 from zaza.openstack.utilities import ObjectRetrierWraps
+from zaza.openstack.utilities.exception import (
+    LoadBalancerUnexpectedState,
+    LoadBalancerUnrecoverableError,
+)
 
 LBAAS_ADMIN_ROLE = 'load-balancer_admin'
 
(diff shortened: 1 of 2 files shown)


freyes:
This patch https://review.opendev.org/c/openstack/charm-octavia/+/819438 is running with this change.

zaza/openstack/charm_tests/octavia/tests.py
```diff
@@ -236,13 +236,21 @@ def wait_for_lb_resource(octavia_show_func, resource_id,
...
+        logging.info("Current provisioning status: {}, waiting for {}"
+                     .format(resp['provisioning_status'], provisioning_status))
+
+        msg = ('load balancer resource has not reached '
+               'expected provisioning status: {}'.format(resp))
+
+        # ERROR is a final state, once it's reached there is no reason to keep
+        # retrying and delaying the failure.
+        if resp['provisioning_status'] == 'ERROR':
+            raise ValueError(msg)
```
wolsen, line 248: I think raising an AssertionError would probably be a bit better than ValueError. It's effectively the same end result, but when reading the trace backs it would show that the assertion failed which is probably more practical for debugging from logs in test runs.

ajkavanagh: I wondered about that too, @wolson, but oddly (?) the code uses tenacity and the `AssertionError` triggers a retry.  So `ValueError` is one way to abort and escape.

I think a proper fix would be to use a custom `Exception` for the 'normal' retries (i.e.  not re-purpose `AssertionError`) and then use `AssertionError` here to allow the escape.  e.g. rewrite the tenacity block with a Custom Exception, and retrieve asserts to be asserts rather that causing a retry.

freyes: yes, I didn't go for `AssertionError` because it doesn't break the retries. I will add the custom Exception as @ajkavanagh suggests.

ajkavanagh (approved):
LGTM.  Thanks for the changes.  Sorry for the ridiculously long time to get it merged.

PR #683, opened 2021-12-09
Implement test for disk replacement

lmlg:
This PR builds on the add-disk one, testing the complete flow of adding a disk, removing it and finally re-adding it after recycling the original OSD id.


ChrisMacNaughton (approved):
LGTM, holding off landing until the [functionality[(https://review.opendev.org/c/openstack/charm-ceph-osd/+/820191) lands

PR #684, opened 2021-12-14
Search for message in rabbit tests

gnuoy:
The current rabbit tests post a message on one unit and then
consumes the first message on another unit. If the two do not match
then the test fails. This means that a single pre-existing message
can break all these tests as the consumer always gets the wrong
message. This change updates the tests to search for the target
message rather than assuming that the first message is the target
message.

Messages that are not the target message are not reposted so any
pre-existing messages are consumed and in effect thrown away but
given these are all test messages in a test queue and the tests do
not run in parallel this should be fine.

### DIFF
--- a/zaza/openstack/charm_tests/rabbitmq_server/tests.py
+++ b/zaza/openstack/charm_tests/rabbitmq_server/tests.py
@@ -57,6 +57,41 @@ class RmqTests(test_utils.OpenStackBaseTest):
                                                   ssl=ssl,
                                                   port=port)
 
+    @tenacity.retry(reraise=True, stop=tenacity.stop_after_attempt(30),
+                    retry=tenacity.retry_if_exception_type(AssertionError))
+    def _search_for_message(self, amqp_msg, check_unit, ssl, port,
+                            amqp_msg_counter):
+        """Search for message in message queue.
+
+        WARNING: This will consume messages until it finds the target message.
+
+        :param amqp_msg: Message to search for
+        :type amqp_msg: string
+        :param check_unit: Unit to retrieve messages from
+        :type check_unit: juju.unit.Unit
+        :param ssl: Whether to use SSL when connecting to rabbit
+        :type ssl: bool
+        :param port: Port to use when connecting to rabbit
+        :type port: Union[int, None]
+        :param amqp_msg_counter: Number in test sequence of this message.
+        :type amqp_msg: int
+        :raises: tenacity.RetryError
+        """
+        amqp_msg_rcvd = self._retry_get_amqp_message(
+            check_unit,
+            ssl=ssl,
+            port=port)
+
+        try:
+            logging.info('Looking for message {}'.format(amqp_msg))
+            # Validate amqp message content
+            assert amqp_msg == amqp_msg_rcvd
+            logging.info('Message {} received OK.'.format(amqp_msg_counter))
+        except AssertionError as err:
+            logging.info('Expected: {}'.format(amqp_msg))
+            logging.info('Actual:   {}'.format(amqp_msg_rcvd))
+            raise err
+
     def _test_rmq_amqp_messages_all_units(self, units,
                                           ssl=False, port=None):
         """Reusable test to send/check amqp messages to every listed rmq unit.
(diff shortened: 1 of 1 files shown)


gnuoy:
Tested locally: https://paste.ubuntu.com/p/dkpCSQk6qS/

zaza/openstack/charm_tests/rabbitmq_server/tests.py
```diff
@@ -57,6 +57,41 @@ def _retry_get_amqp_message(self, check_unit, ssl=None, port=None):
...
+        :type amqp_msg: string
+        :param check_unit: Unit to retrieve messages from
+        :type check_unit: juju.unit.Unit
+        :param ssl: Whether to use SSL when connecting to rabbit
+        :type ssl: bool
+        :param port: Port to use when connecting to rabbit
+        :type port: Union[int, None]
+        :param amqp_msg_counter: Number in test sequence of this message.
+        :type amqp_msg: int
+        :raises: tenacity.RetryError
```
AurelienLourot, line 78: I don't think this is the right exception type because you are using `reraise=True`, which means we won't raise a tenacity exception but reraise the original exception, if I'm not mistaken?

zaza/openstack/charm_tests/rabbitmq_server/tests.py
```diff
@@ -111,25 +146,21 @@ def _test_rmq_amqp_messages_all_units(self, units,
...
-                    logging.error('Actual:   {}'.format(amqp_msg_rcvd))
-                    msg = 'Message {} mismatch.'.format(amqp_msg_counter)
+                try:
+                    self._search_for_message(
+                        amqp_msg,
+                        check_unit,
+                        ssl,
+                        port,
+                        amqp_msg_counter)
+                except tenacity.RetryError:
```
AurelienLourot, line 161: See my other comments: I would expect you'd have to catch an AssertionError instead?

AurelienLourot (changes requested):
LGTM in general, thanks. See my inline comments

zaza/openstack/charm_tests/rabbitmq_server/tests.py
gnuoy: Yeah, this is right in the sense that tenacity.RetryError is whats raised because _retry_get_amqp_message does not have reraise=True. But I think I should fix that so brace yourself for a follow up PR!

zaza/openstack/charm_tests/rabbitmq_server/tests.py
```diff
@@ -57,6 +58,41 @@ def _retry_get_amqp_message(self, check_unit, ssl=None, port=None):
...
+        :type amqp_msg: string
+        :param check_unit: Unit to retrieve messages from
+        :type check_unit: juju.unit.Unit
+        :param ssl: Whether to use SSL when connecting to rabbit
+        :type ssl: bool
+        :param port: Port to use when connecting to rabbit
+        :type port: Union[int, None]
+        :param amqp_msg_counter: Number in test sequence of this message.
+        :type amqp_msg: int
+        :raises: RmqNoMessageException
```
ajkavanagh, line 79: If I was being very picky, I'd say that it also raises AssertionError if it runs out of retries?

zaza/openstack/charm_tests/rabbitmq_server/tests.py
```diff
@@ -57,6 +58,41 @@ def _retry_get_amqp_message(self, check_unit, ssl=None, port=None):
                                                   ssl=ssl,
                                                   port=port)
 
+    @tenacity.retry(reraise=True, stop=tenacity.stop_after_attempt(30),
+                    retry=tenacity.retry_if_exception_type(AssertionError))
```
ajkavanagh, line 62: It seems a little odd to me to use tenacity to implement a search.  Wouldn't a while loop just be easier until it runs out of messages on the queue?  The delay is implemented in `_retry_get_amqp_message`, so maybe tenacity is just complicating things here?

zaza/openstack/charm_tests/rabbitmq_server/tests.py
```diff
@@ -57,6 +58,41 @@ def _retry_get_amqp_message(self, check_unit, ssl=None, port=None):
...
+
+        try:
+            logging.info('Looking for message {}'.format(amqp_msg))
+            # Validate amqp message content
+            assert amqp_msg == amqp_msg_rcvd
+            logging.info('Message {} received OK.'.format(amqp_msg_counter))
+        except AssertionError as err:
+            logging.info('Expected: {}'.format(amqp_msg))
+            logging.info('Actual:   {}'.format(amqp_msg_rcvd))
+            raise err
```
AurelienLourot, line 94: Thanks, this is now indeed better but I still see this:

This line may raise an AssertionError 31 times, swallowed 30 times by tenacity.retry and then finally re-raised, correct? So then `_search_for_message()` can in theory raise an AssertionError too, correct?

If you agree, these two things should still be extended:
* rtype on line 79
* except type on line 162

Thanks!

ajkavanagh (commented):
Just minor comments for thoughts.

AurelienLourot (changes requested):

zaza/openstack/charm_tests/rabbitmq_server/tests.py
AurelienLourot: yes exactly, thank you :D

AurelienLourot (approved):
LGTM now and all comments have been addressed, thanks!

PR #686, opened 2021-12-17
Use full binary paths for neutron test_901_pause_resume

coreycb:
When pgrep_full=True is used, the resulting pgrep can match
more binaries than expected if the search pattern isn't specific
enough. This change makes the search patterns more specific
in order to only match the expected binaries.

Closes-Bug: #1933338


ajkavanagh (approved):
LGTM; thanks for the patch.

PR #688, opened 2022-01-07
Switch unit.public_address to unit.get_public_address()

ajkavanagh:
Due to the bug [1] on OpenStack providers, unit.public_address doesn't
actually work reliably.  The fix [2] is only for the async function
unit.get_public_address().  Sadly, zaza relied on unit.public_address
and so it needs this patch for juju 2.9 support on OpenStack providers.

This patch relies on an associated patch in zaza [3]; thus this will
fails its tests until that passes.

[1]: https://github.com/juju/python-libjuju/issues/551
[2]: https://github.com/juju/python-libjuju/pull/600
[3]: https://github.com/openstack-charmers/zaza/pull/468

### DIFF
--- a/unit_tests/utilities/test_zaza_utilities_swift.py
+++ b/unit_tests/utilities/test_zaza_utilities_swift.py
@@ -91,12 +91,21 @@ class TestSwiftUtils(ut_utils.BaseTestCase):
         self.patch_object(juju_utils, 'get_full_juju_status')
         self.patch_object(zaza.model, 'get_application_config')
         self.patch_object(zaza.model, 'get_units')
+        self.patch_object(zaza.model, 'get_unit_public_address')
+
+        def _get_unit_public_address(u):
+            return u.public_address
+
+        self.get_unit_public_address.side_effect = _get_unit_public_address
+
         juju_status = mock.MagicMock()
         juju_status.applications = {}
         self.get_full_juju_status.return_value = juju_status
 
         for app_name, units in app_units.items():
-            expected_topology[units[0].public_address]['unit'] = units[0]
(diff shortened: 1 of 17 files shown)


unit_tests/utilities/test_zaza_utilities_swift.py
```diff
@@ -91,12 +91,21 @@ def test_get_swift_storage_topology(self):
...
+        self.get_unit_public_address.side_effect = _get_unit_public_address
+
         juju_status = mock.MagicMock()
         juju_status.applications = {}
         self.get_full_juju_status.return_value = juju_status
 
         for app_name, units in app_units.items():
-            expected_topology[units[0].public_address]['unit'] = units[0]
+            # ip = zaza.model.get_unit_public_address(units[0])
+            ip = units[0].public_address
```
AurelienLourot, line 107: shouldn't it be the previous line instead, i.e. shouldn't we call get_unit_public_address() instead of using the public_address property?

AurelienLourot (changes requested):
LGTM, just one question, thanks!

unit_tests/utilities/test_zaza_utilities_swift.py
ajkavanagh: Hmm, strange.  Good catch; I will check the code again.

ajkavanagh: So, the commented out line is incorrect. This is actually a unit test of a test (yes, really!) and so *here* units[0].public_address is the mock further up in the code.

AurelienLourot (approved):

PR #691, opened 2022-01-12
add glance-simplestreams-syns to openstack upgrade exclude list

Basdbruijne:
glance-simplestreams-sync does have the `source` option, but this does not seem to point to the openstack version:

```
$ juju config glance-simplestreams-sync
...
  source:
    description: DEPRECATED - option no longer used and will be removed
    source: user
    type: string
    value: distro
...
```

The value here is set by me to avoid errors with the upgrade testing, but is normally left unset in sqa deployments.


ajkavanagh (approved):
LGTM

PR #694, opened 2022-01-19
Make run_package_change_test more robust

gnuoy:
The run_package_change_test disables auto restarts and then
simulates a packages update. It then checks the service has been
marked as in need of restarting. However bug #1955498 shows that
if a new service is introduced that restarts a disabled service
then this is not picked up by the tests.

This change queries the last active time of a service proir to
the package being updated. If this changes after the package
reconfiguration then the service has incorrectly been restarted.

Depends-On: https://github.com/openstack-charmers/zaza/pull/478


ajkavanagh (approved):
LGTM

PR #701, opened 2022-02-07
Add SR-IOV compute node reboot

coreycb:
SR-IOV compute nodes must be rebooted after deployment. Setup code
is added to do that. The computes with application name specified
in sriov_compute_application_name will be rebooted. Sample usage is
shown below.

configure:
  - zaza.openstack.charm_tests.nova.setup.reboot_sriov_computes
configure_options:
  sriov_compute_application_name: "nova-compute-sriov"

### DIFF
--- a/zaza/openstack/charm_tests/nova/setup.py
+++ b/zaza/openstack/charm_tests/nova/setup.py
@@ -14,11 +14,15 @@
 
 """Code for configuring nova."""
 
+import logging
 import tenacity
 
+from zaza import model
+import zaza.charm_lifecycle.utils as lifecycle_utils
 import zaza.openstack.utilities.openstack as openstack_utils
 from zaza.openstack.utilities import (
     cli as cli_utils,
+    generic as generic_utils,
 )
 import zaza.openstack.charm_tests.nova.utils as nova_utils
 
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/nova/setup.py
```diff
@@ -71,3 +75,20 @@ def manage_ssh_key(nova_client=None):
...
+    options = (lifecycle_utils
+               .get_charm_config(fatal=False)
+               .get('configure_options', {}))
+    sriov_compute = options.get('sriov_compute_application_name', None)
+
+    for unit in model.get_units(sriov_compute):
+        generic_utils.reboot(unit.entity_id)
+    logging.info("Waiting for {} units to reboot".format(sriov_compute))
+    model.block_until_wl_status_info_starts_with(sriov_compute,
+                                                 'Unit is ready')
```
ajkavanagh, line 91: I think that there is probably a race here?  i.e. `generic_utils.reboot(..)` is async in terms of returning from the subprocess call without waiting for the unit to reboot.

The model block waits until the `sriov_compute` application has all its units at "Unit is ready".  However, it's possible that the units won't have started rebooting (i.e. delays, etc) and that function could return before any units have started.  It's probably better to verify that the unit has rebooted by using a sentinel on the unit to make sure it reboots.

e.g. `/var/run/reboot-required` could be used, as if you populate it with 'restart', and then reboot, the reboot will remove the file.  e.g. [here](https://github.com/openstack-charmers/zaza-openstack-tests/blob/d3ab250984c5766267c5be40425aa8cff43dc372/zaza/openstack/utilities/parallel_series_upgrade.py#L620) the reboot is done and then the unit is waited on its absence.  This is probably something that should be built into reboot (i.e. injecting the file) so that the confirmation of the reboot can be determined.  I guess this will only work for Debian based distros?

coreycb: Thanks for the review @ajkavanagh. I'm not sure that generic_utils.reboot() is async but I may be confused. There's a reboot() and an async_reboot() defined in here: https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/utilities/generic.py#L345

ajkavanagh: It wasn't really a sync/async comment; more that there is a race between asking for reboot and looking for "Unit is ready".  My suggestion is:

1. Ensure that `/var/run/reboot-required` with `restart` in it, is on the unit to be rebooted.
2. Reboot the unit
3. use a different function (i.e. not `block_until_wl_status_info_starts_with(...)`) is used that actually looks on the unit that the `/var/run/reboot-required` file is missing.  This would mean that the unit had been rebooted.

i.e. it eliminates the race between asking for a reboot and checking for the unit being rebooted by using a sentinel file to verify that the reboot took place.

If this is coded into a utility for rebooting a bunch of units then `generic_utils.reboot_units(...)` would block until the units were blocked (with an appropriate timeout).  Does this make sense?

PR #710, opened 2022-02-24
Use snapshot_support=True for manila share type

coreycb:
The extra specs of a manila share type need to match the
share service's capabilities. [1]

[1] https://docs.openstack.org/manila/pike/admin/shared-file-systems-troubleshoot.html#id2

Closes-Bug: #1962204


coreycb:
I merged this on my own accidentally, clicking too quick. Billy to review.

wolsen (commented):
Looks okay to me

PR #719, opened 2022-03-03
Add NovaComputeNvidiaVgpuTest

AurelienLourot:
Will be validated here: https://review.opendev.org/c/openstack/charm-nova-compute-nvidia-vgpu/+/831743

### DIFF
--- a/zaza/openstack/charm_tests/nova/tests.py
+++ b/zaza/openstack/charm_tests/nova/tests.py
@@ -26,6 +26,7 @@ import zaza.model
 import zaza.openstack.charm_tests.glance.setup as glance_setup
 import zaza.openstack.charm_tests.test_utils as test_utils
 import zaza.openstack.configure.guest
+import zaza.openstack.utilities.generic as generic_utils
 import zaza.openstack.utilities.openstack as openstack_utils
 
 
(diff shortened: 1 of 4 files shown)


zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -434,6 +435,25 @@ def test_virsh_audit_action(self):
...
+                                                            nova_conf_file))
+
+            # See
+            # https://docs.openstack.org/nova/queens/admin/virtual-gpu.html
+            # https://docs.openstack.org/nova/ussuri/admin/virtual-gpu.html
+            # https://docs.openstack.org/releasenotes/nova/xena.html#deprecation-notes
+            self.assertTrue(('enabled_vgpu_types' in nova_conf) or
+                            ('enabled_mdev_types' in nova_conf))
+
+
```
coreycb, line 456: Out of curiosity what is the end result of the deployed bundles for the nova-compute-nvidia-vgpu charm? I assume there isn't a usable vgpu? If not we might consider testing the rest of the conf that gets written according to https://docs.openstack.org/nova/ussuri/admin/virtual-gpu.html etc.

PR #724, opened 2022-03-15
Fix Octavia wait_for_lb_resource()

freyes:
This change drops the use of AssertionError in favor of
LoadBalancerUnexpectedState that was introduced by commit acaeb62 which
stopped retrying on AssertionError.


gnuoy (approved):
LGTM, thanks

PR #680, opened 2021-12-01
Add glance tests for Cinder Storage backend

hemanthnakkina:
Add zaza tests for charm glance when cinder is configured
as storage backend.

Related-Bug: LP#1905042

PR #735, opened 2022-03-29
Add glance tests for Cinder Storage backend (#680)

hemanthnakkina:
Add zaza tests for charm glance when cinder is configured
as storage backend.

Related-Bug: LP#1905042

Co-authored-by: Hemanth Nakkina <hemanth.nakkina@canonical.com>
(cherry picked from commit 1a1dc0a79e2045f89c07af4b24a76b646456001e)

### DIFF
--- a/zaza/openstack/charm_tests/glance/tests.py
+++ b/zaza/openstack/charm_tests/glance/tests.py
@@ -17,6 +17,7 @@
 """Encapsulate glance testing."""
 
 import logging
+import math
 
 import boto3
 import zaza.model as model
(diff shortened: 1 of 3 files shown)


zaza/openstack/charm_tests/glance/tests.py
```diff
@@ -222,3 +223,46 @@ def test_100_create_delete_image(self):
...
+            image_url=openstack_utils.find_cirros_image(arch="x86_64"),
+            image_name=image_name,
+            backend="cinder",
+        )
+        images = openstack_utils.get_images_by_name(
+            self.glance_client, image_name)
+        self.assertEqual(len(images), 1)
+        image = images[0]
+
+        volume_name = 'image-'+image["id"]
```
ajkavanagh, line 257: This ought to fail pep8 - no spaces between the operator.  It would probably be better as:

```python
volume_name = f'image-{image["id"]}'
# or
volume_name = 'image-{}'.format(image["id"])
```

I'm trying to remember if we still support xenial (python3.5) in zaza - we probably do.  In which case the f-string won't work and it'll need to be the `.format()` version.

zaza/openstack/charm_tests/glance/tests.py
```diff
@@ -222,3 +223,46 @@ def test_100_create_delete_image(self):
...
+            self.cinder_client, volume_name)
+        self.assertEqual(len(volumes), 1)
+        volume = volumes[0]
+
+        logging.info(
+            "Checking glance image size {} matches volume size {} "
+            "GB".format(image["size"], volume.size))
+        image_size_in_gb = int(math.ceil(float(image["size"]) / 1024 ** 3))
+        self.assertEqual(image_size_in_gb, volume.size)
+        openstack_utils.delete_image(self.glance_client, image["id"])
```
ajkavanagh, line 268: Is this async in the cloud?  Worth checking if it is, and whether the test should wait here until the image is deleted so that it doesn't have the potential to pollute other tests.  If no async in the cloud, then please ignore this comment.

ajkavanagh (commented):
Please see my inline comments.  Thanks!

zaza/openstack/charm_tests/glance/tests.py
hemanthnakkina: Done

zaza/openstack/charm_tests/glance/tests.py
hemanthnakkina: DELETE API call returns 204 and so delete image is synchronous. Ignoring the comment

zaza/openstack/charm_tests/glance/tests.py
```diff
@@ -222,3 +223,46 @@ def test_100_create_delete_image(self):
...
         openstack_utils.delete_image(self.glance_client, image["id"])
+
+
+class GlanceCinderBackendTest(test_utils.OpenStackBaseTest):
+    """Encapsulate glance tests using cinder backend."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Run class setup for running glance tests with cinder backend."""
+        super(GlanceCinderBackendTest, cls).setUpClass()
```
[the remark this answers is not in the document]
hemanthnakkina: Done

zaza/openstack/charm_tests/glance/tests.py
ajkavanagh: > DELETE API call returns 204 and so delete image is synchronous. Ignoring the comment

Just to be slightly pedantic, I don't think this is true, but tracing the `openstack_utils.delete_image()` eventually calls `resource_removed()` which does wait for the resource to be deleted, so `delete_image()` is forced to be sync.

I think HTTP code 204 just means that the service accepted the request and there is no content.  e.g. it is a 200 with no content.  The semantics of the request, being sync or async, isn't embodied in the return code.  Happy to be corrected, though!

ajkavanagh (approved):

PR #745, opened 2022-04-11
Workaround the URL formatting issue for product streams

ajkavanagh:
When glance-simplestreams-sync charm registers the 'product-streams'
endpoint it uses a trailing slash.  It's not clear whether this is a bug
or not, and the behaviour may have changed in 2021.  The impact is that
the test fails, as it attempts to fetch with two '//' in the URL.  This
patch works around the test issue.  It's not clear whether endpoints
MUST be registered without a trailing slash or not; it would appear that
most are not, but a quick glance at ServerStack show a few with trailing
slashes.


gnuoy (approved):
lgtm

ajkavanagh:
I'm closing this as I've fixed it in the charm (https://review.opendev.org/c/openstack/charm-glance-simplestreams-sync/+/837396) and thus this isn't needed and would mask an actual bug.

PR #748, opened 2022-04-12
Stop overwriting files from upstream dependency

fnordahl:
The `zaza-openstack-tests` must be used with its upstream
dependency `zaza`.

However at present both projects provide a `zaza/__init__.py`
which in reality means that the file from the upstream dependency
is overwritten on install.

Remove the offending file.


fnordahl:
Interestingly not having a ``__init__.py`` in this project will remove the one from zaza on install anyway, probably the reason for having them duplicated before. The question then becomes, is that possible to avoid, or do we just need to keep having them duplicated and in sync?

ajkavanagh:
> Interestingly not having a `__init__.py` in this project will remove the one from zaza on install anyway, probably the reason for having them duplicated before. The question then becomes, is that possible to avoid, or do we just need to keep having them duplicated and in sync?

Really?  How strange.  I'm now reading https://peps.python.org/pep-0420/ with interest.

ajkavanagh:
@fnordahl I guess we can close this one as overtaken-by-events?

fnordahl:
Superseded by https://github.com/openstack-charmers/zaza-openstack-tests/pull/750

PR #751, opened 2022-04-12
Setup loop devices early on in the Ceph tests

lmlg:
There appears to be a race condition between the loop device creation and its actual usage. In order to prevent that, this PR setups these devices during the test initialization instead of right before using them in the 'add-disk' tests.


ajkavanagh:
Using https://review.opendev.org/c/openstack/charm-cinder-ceph/+/835283 to test this.

ChrisMacNaughton (approved):

PR #752, opened 2022-04-17
Fix requirements for zaza on stable/yoga

ajkavanagh:
zaza.openstack.tests requires zaza, but it should be the
stable/yoga version rather than the development tip.


wolsen (approved):
LGTM

PR #763, opened 2022-05-09
Switch zaza to a namespaced package

ajkavanagh:
This is to remove the need to sync the __init__.py file between zaza and
zaza.openstack.tests.  This PR is to enable zaza.openstack to be a
properly namespaces module.


ajkavanagh:
This should be landed with [zaza#480](https://github.com/openstack-charmers/zaza/pull/480) which is the async set of changes to work 'properly' with libjuju.

PR #767, opened 2022-05-20
Upgradeable tests

lmlg:
This PR allows functional tests to specify a list of charms that should be upgraded (via the 'upgrade_charms' key in the tests.yaml file), which makes test subclasses of a new type to upgrade said charms before running the tests.

### DIFF
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -38,7 +38,7 @@ import zaza.openstack.utilities.openstack as zaza_openstack
 import zaza.openstack.utilities.generic as generic_utils
 
 
-class CephLowLevelTest(test_utils.OpenStackBaseTest):
+class CephLowLevelTest(test_utils.UpgradeableTest):
     """Ceph Low Level Test Class."""
 
     @classmethod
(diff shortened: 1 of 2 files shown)


freyes:
Hi Luciano,

I haven't looked in detail the implementation proposed in this PR, although the part that the "stable" charm doesn't get validated (e.g. run all the z-o-t tests defined for the environment) seems to be inadequate, since many charms store local state (e.g. key-value local sqlite3 db), so those are probably the hardest scenarios to come up with a good migration strategy since the new charm will run the upgrade-charm without information about the system and what the previous charm did to it.

lmlg:
> Hi Luciano,
> 
> I haven't looked in detail the implementation proposed in this PR, although the part that the "stable" charm doesn't get validated (e.g. run all the z-o-t tests defined for the environment) seems to be inadequate, since many charms store local state (e.g. key-value local sqlite3 db), so those are probably the hardest scenarios to come up with a good migration strategy since the new charm will run the upgrade-charm without information about the system and what the previous charm did to it.

Good observation. I'm now looking into implementing a decorator for test methods so that the test writer can specify for which tests the upgrade is necessary. Maybe both solutions (i.e: a coarse and fine-grained) can coexist and be used depending on the needs of the tests.

ajkavanagh:
Hi @lmlg ; is this draft or ready-for-review; it's a bit hard to tell from the last comments on the PR; marking as stale.  If not stale, but more work to be done, please remove stale and switch it to draft.  Thanks!

PR #769, opened 2022-05-24
[stable/yoga] Stop failing on unknown ceilometer metrics

freyes:
The metric test is too strict and fails if it finds a metric that
is not explicitly listed in the test. Given the test does check
that the expected metrics are present there seems not need to fail
if a new one has added upstream.
    
(cherry picked from commit 21179a1125c0d4e116a81d20e29abc834dd215ed)


freyes:
this issue was found at https://review.opendev.org/c/openstack/charm-ceilometer-agent/+/842212

PR #770, opened 2022-05-24
Add retry logic around Trilio CRUD commands

gnuoy:


zaza/openstack/charm_tests/trilio/tests.py
```diff
@@ -160,6 +160,27 @@ def __init__(self, keystone_client):
         )
         self.auth_args = openstack_utils.get_cli_auth_args(keystone_client)
 
+    @tenacity.retry(
+        retry=tenacity.retry_if_exception_type(zaza_model.CommandRunFailed),
+        wait=tenacity.wait_fixed(10),  # interval between retries
+        stop=tenacity.stop_after_attempt(5))  # retry 10 times
```
ajkavanagh, line 166: Comment and param don't match?  e.g. 10 times vs 5.

zaza/openstack/charm_tests/trilio/tests.py
```diff
@@ -160,6 +160,27 @@ def __init__(self, keystone_client):
...
+    def wlm_run_cmd(self, remote_cmd):
+        """Run command on unit and return the output.
+
+        :param remote_cmd: Command to execute on unit
+        :type remote_cmd: string
+        :returns: Juju run output
+        :rtype: string
+        :raises: model.CommandRunFailed
+        """
+        logging.info("Running {}".format(remote_cmd))
```
ajkavanagh, line 176: It's entirely optional, but I tend to use the `%s` form of interpolating into strings in logging commands as it's slightly more efficient as if the log doesn't happen (e.g. no info logs logged due to settings), it doesn't actually do the interpolation ... i.e. it makes it a bit lazy.  The `.format()` always runs regardless of whether the log happens or not.  I only found this out this year, and I've been doing it ever since!

ajkavanagh (commented):
LGTM - just some minor comments on things I thought of whilst reviewing.

PR #774, opened 2022-05-26
Add new ceph-fs configuration test

ethanmye-rs:
This PR adds functional tests for the charm-ceph-fs config options added in the following review:
https://review.opendev.org/c/openstack/charm-ceph-fs/+/842555

### DIFF
--- a/zaza/openstack/charm_tests/ceph/fs/tests.py
+++ b/zaza/openstack/charm_tests/ceph/fs/tests.py
@@ -15,14 +15,18 @@
 """Encapsulate CephFS testing."""
 
 import logging
+import time
+import asyncio
 from tenacity import Retrying, stop_after_attempt, wait_exponential
-
 import zaza.model as model
 import zaza.openstack.charm_tests.neutron.tests as neutron_tests
 import zaza.openstack.charm_tests.nova.utils as nova_utils
 import zaza.openstack.charm_tests.test_utils as test_utils
 import zaza.openstack.configure.guest as guest
-import zaza.openstack.utilities.openstack as openstack_utils
+
+from zaza.openstack.utilities import (
+    openstack as openstack_utils,
+)
 
 
 class CephFSTests(test_utils.OpenStackBaseTest):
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/ceph/fs/tests.py
```diff
@@ -103,6 +107,64 @@ def verify(stdin, stdout, stderr):
...
+        def _change_conf_check(mds_config):
+            """Change configs, then assert to ensure config was set.
+
+            Doesn't return a value.
+            """
+            loop = asyncio.get_event_loop()
+            crt = model.async_set_application_config('ceph-fs', mds_config)
+            loop.run_until_complete(crt)
+            results = _get_conf()
+            time.sleep(1)  # needs time to release lock and update conf.
```
lmlg, line 142: Using sleeps like this to assert values later can be rather fragile. I suggest using `tenacity` and retrying a couple of times to make the test more robust.

zaza/openstack/charm_tests/ceph/fs/tests.py
```diff
@@ -103,6 +107,64 @@ def verify(stdin, stdout, stderr):
...
+            assert int(results['mds cache memory limit']) == \
+                   int(mds_config['mds-cache-memory-limit'])
+            assert float(results['mds cache reservation']) == \
```
lmlg, line 145: Comparing floats can be very error prone. Here I'd use `self.assertAlmostEqual`.

ChrisMacNaughton:
This isn't really a functional test; it's purely a test that config was written to a file, without any verification that Ceph even cares about that config line. Broadly, the desired approach for functional testing is to verify that the things you're trying to do actually causes the desired effect; as a basic example: testing that a CephFS share can be created, written to, and read from is infinitely more valuable than testing that we wrote the config we think we did.

It's possible to get the same confidence int his change via unit-testing in the charm itself, rather than `cat`ing the ceph.conf to ensure that the expected changes appeared.

ethanmye-rs:
> This isn't really a functional test; it's purely a test that config was written to a file, without any verification that Ceph even cares about that config line.

In my experience, Ceph ignores invalid config options in the .conf file, but that is anecdotal. From our earlier discussion, I agree querying the Ceph daemon directly is a better test to ensure Ceph actually understood and accepted the config values passed.

> It's possible to get the same confidence int his change via unit-testing in the charm itself, rather than `cat`ing the ceph.conf to ensure that the expected changes appeared.

I've left the unit tests as they are in the opendev repo and updated this repo with the 'functional' tests that ask Ceph for the configuration directly. Thanks for the help!

zaza/openstack/charm_tests/ceph/fs/tests.py
```diff
@@ -103,6 +106,73 @@ def verify(stdin, stdout, stderr):
...
+            """
+            configs = ["mds_cache_memory_limit",
+                       "mds_cache_reservation",
+                       "mds_health_cache_threshold"]
+            holder = {}
+            for config in configs:
+                cmd = "sudo ceph daemon mds." \
+                      "$HOSTNAME config show | grep {}".format(config)
+                conf = model.run_on_unit(self.TESTED_UNIT, cmd)
+                print(conf)
```
ChrisMacNaughton, line 127: It's generally preferred to use a log statement, rather than `print`

ChrisMacNaughton (commented):
I just have one issue with the tests now, it looks like a debug print was left.

zaza/openstack/charm_tests/ceph/fs/tests.py
ethanmye-rs: Good catch -- I just removed it, not really needed outside debugging.

ChrisMacNaughton (approved):

Issue #775, opened 2022-05-31
urllib.error.URLError: <urlopen error [Errno -3] Temporary failure in name resolution>


freyes:
Due to dns instability the download_image() could fail, it would be nice if it was wrapped with a retry decorator

https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/utilities/openstack.py#L2290

```python
# error codes that represent a dns issue
DNS_ISSUES = [...]

class DnsProblemError(Exception):
    pass

@retry(retry=retry_if_exception_type(DnsProblemError))
def download_image(image_url, target_file):
    ...
    try:
        urllib.request.urlretrieve(image_url, target_file)
    except urllib.error.URLError as ex:
        if type(ex.reason) == socket.gaierror and ex.reason.errno in DNS_ISSUES:
            raise DnsProblemError(ex)
```

Stacktrace seen in the CI

```
2022-05-21 12:10:00 [INFO] Downloading image bionic
2022-05-21 12:10:00 [INFO] Downloading http://cloud-images.ubuntu.com/bionic/current/bionic-server-cloudimg-amd64.img ...
2022-05-21 12:10:10 [ERROR] {'ceph': 'zaza-5c6d9259a4b4'}
2022-05-21 12:10:10 [ERROR] Model ceph (zaza-5c6d9259a4b4)
Traceback (most recent call last):
  File "/usr/lib/python3.8/urllib/request.py", line 1354, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/usr/lib/python3.8/http/client.py", line 1256, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1302, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1251, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1011, in _send_output
    self.send(msg)
  File "/usr/lib/python3.8/http/client.py", line 951, in send
    self.connect()
  File "/usr/lib/python3.8/http/client.py", line 922, in connect
    self.sock = self._create_connection(
  File "/usr/lib/python3.8/socket.py", line 787, in create_connection
    for res in getaddrinfo(host, port, 0, SOCK_STREAM):
[...]

PR #777, opened 2022-05-31
Add retry decorator for download_image function

Regis-Caelum:
## Issues/PR

Fix [Issue #775](https://github.com/openstack-charmers/zaza-openstack-tests/issues/775).

## Changes

```
*************Module*************
zaza/openstack/utilities/openstack.py
```

### DIFF
--- a/zaza/openstack/utilities/openstack.py
+++ b/zaza/openstack/utilities/openstack.py
@@ -22,6 +22,8 @@ import datetime
 import enum
 import io
 import itertools
+import socket
+
 import juju_wait
 import logging
 import os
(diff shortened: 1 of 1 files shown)


ajkavanagh (changes requested):
Thanks very much for your patch.  There are however, a couple of problems.

One minor comment on the actual change; however, the multiple formatting changes obscure the change and should not be part of the PR please.  Also, they cause a PEP8 failure as many of them are not compatible with the PEP8 options for the project.  Please use `tox -e pep8` and `tox -e py3` to verify your changes prior to submission.  Thanks.

PR #778, opened 2022-06-01
Add method for restarting tvault-contego

gnuoy:
This is a temporary workaround to the bug while a packaging fix
is investigated.


ajkavanagh (approved):
LGTM - understand that it's just a patch for the moment.

PR #784, opened 2022-06-09
Refresh kernel partition table between tests (xena)

lmlg:
Backport the change introduced in: https://github.com/openstack-charmers/zaza-openstack-tests/commit/3ab81b4ebb193f01ff4f2ef7e7dd11bbe6921976


coreycb:
Hi @lmlg , is this change still needed?

lmlg:
> Hi @lmlg , is this change still needed?

The last few runs have gone fine without this patchset, so I think we can close it. In case it's needed, I'll reopen these.

PR #787, opened 2022-06-15
Add launch_instance_retryer

gnuoy:
Add a method called launch_instance_retryer which will attempt
to launch an instance. In the event that it fails the instance
will be removed and another attempt to launch an instance will
be attempted.


ajkavanagh (approved):
LGTM

PR #791, opened 2022-06-22
Add jammy to list of ubuntu releases.

freyes:
The list UBUNTU_RELEASES is used by the CompareHostReleases class.


ajkavanagh (approved):
LGTM

PR #793, opened 2022-06-22
Don't test nrpe bits on rabbitmq-server.

freyes:
When the nrpe application is not deployed in the model skip the test
that checks for the correct configuration of the nagios checks, this is
because at the moment nrpe has no support for jammy, hence is not being
deployed in the jammy-yoga bundles.


ajkavanagh (approved):
LGTM

PR #797, opened 2022-06-24
Get ceph-dashboard's public addresses from the mon

ChrisMacNaughton:
At some point, Juju stopped returning a list of units for subordinate
charms, so this test now fails to find the address for the dashboard units


freyes (approved):
LGTM.

PR #803, opened 2022-06-24
Get ceph-dashboard's public addresses from the mon

ChrisMacNaughton:
At some point, Juju stopped returning a list of units for subordinate
charms, so this test now fails to find the address for the dashboard units


ChrisMacNaughton:
This change has been verified [here](https://review.opendev.org/c/openstack/charm-ceph-dashboard/+/847618)

wolsen (approved):
Not a blocker, but I am curious if it is intentional that Juju is not returning a list of units for subordinate charms - that feels broken to me.

ChrisMacNaughton:
It's worth mentioning that we're hitting this because of a workaround that we've added in getting unit addresses, where we use `juju status --format=yaml` and parse that output rather than libjuju's implementation.

freyes (approved):
LGTM.

PR #805, opened 2022-06-27
Refresh kernel partition table between tests (ussuri)

lmlg:
Backport the change introduced in: https://github.com/openstack-charmers/zaza-openstack-tests/commit/3ab81b4ebb193f01ff4f2ef7e7dd11bbe6921976


jadonn:
Do you still need this PR? If you do, what can we do to move it forward?

PR #806, opened 2022-06-30
Add nrpe check manila ganesha

chanchiwai-ray:


zaza/openstack/charm_tests/manila_ganesha/tests.py
```diff
@@ -56,3 +58,78 @@ def _restart_share_instance(self):
...
+            with attempt:
+                ret = generic_utils.check_commands_on_units(cmds, units)
+                self.assertIsNone(ret, msg=ret)
+
+    def test_903_nrpe_custom_cronjob_checks(self):
+        """Confirm that the NRPE custom cron job files are created."""
+        try:
+            zaza.model.get_application("nrpe")
+        except KeyError:
+            logging.warn("Skipping as nrpe is not deployed.")
```
ajkavanagh, line 91: Please use `self.skipTest(...)` to skip a test.

ajkavanagh (changes requested):
Please see inline; also note that the PR description needs filling out too.  Thanks.

chanchiwai-ray:
Thanks for the comments. I have refactored the code and update the description of the PR, please have a look again.

ajkavanagh (approved):
LGTM; that's a really nice refactor of the original patch and it is much easier to understand now.  Thanks.

zaza/openstack/charm_tests/manila_ganesha/tests.py
```diff
@@ -56,3 +58,65 @@ def _restart_share_instance(self):
...
+    def test_903_nrpe_custom_plugin_checks(self):
+        """Confirm that the NRPE custom plugin files are created."""
+        plugins = [
+            "check_nfs_conn",
+            "check_nfs_exports",
+            "check_nfs_services",
+        ]
+
+        commands = [
+            f"ls /usr/local/lib/nagios/plugins/{plugin}"
```
ajkavanagh, line 88: From the py35 test run:

```
zaza/openstack/charm_tests/manila_ganesha/tests.py:88:56: E999 SyntaxError: invalid syntax
```

Py3.5.sadly doesn't support f-strings.  You'll need to use `.format(...)` here instead.

ajkavanagh (changes requested):
See inline; unfortunately f-strings are not compatible with Python 3.5

zaza/openstack/charm_tests/manila_ganesha/tests.py
chanchiwai-ray: Done, thanks for the hints!

ajkavanagh (approved):
LGTM - thanks for the update to fix the pep8 error.

PR #811, opened 2022-07-01
RBD mirror test fixes (wallaby)

lmlg:


zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -153,6 +154,12 @@ def create_volume(cinder, volume_params, retry=20):
     return create_volume(cinder, volume_params)
 
 
+def remove_internal_pools(pools):
+    """Exclude the internal pools from the passed dict."""
+    for pool in INTERNAL_POOLS:
+        pools.pop(pool, None)
```
ajkavanagh, line 160: May I caution about writing this kind of function.  Essentially, it's using the `pools` parameter as both and in and out parameter, which is generally considered bad practice as it breaks the principle of least surprise.  It also increases cognitive load, particularly as the function's parameters are not described.  It would be much better to implement the function the following way which makes it pure (side-effect free) and thus easier to reason about.

```python
def remove_internal_pools(pools):
    """Remove the internal pools from the passed dictionary.

    :param pools: the pools passed.
    :type pools: Dict[str, Any]
    :returns: pools with the INTERNAL pools remove, if they were present
    :rtype: Dict[str, Any]
    """
    _pools = pools.copy()  # top level copy only
    for pool in INTERNAL_POOLS:
        _pools.pop(pool, None)
    return _pools

# ... later

site_a_pools = remove_internal_pools(site_a_pools)
site_b_pools = remove_internal_pools(site_b_pools)
```

Now it's really obvious that `site_a_pools` and `site_b_pools` are being updated purposefully.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -197,9 +204,12 @@ def get_pools(self):
                 'ceph-mon' + self.site_b_app_suffix,
                 model_name=self.site_b_model),
             model_name=self.site_b_model)
+        if not include_internal_pools:
+            remove_internal_pools(site_a_pools)
+            remove_internal_pools(site_b_pools)
         return sorted(site_a_pools.keys()), sorted(site_b_pools.keys())
```
ajkavanagh, line 210: Having just indicated above in `remove_internal_pools` that the function should be immutable, it occurs to me here, that this could be better written as:

```python
pools = set(itertools.chain(*site_a_pools,.keys(), *site_b_pools.keys())
return sorted(pools.difference(INTERNAL_POOLS))
```

Note, sorted returns an iterator, so the :rtype: on the function is wrong anyway.  It would need `tuple(..)` or `list(..)` to force a data structure return

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -197,9 +198,14 @@ def get_pools(self):
...
             model_name=self.site_b_model)
-        return sorted(site_a_pools.keys()), sorted(site_b_pools.keys())
-
-    def get_failover_pools(self):
+        site_a_pools = set(site_a_pools.keys())
+        site_b_pools = set(site_b_pools.keys())
+        if not include_internal_pools:
+            site_a_pools -= INTERNAL_POOLS
+            site_b_pools -= INTERNAL_POOLS
+        return list(site_a_pools), list(site_b_pools)
```
ajkavanagh, line 206: I don't know if it matters, but this is now returns both a different type (a `Tuple[List, List]` rather than `Tuple[Iterator, Iterator]`) and they are no longer sorted?

lmlg: The function did return lists (calling `sorted` on a dict's keys returns a mutable list), and the call to `sorted` was unneeded.

PR #820, opened 2022-07-07
Prometheus has registered a non-zero num of osds

gnuoy:
Change prometheus test to check it has a non-zero amount of OSDs
registered rather than a specific number. This stops the number of
OSDs having to be managed in the tests.


ajkavanagh (approved):
LGTM; this is much better from a testing perspective.

PR #821, opened 2022-07-08
Add zaza.clean_up_libjuju_thread

gnuoy:
Since a recent update to zaza to put libjuju asyncio on a background
thread the openstack upgrade action tests have been causing the
whole unit tests run to hang. It looks like nose is waiting for
threads to be cleaned up. This change mirrors the chance from zaza
to add the cleanup.


gnuoy:
Superseded by https://github.com/openstack-charmers/zaza-openstack-tests/pull/822

PR #822, opened 2022-07-08
Ensure unittests use zaza non-threaded

ajkavanagh:
zaza normally runs libjuju async functions in a background thread
to enable zaza to be blocking, but keep the libjuju async functions
'ticking over' between blocking calls.

This can break unit tests as starting the zaza background thread will
hang unless it is cleaned up.  A simple solution is just to use zaza in
non-threaded mode in the unit tests, and zaza provides a mechanism for
this via zaza.RUN_LIBJUJU_IN_THREAD = False.


gnuoy (approved):
LGTM, thanks

Issue #825, opened 2022-07-08
Assertion fail on charm-manila: KeyError: 'manila-ganesha-az1'


FrancescoDeSimone:
It appears to be that the [assertion in the ```_run_nrpe_check_command``` function](https://github.com/openstack-charmers/zaza-openstack-tests/blob/5aa0f13eaa7eb0f64e7d659fbd4ab1b1ff743bb5/zaza/openstack/charm_tests/manila_ganesha/tests.py#L68) does not take into account manila-ganesha unit in charm-manila  

Here the ```tox -e func-target ganesha:focal-xena``` output 

```
2022-07-08 17:18:02.441018 | focal-medium | 2022-07-08 17:18:02 [INFO] ======================================================================
2022-07-08 17:18:02.441300 | focal-medium | 2022-07-08 17:18:02 [INFO] ERROR: test_903_nrpe_custom_plugin_checks (zaza.openstack.charm_tests.manila_ganesha.tests.ManilaGaneshaTests)
2022-07-08 17:18:02.441335 | focal-medium | 2022-07-08 17:18:02 [INFO] Confirm that the NRPE custom plugin files are created.
2022-07-08 17:18:02.441346 | focal-medium | 2022-07-08 17:18:02 [INFO] ----------------------------------------------------------------------
2022-07-08 17:18:02.441355 | focal-medium | 2022-07-08 17:18:02 [INFO] Traceback (most recent call last):
2022-07-08 17:18:02.441363 | focal-medium | 2022-07-08 17:18:02 [INFO]   File "/home/ubuntu/src/review.opendev.org/openstack/charm-manila/src/.tox/func-target/lib/python3.8/site-packages/zaza/openstack/charm_tests/manila_ganesha/tests.py", line 92, in test_903_nrpe_custom_plugin_checks
2022-07-08 17:18:02.441441 | focal-medium | 2022-07-08 17:18:02 [INFO]     self._run_nrpe_check_command(commands)
2022-07-08 17:18:02.441500 | focal-medium | 2022-07-08 17:18:02 [INFO]   File "/home/ubuntu/src/review.opendev.org/openstack/charm-manila/src/.tox/func-target/lib/python3.8/site-packages/zaza/openstack/charm_tests/manila_ganesha/tests.py", line 68, in _run_nrpe_check_command
[...]

PR #826, opened 2022-07-11
Skip nrpe check if manila-ganesha-az1 not deployed

FrancescoDeSimone:
nrpe check will be skip if manila-ganesha-az1 unit can not be found as reported in https://github.com/openstack-charmers/zaza-openstack-tests/issues/825

Closes Issue:  #825


ajkavanagh (approved):
LGTM - thanks for the patch.

PR #827, opened 2022-07-12
Fix test to work with vault with raft backend

swalladge:
For https://review.opendev.org/c/openstack/charm-vault/+/847964

The tests here run a series of commands on the vault client,
without waiting in between.
This is usually fine, but with the raft backend,
there can be delay between unsealing vault,
and the raft cluster settling.
During this time, most vault commands will return errors.

This should only affect these tests;
the charm itself will set status as appropriate while raft cluster
is settling.

### DIFF
--- a/zaza/openstack/charm_tests/vault/tests.py
+++ b/zaza/openstack/charm_tests/vault/tests.py
@@ -63,8 +63,18 @@ class BaseVaultTest(test_utils.OpenStackBaseTest):
         if cls.vip_client:
             cls.clients.append(cls.vip_client)
         cls.vault_creds = vault_utils.get_credentials()
+
+        # This little dance is to ensure a correct init and unseal sequence,
+        # for the case of vault with the raft backend.
+        # It will also work fine in other cases.
+        # The wait functions will raise AssertionErrors on timeouts.
+        init_client = vault_utils.wait_and_get_initialized_client(cls.clients)
+        vault_utils.unseal_all([init_client], cls.vault_creds['keys'][0])
+        vault_utils.wait_until_all_initialised(cls.clients)
         vault_utils.unseal_all(cls.clients, cls.vault_creds['keys'][0])
+
         vault_utils.auth_all(cls.clients, cls.vault_creds['root_token'])
+        vault_utils.wait_for_ha_settled(cls.clients)
         vault_utils.ensure_secret_backend(cls.clients[0])
 
     def tearDown(self):
--- a/zaza/openstack/charm_tests/vault/utils.py
+++ b/zaza/openstack/charm_tests/vault/utils.py
@@ -21,6 +21,7 @@ import hvac
 import logging
 import requests
 import tempfile
+import time
 import urllib3
 import yaml
 import tenacity
@@ -278,6 +279,75 @@ def ensure_secret_backend(client):
             options={'version': 1})
     except hvac.exceptions.InvalidRequest:
         pass
+
+
+def wait_for_ha_settled(clients):
+    """Wait until vault ha is settled (for all passed clients).
+
+    Raise an AssertionError if any are not settled within 2 minutes.
+    This function is effectively a no-op for non-ha vault.
+    Requires all vault units to be unsealed.
+
+    :param clients: Clients to use to talk to vault
+    :type clients: List[CharmVaultClient]
+    :raises: AssertionError
+    """
+    for client in clients:
+        for attempt in tenacity.Retrying(
+            reraise=True,
+            wait=tenacity.wait_fixed(10),
+            stop=tenacity.stop_after_attempt(12),  # wait for max 2 minutes
+        ):
+            with attempt:
+                # ha_status could also raise other errors,
+                # eg. if unsealing still in progress.
+                # This is why we're using tenacity here;
+                # avoids needing to manually handle other exceptions.
+                ha_status = client.hvac_client.ha_status
+                if (
+                    not ha_status.get('leader_address') and
+                    ha_status.get('ha_enabled')
+                ):
+                    raise AssertionError('Timeout waiting for ha to settle')
+
+
+def wait_until_all_initialised(clients):
+    """Wait until vault is initialized (for all passed clients).
+
+    Raise an AssertionError if any are not initialized within 2 minutes.
+
+    :param clients: Clients to use to talk to vault
+    :type clients: List[CharmVaultClient]
+    :raises: AssertionError
+    """
+    for client in clients:
+        for _ in range(12):
+            if is_initialized(client):
+                break
+            time.sleep(10)  # max 2 minutes (12 x 10s)
+        else:
+            raise AssertionError("Timeout waiting for vault to initialize")
+
+
+def wait_and_get_initialized_client(clients):
+    """Wait until at least one vault unit is initialized.
+
+    And return the initialized client.
+    Raise an AssertionError
+    if no initialized clients are found within 2 minutes.
+
+    :param clients: Clients to use to talk to vault
+    :type clients: List[CharmVaultClient]
+    :raises: AssertionError
+    :returns: an initialized client
+    :rtype: CharmVaultClient
+    """
+    for _ in range(12):
+        for client in clients:
+            if is_initialized(client):
+                return client
+        time.sleep(10)  # max 2 minutes (12 x 10s)
+    raise AssertionError("Timeout waiting for vault to initialize")
 
 
 def find_unit_with_creds():


zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -264,6 +264,16 @@ def is_initialized(client):
...
+# auto-retry this because this is called immediately after
+# unsealing vault, and vault can take some time to be ready.
+@tenacity.retry(
+    retry=tenacity.retry_if_exception_type((
+        hvac.exceptions.InternalServerError,  # raft cluster may be settling
+        hvac.exceptions.VaultDown, # vault may still be unsealing
+    )),
+    reraise=True,
+    wait=tenacity.wait_fixed(5),
+    stop=tenacity.stop_after_attempt(10))
```
ajkavanagh, line 276: I'd much rather actually check that vault is ready rather than failing and relying on timeouts.  This makes it obvious that the issue isn't, say, slow hardware or a regression in vault/charm, rather than it simply not working.

ajkavanagh:
Hi @swalladge ; this is shaping up nicely.  Prior to merging it would be good to have a gerrit review where the test is exercised.  Would it be possible to add it to the description?  Thanks.

UtkarshBhatthere (changes requested):
Few Suggestions on docstring and coding style.

ajkavanagh:
> @ajkavanagh it looks like one of the linters (flake8?) has updated and is now complaining about code style issues in unrelated areas of the code. Would you like them fixed here (in a separate commit), or in a separate PR?
> 
> snippet:
> 
> ```
>   pep8 run-test: commands[0] | flake8 zaza unit_tests
>   zaza/openstack/configure/network.py:192:80: E501 line too long (80 > 79 characters)
>   zaza/openstack/utilities/openstack.py:569:80: E501 line too long (80 > 79 characters)
>   zaza/openstack/utilities/openstack.py:2985:18: E275 missing whitespace after keyword
>   zaza/openstack/charm_tests/barbican/tests.py:39:23: E275 missing whitespace after keyword
>   zaza/openstack/charm_tests/barbican/tests.py:60:15: E275 missing whitespace after keyword
> ```

Good question! It's usually better to fix in a separate PR and then merge that (once merged) into this code so that it will pass.

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -76,7 +77,14 @@ def initialize(self):
         else:
             self.vault_creds = init_vault(self.unseal_client)
             store_credentials(self.vault_creds)
-        self.initialized = is_initialized(self.unseal_client)
+
+        for _ in range(10):
+            if is_initialized(self.unseal_client):
+                break
+            time.sleep(10)
```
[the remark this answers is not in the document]
ajkavanagh: I think using tenacity might be a bit of overkill here.  The method is "initialize" and the outcome of the function should be "it's initialised".  I'm not sure it's worth pulling out a wait-for-initialised function as it would probably only be used from this function.

However, the original code is a bit of a mess; it has a `@property`  `is_initialized` which just returns the attribute `initialized`, which seems like a waste; the intention of using a property would be to do some work to find out if the vault was initialised.  e.g. the property auto to return `is_initialized(self.unseal_client)`; it's all a bit strange.

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -280,6 +288,35 @@ def ensure_secret_backend(client):
...
+    for client in clients:
+        wait_for_ready(client)
+
+
+@tenacity.retry(
+    reraise=True,
+    wait=tenacity.wait_fixed(10),
+    stop=tenacity.stop_after_attempt(20))
+def wait_for_ready(client):
+    """Return once vault is ready.
```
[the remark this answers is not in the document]
ajkavanagh: @UtkarshBhatthere if you were being really picky ( :) ) then I'd actually suggest a rename of the function.  If you look at what it does:

1. It checks to see if the client is ready.
2. It repeats that check a few times after 200 seconds (which seems like a long time to me).
3. It raises an exception.

I'd probably rename the function to `wait_until_ready`, and instead of raising `Exception`, I'd raise `AssertionError`.

Also, the docstring ought to contain the exception that may be raised and under what circumstances.  I'd also move the retry code into the function using a `with` statement; after all the function is supposed to wait until it's ready not just check and bail out; the wrap-around of tenacity makes that less obvious; moving tenacity *into* the function makes it clear that the function waits until it's ready for a certain during and then fails.

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -280,6 +288,35 @@ def ensure_secret_backend(client):
...
+def wait_for_ready_all(clients):
+    """Return once vault is ready (for all clients).
+
+    Raise an error if it's not ready within a reasonable time.
+
+    :param clients: Clients to use to talk to vault
+    :type clients: List[CharmVaultClient]
+    """
+    for client in clients:
+        wait_for_ready(client)
```
ajkavanagh, line 300: This code could wait up to 3x200 seconds for the vault to settle; is that really what's wanted?

zaza/openstack/charm_tests/vault/utils.py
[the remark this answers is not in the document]
UtkarshBhatthere: That's even better! As a general rule I stay away from using terms which overlap with syntax (return, yield etc) since they make it a mix bag. Also +1 on AssertionError suggestion, catching Exception directly can result in unexpected errors sneak past quietly. 
PS: I was being picky :)

zaza/openstack/charm_tests/vault/utils.py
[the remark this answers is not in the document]
UtkarshBhatthere: Sure, for some reason, I am not able to mark this comment as resolved (Where's the button gone ?). Please feel free to do so.

UtkarshBhatthere (approved):

zaza/openstack/charm_tests/vault/utils.py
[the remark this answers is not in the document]
ajkavanagh: > Thanks for your comments, I'll make updates as necessary. :)
> 
> > It repeats that check a few times after 200 seconds (which seems like a long time to me).
> 
> Maybe I'm not interpreting the tenacity rules correctly? I `wait_fixed(10), stop_after_attempt(20)` means it will check every 10 seconds, and stop after 20 checks.

Yup, that's correct, and 200 seconds if all the retries are exhausted.  Which seems a long time for the vault to be initialised.  In practice it'll probably be "a few seconds" or "never", but I could be wrong about that.

zaza/openstack/charm_tests/vault/utils.py
ajkavanagh: Actually, I was wrong; it's just 200 seconds and then "bang" with an exception, so it's *probably* okay.

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -280,6 +281,65 @@ def ensure_secret_backend(client):
         pass
 
 
+def wait_until_ready(clients):
+    """Wait until vault is ready (for all clients).
```
ajkavanagh, line 285: Strictly speaking, it's for the passed param `clients`?

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -280,6 +281,65 @@ def ensure_secret_backend(client):
...
+
+    Raise an error if it's not ready within a reasonable time.
```
ajkavanagh, line 287: Okay, we're reaching end-game reviews now!  :)

Let's be specific with the docstrings please:

1. define what 'ready' means in this function.  e.g. has a `leader_address` and has an `ha_enabled` attribute on the clients' `ha_status`.
2. Poll for up to 2 minutes for this status to be achieved.
3. Raise an AssertionError exception if the status is not achieved for all the clients passed.

docstrings are for people to read to understand what the intent of the function is and what they can expect without reading the code.

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -280,6 +281,65 @@ def ensure_secret_backend(client):
...
+        for attempt in tenacity.Retrying(
+            reraise=True,
+            wait=tenacity.wait_fixed(10),
+            stop=tenacity.stop_after_attempt(12),  # wait for max 2 minutes
+        ):
+            with attempt:
+                ha_status = client.hvac_client.ha_status
+                if (
+                    not ha_status.get('leader_address') and
+                    ha_status.get('ha_enabled')
```
ajkavanagh, line 303: Am I reading this correctly. "raise an exception if it's not got a leader address AND does have ha_enabled"?  So this function only works for units that are 'ha_enabled'?  That might be useful for the docstring and the name of the function?

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -280,6 +281,65 @@ def ensure_secret_backend(client):
...
+                ):
+                    raise AssertionError('vault not ready: no leader_address')
+
+
+def wait_until_all_initialised(clients):
+    """Wait until vault is initialised (for all clients).
+
+    Raise an error if it's not initialised within a reasonable time.
```
ajkavanagh, line 311: As above.

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -280,6 +281,65 @@ def ensure_secret_backend(client):
...
+                break
+            time.sleep(10)
+        else:
+            raise AssertionError("Timeout waiting for vault to initialise")
+
+
+def wait_until_one_initialised(clients):
+    """Wait until at least one vault unit is initialised.
+
+    Raise an error if it's not initialised within a reasonable time.
```
ajkavanagh, line 329: As above.

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -280,6 +281,65 @@ def ensure_secret_backend(client):
...
+
+    :param clients: Clients to use to talk to vault
+    :type clients: List[CharmVaultClient]
+    :raises: AssertionError
```
ajkavanagh, line 333: This block needs an `:returns:` and `:rtype:` as it returns the `CharmVaultClient` that is initialised.

zaza/openstack/charm_tests/vault/tests.py
```diff
@@ -63,8 +63,17 @@ def setUpClass(cls):
         if cls.vip_client:
             cls.clients.append(cls.vip_client)
         cls.vault_creds = vault_utils.get_credentials()
+
+        # This little dance is to ensure a correct init and unseal sequence,
+        # for the case of vault with the raft backend.
+        # It will also work fine in other cases.
```
ajkavanagh, line 69: Note, that I'd add a note here that this will raise an AssertionError if the vault units don't unseal for the class.

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -280,6 +281,65 @@ def ensure_secret_backend(client):
...
+    for client in clients:
+        for _ in range(12):
+            if is_initialized(client):
+                break
+            time.sleep(10)
+        else:
+            raise AssertionError("Timeout waiting for vault to initialise")
+
+
+def wait_until_one_initialised(clients):
```
ajkavanagh, line 326: In the true spirit of "naming things is hard", this function actually does "return an initialised client", so I'd probably name it differently: `def get_any_initialised_client_from(clients):`

The docstring then says that it's polling for up to 2 minutes to get the first client.  You can also add the prefix `wait_and_` that is even more explicit, but then the function name starts to get "too" long, which can be annoying.

zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -280,6 +281,65 @@ def ensure_secret_backend(client):
...
+
+    :param clients: Clients to use to talk to vault
+    :type clients: List[CharmVaultClient]
+    :raises: AssertionError
+    """
+    for client in clients:
+        for attempt in tenacity.Retrying(
+            reraise=True,
+            wait=tenacity.wait_fixed(10),
+            stop=tenacity.stop_after_attempt(12),  # wait for max 2 minutes
```
ajkavanagh, line 297: So using `10` and `12` as bare literals *once* is probably okay, but using them multiple times through the code starts to look a bit suspect.  I'd be inclined to add some consts to either the class or the module for them if they are repeated, but there are no hard-and-fast rules about this, so it's also okay to not do that (as they aren't really configurable).

ajkavanagh (commented):
All my comments are just that; they are purely from a maintenance perspective when reading the code - the actual code changes are fine.

ajkavanagh (approved):
LGTM

PR #828, opened 2022-07-15
RBD mirror test fixes (ussuri)

lmlg:
This PR is being tested here: https://review.opendev.org/c/openstack/charm-ceph-rbd-mirror/+/848477

### DIFF
--- a/zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
+++ b/zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
@@ -31,7 +31,7 @@ from zaza.openstack.charm_tests.glance.setup import (
 
 
 DEFAULT_CINDER_RBD_MIRRORING_MODE = 'pool'
-INTERNAL_POOLS = set(('.mgr', 'device_health_metrics'))
+INTERNAL_POOLS = frozenset(('.mgr', 'device_health_metrics'))
 
 
 def get_cinder_rbd_mirroring_mode(cinder_ceph_app_name='cinder-ceph'):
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -31,6 +31,7 @@
 
 
 DEFAULT_CINDER_RBD_MIRRORING_MODE = 'pool'
+INTERNAL_POOLS = set(('.mgr', 'device_health_metrics'))
```
ajkavanagh, line 34: It's probably going to sound a bit mad, but this would be better as a tuple, rather than a set.  This is because, as a tuple, it would be an immutable collection, rather than a mutable one, which just removes a whole class of potential errors; as it's being used as a constant, and only as something to iterate over (e.g. used as an iterator), then this won't change the semantics of the code:

```python
INTERNAL_POOLS = (('.mgr', 'device_health_metrics'), )
```

Note the extra `,` above to ensure that it's a single item tuple.

lmlg: what about using a frozenset ? It would be a constant, and none of the code would need to change (it would still be usable by `set.difference`).

lmlg: yeah, I'll go with a `frozenset`. It's the closest to the current semantics while also being a constant and avoiding any potential pitfalls.

ajkavanagh:
@lmlg one fine request, if you get a moment please?  Could you add the gerrit review that is targeting to use this change (if possible).  That way we get an additional check that the test code is okay, and it also helps to tie things together in the histories?  Thanks!

lmlg:
> @lmlg one final request, if you get a moment please? Could you add the gerrit review that is targeting to use this change (if possible). That way we get an additional check that the test code is okay, and it also helps to tie things together in the histories? Thanks!

Certainly. I'll add it in the PR description.

PR #840, opened 2022-07-20
Add Testcases for Ceph-RGW multisite migration

UtkarshBhatthere:
1. Added Auxiliary methods and refactored code.
2. Switched to Boto3 (from Swift Client) for accessing individual endpoints.
3. A new test for (non-pristine) secondary site migration block is added.
4. 2 existing (but redundant) Data Integrity test cases are merged as 1 logically structured test-case.
5. Failover and Recovery test-case is passing on multisite models.

Tests used by: https://review.opendev.org/c/openstack/charm-ceph-radosgw/+/848739

### DIFF
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -892,8 +892,8 @@ class CephRGWTest(test_utils.OpenStackBaseTest):
                 "ceph-radosgw", "ceph-radosgw:master",
                 "slave-ceph-radosgw:slave"
             )
+            zaza_model.block_until_unit_wl_status('ceph-radosgw/0', "blocked")
 
-        zaza_model.block_until_unit_wl_status('ceph-radosgw/0', "blocked")
         zaza_model.block_until_unit_wl_status('ceph-radosgw/0', "active")
         self.wait_for_sync('slave-ceph-radosgw', isPrimary=False)
         self.wait_for_sync('ceph-radosgw', isPrimary=True)
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -1023,11 +1023,11 @@ class CephRGWTest(test_utils.BaseCharmTest):
                 self.secondary_rgw_app + ":slave"
             )
             zaza_model.block_until_unit_wl_status(
-                self.secondary_rgw_app + "/0", "waiting"
+                self.secondary_rgw_unit, "waiting"
             )
-        logging.info('Waiting for model to stabalize')
+
         zaza_model.block_until_unit_wl_status(
-            self.secondary_rgw_app + "/0", "active"
+            self.secondary_rgw_unit, "active"
         )
         logging.info('Waiting for Data and Metadata to Synchronize')
         self.wait_for_sync(self.secondary_rgw_app, isPrimary=False)
@@ -1075,12 +1075,8 @@ class CephRGWTest(test_utils.BaseCharmTest):
 
         logging.info('Checking multisite failover/failback')
         # Create RGW IO client.
-        primary_endpoint = self.get_rgw_endpoint(
-            self.primary_rgw_app + "/0"
-        )
-        secondary_endpoint = self.get_rgw_endpoint(
-            self.secondary_rgw_app + "/0"
-        )
+        primary_endpoint = self.get_rgw_endpoint(self.primary_rgw_unit)
+        secondary_endpoint = self.get_rgw_endpoint(self.secondary_rgw_unit)
         access_key, secret_key = self.get_client_keys() 
         primary_client = boto3.resource("s3",
                                         verify=False,
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -28,8 +28,6 @@ import urllib3
 
 import tenacity
 
-from swiftclient.exceptions import ClientException
-
 import zaza.openstack.charm_tests.test_utils as test_utils
 import zaza.model as zaza_model
 import zaza.openstack.utilities.ceph as zaza_ceph
@@ -43,6 +41,7 @@ import zaza.openstack.utilities.generic as generic_utils
 urllib3.disable_warnings(
     urllib3.exceptions.InsecureRequestWarning
 )
+
 
 class CephLowLevelTest(test_utils.OpenStackBaseTest):
     """Ceph Low Level Test Class."""
@@ -707,8 +706,15 @@ class CephRGWTest(test_utils.BaseCharmTest):
         except KeyError:
             return False
 
-    def get_rgwadmin_cmd_skeleton(self, unit_name:str):
-        '''Get radosgw-admin cmd skeleton with rgw.hostname populated key'''
+    def get_rgwadmin_cmd_skeleton(self, unit_name):
+        """
+        Get radosgw-admin cmd skeleton with rgw.hostname populated key.
+
+        :param unit_name: Unit on which the complete command would be run.
+        :ptype unit_name: str
+        :returns: hostname filled basic command skeleton
+        :rtype: str
+        """
         app_name = unit_name.split('/')[0]
         juju_units = zaza_model.get_units(app_name)
         unit_hostnames = generic_utils.get_unit_hostnames(juju_units)
@@ -718,15 +724,22 @@ class CephRGWTest(test_utils.BaseCharmTest):
     @tenacity.retry(wait=tenacity.wait_exponential(multiplier=10, max=300),
                     reraise=True, stop=tenacity.stop_after_attempt(10),
                     retry=tenacity.retry_if_exception_type(AssertionError))
-    def wait_for_sync(self, application, isPrimary=False):
-        """Wait for slave to secondary to show it is in sync."""
+    def wait_for_sync(self, application, is_primary=False):
+        """
+        Wait for slave to secondary to show it is in sync.
+
+        :param application: RGW application which has to be waited for
+        :ptype application: str
+        :param is_primary: whether application is primary site endpoint
+        :ptype is_primary: boolean
+        """
         juju_units = zaza_model.get_units(application)
         unit_hostnames = generic_utils.get_unit_hostnames(juju_units)
         sync_states = list()
         data_check = 'data is caught up with source'
         meta_primary = 'metadata sync no sync (zone is master)'
         meta_secondary = 'metadata is caught up with master'
-        meta_check = meta_primary if isPrimary else meta_secondary
+        meta_check = meta_primary if is_primary else meta_secondary
         for unit_name, hostname in unit_hostnames.items():
             key_name = "rgw.{}".format(hostname)
             cmd = 'radosgw-admin --id={} sync status'.format(key_name)
@@ -738,18 +751,27 @@ class CephRGWTest(test_utils.BaseCharmTest):
     @tenacity.retry(wait=tenacity.wait_exponential(multiplier=1, max=60),
                     reraise=True, stop=tenacity.stop_after_attempt(12))
     def fetch_rgw_object(self, target_client, container_name, object_name):
-        '''
-        Fetch RGW object content
+        """
+        Fetch RGW object content.
+
         :param target_client: boto3 client object configured for an endpoint.
+        :ptype target_client: str
         :param container_name: RGW bucket name for desired object.
+        :ptype container_name: str
         :param object_name: Object name for desired object.
-        '''
+        :ptype object_name: str
+        """
         return target_client.Object(container_name, object_name).get()[
             'Body'
         ].read().decode('UTF-8')
 
-    def promote_rgw_to_primary(self, app_name:str):
-        '''Promote provided app to Primary and update period at new secondary'''
+    def promote_rgw_to_primary(self, app_name: str):
+        """
+        Promote provided app to Primary and update period at new secondary.
+
+        :param app_name: Secondary site rgw Application to be promoted.
+        :ptype app_name: str
+        """
         if app_name is self.primary_rgw_app:
             new_secondary = self.secondary_rgw_unit
         else:
@@ -769,7 +791,12 @@ class CephRGWTest(test_utils.BaseCharmTest):
         ).get('Stdout', '')
 
     def get_client_keys(self, rgw_app_name=None):
-        """Create access_key and secret_key for boto3 client"""
+        """
+        Create access_key and secret_key for boto3 client.
+
+        :param rgw_app_name: RGW application for which keys are required.
+        :ptype rgw_app_name: str
+        """
         unit_name = self.primary_rgw_unit
         if rgw_app_name is not None:
             unit_name = rgw_app_name + '/0'
@@ -800,8 +827,13 @@ class CephRGWTest(test_utils.BaseCharmTest):
         wait=tenacity.wait_fixed(10),
         stop=tenacity.stop_after_attempt(5)
     )
-    def get_rgw_endpoint(self, unit_name:str, isHttps=False):
-        """Fetch Application endpoint for RGW unit"""
+    def get_rgw_endpoint(self, unit_name: str):
+        """
+        Fetch Application endpoint for RGW unit.
+
+        :param unit_name: Unit name for which RGW endpoint is required.
+        :ptype unit_name: str
+        """
         unit = zaza_model.get_unit_from_name(unit_name)
         unit_address = zaza_model.get_unit_public_address(
             unit,
@@ -811,14 +843,15 @@ class CephRGWTest(test_utils.BaseCharmTest):
         logging.debug("Unit: {}, Endpoint: {}".format(unit_name, unit_address))
         if unit_address is None:
             return None
-
-        if isHttps:
+        # Evaluate port
+        try:
+            zaza_model.get_application("vault")
             return "https://{}:443".format(unit_address)
-
-        return "http://{}:80".format(unit_address)
+        except KeyError:
+            return "http://{}:80".format(unit_address)
 
     def configure_rgw_apps_for_multisite(self):
-        """Configure Multisite values on primary and secondary apps"""
+        """Configure Multisite values on primary and secondary apps."""
         realm = 'zaza_realm'
         zonegroup = 'zaza_zg'
 
@@ -840,6 +873,12 @@ class CephRGWTest(test_utils.BaseCharmTest):
         )
 
     def clean_rgw_multisite_config(self, app_name):
+        """
+        Clear Multisite Juju config values to default.
+
+        :param app_name: App for which config values are to be cleared
+        :ptype app_name: str
+        """
         unit_name = app_name + "/0"
         zaza_model.set_application_config(
             app_name,
@@ -851,11 +890,10 @@ class CephRGWTest(test_utils.BaseCharmTest):
         )
         # Commit changes to period.
         cmd = self.get_rgwadmin_cmd_skeleton(unit_name)
-        output = zaza_model.run_on_unit(
+        zaza_model.run_on_unit(
             unit_name, cmd + 'period update --commit --rgw-zone=default '
             '--rgw-zonegroup=default'
         ).get('Stdout', '')
-
 
     def test_001_processes(self):
         """Verify Ceph processes.
@@ -897,9 +935,10 @@ class CephRGWTest(test_utils.BaseCharmTest):
 
     def test_003_secondary_migration_block(self):
         """
-        1. Verify object storage API with secondary endpoint.
-        2. Relate Primary-Secondary for multisite.
-        3. Verify secondary fails migration due to existing Bucket.
+        Performs the following checks (in order).
+
+        1. Object storage API works with secondary endpoint.
+        2. Secondary fails migration due to existing Bucket.
         """
         if not self.multisite:
             logging.info('Skipping Secondary Migration Block Test')
@@ -912,7 +951,7 @@ class CephRGWTest(test_utils.BaseCharmTest):
         secondary_endpoint = self.get_rgw_endpoint(self.secondary_rgw_unit)
         self.assertNotEqual(secondary_endpoint, None)
 
-        access_key, secret_key = self.get_client_keys(self.secondary_rgw_app) 
+        access_key, secret_key = self.get_client_keys(self.secondary_rgw_app)
         secondary_client = boto3.resource("s3",
                                           verify=False,
                                           endpoint_url=secondary_endpoint,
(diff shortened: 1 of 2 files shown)


zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -23,6 +23,8 @@
 )
 import requests
 import tempfile
+import boto3
```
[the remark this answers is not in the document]
UtkarshBhatthere: I went ahead with boto3 because I think it is already being used. The [requirements.txt](https://github.com/openstack-charmers/zaza-openstack-tests/blob/66d0c9237c792d286e713921a8d3ec298b15c53b/requirements.txt#L8) has it and it is used in Glance, Gnocchi, Swift and Trillio scripts (setup.py or tests.py). Can you please confirm once whether using boto3 would require dependency change as well ? Otherwise I will switch to using requests and crafting rest api calls.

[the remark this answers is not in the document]
ajkavanagh: Oh, I wonder how that snuck in!  Please could you do me a favor and also pin it as part of this PR?

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -695,23 +701,95 @@ def multisite(self):
...
             stdout = zaza_model.run_on_unit(unit_name, cmd).get('Stdout', '')
-            sync_states.append(sync_check_str in stdout)
+            sync_states.append((data_check in stdout) and
+                               (meta_check in stdout))
         assert all(sync_states)
 
-    def test_processes(self):
+    @tenacity.retry(wait=tenacity.wait_exponential(multiplier=1, max=60),
+                    reraise=True, stop=tenacity.stop_after_attempt(12))
+    def fetch_rgw_object(self, target_client, container_name, object_name):
```
lmlg, line 734: Looking at the code, it's not entirely clear to me what `target_client` is. I think some typing hints and/or documentation is needed.

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -695,23 +701,95 @@ def multisite(self):
...
+        return target_client.Object(container_name, object_name).get()[
+            'Body'
+        ].read().decode('UTF-8')
+
+    def promote_to_primary(self, app_name:str):
+        '''Promote provided app to Primary and update period at new secondary'''
+        new_secondary = 'ceph-radosgw/0' 
+        if app_name == 'ceph-radosgw':
+            new_secondary = 'slave-ceph-radosgw/0'
```
lmlg, line 743: If we're going to change the nomenclature at some point, I think we should store the prefix ('slave-' in this case) in a member variable and then produce the secondary name by concatenation.

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -755,132 +833,153 @@ def test_services(self):
...
+        primary_endpoint = self.get_rgw_endpoint('ceph-radosgw/0')
+        secondary_endpoint = self.get_rgw_endpoint('slave-ceph-radosgw/0')
+        access_key, secret_key = self.get_client_keys() 
+        primary_client = boto3.resource("s3",
+                                        verify=False,
+                                        endpoint_url=primary_endpoint,
+                                        aws_access_key_id=access_key,
+                                        aws_secret_access_key=secret_key)
+        secondary_client = boto3.resource("s3",
+                                       verify=False,
```
lmlg, line 947: Wrong indentation.

UtkarshBhatthere:
@ajkavanagh could you please approve the workflow for this PR as well ? It says "First-time contributors need a maintainer to approve running workflows."

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -695,23 +701,95 @@ def multisite(self):
...
+        )
+        output = json.loads(
+            zaza_model.run_on_unit(unit_name, create_cmd).get('Stdout', '')
+        )
+        keys = output['keys'][0]
+        return keys['access_key'], keys['secret_key']
+
+    def get_rgw_endpoint(self, unit_name:str):
+        unit = zaza_model.get_unit_from_name(unit_name)
+        unit_address = zaza_model.get_unit_public_address(
```
lmlg, line 785: This function can return `None` if the underlying provider hasn't given out an address yet. The tests use it at a fairly late point in time, but just for safety, I'd wrap this around a tenacity decorator to make sure it retries if the returned address is `None`.

lmlg (changes requested):

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: Added param description.

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: Have changed this naming logic to add prefix (currnetly 'slave') if not present, otherwise remove the prefix from app-name to toggle between "slave-$app_name" and "$app-name"

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -749,138 +895,235 @@ def test_services(self):
...
 
         logging.info('Checking multisite failover/failback')
-        keystone_session = zaza_openstack.get_overcloud_keystone_session()
-        source_client = zaza_openstack.get_swift_session_client(
-            keystone_session,
-            region_name='east-1',
-            cacert=self.cacert,
+        # Create RGW IO client.
+        primary_endpoint = self.get_rgw_endpoint(
+            self.primary_rgw_app + "/0"
```
UtkarshBhatthere, line 1079: Change to
```
self.primary_rgw_unit
```

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -749,138 +895,235 @@ def test_services(self):
...
         )
-        target_client = zaza_openstack.get_swift_session_client(
-            keystone_session,
-            region_name='west-1',
-            cacert=self.cacert,
+        secondary_endpoint = self.get_rgw_endpoint(
+            self.secondary_rgw_app + "/0"
```
UtkarshBhatthere, line 1082: Change to
```
self.primary_rgw_unit
```

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -749,138 +895,235 @@ def test_services(self):
...
+            self.configure_rgw_apps_for_multisite()
+            zaza_model.add_relation(
+                self.primary_rgw_app,
+                self.primary_rgw_app + ":master",
+                self.secondary_rgw_app + ":slave"
+            )
+            zaza_model.block_until_unit_wl_status(
+                self.secondary_rgw_app + "/0", "waiting"
+            )
+        logging.info('Waiting for model to stabalize')
```
UtkarshBhatthere, line 1028: Remove

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -37,6 +39,10 @@
 import zaza.openstack.utilities.openstack as zaza_openstack
 import zaza.openstack.utilities.generic as generic_utils
 
+# Disable warnings for ssl_verify=false
+urllib3.disable_warnings(
+    urllib3.exceptions.InsecureRequestWarning
```
ChrisMacNaughton, line 44: We have the CA in use when we make requests, can we not setup the urllib3 requests in such a way that they successfully verify with the test's CA?

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -690,28 +702,162 @@ def expected_apps(self):
...
+                               (meta_check in stdout))
         assert all(sync_states)
 
-    def test_processes(self):
+    @tenacity.retry(wait=tenacity.wait_exponential(multiplier=1, max=60),
+                    reraise=True, stop=tenacity.stop_after_attempt(12))
+    def fetch_rgw_object(self, target_client, container_name, object_name):
+        '''
+        Fetch RGW object content
+        :param target_client: boto3 client object configured for an endpoint.
```
ChrisMacNaughton, line 743: Can you include type information as well as descriptions about what the params are? I'm happy with either python typing or adding it via docstring formatted bits ([an example](https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/utilities/ceph.py#L22-L23)), but we have, so far, kind of standardised on the docstrings

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -690,28 +702,162 @@ def expected_apps(self):
...
+        unit_hostnames = generic_utils.get_unit_hostnames(juju_units)
+        hostname = unit_hostnames[unit_name]
+        return 'radosgw-admin --id=rgw.{} '.format(hostname)
+
     @tenacity.retry(wait=tenacity.wait_exponential(multiplier=10, max=300),
                     reraise=True, stop=tenacity.stop_after_attempt(10),
                     retry=tenacity.retry_if_exception_type(AssertionError))
-    def wait_for_sync(self, application):
+    def wait_for_sync(self, application, isPrimary=False):
         """Wait for slave to secondary to show it is in sync."""
```
ChrisMacNaughton, line 722: It looks like this function is more "Checking if the applications are in sync, rather than waiting for that state?

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -690,28 +702,162 @@ def expected_apps(self):
...
+        )
+
+        # Period Update Commit new secondary.
+        cmd = self.get_rgwadmin_cmd_skeleton(new_secondary)
+        zaza_model.run_on_unit(
+            new_secondary, cmd + 'period update --commit'
+        ).get('Stdout', '')
+
+    def get_client_keys(self, rgw_app_name=None):
+        """Create access_key and secret_key for boto3 client"""
```
ChrisMacNaughton, line 772: Please document the parameters here

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -690,28 +702,162 @@ def expected_apps(self):
...
+            action_params={},
+        )
+
+        # Period Update Commit new secondary.
+        cmd = self.get_rgwadmin_cmd_skeleton(new_secondary)
+        zaza_model.run_on_unit(
+            new_secondary, cmd + 'period update --commit'
+        ).get('Stdout', '')
+
+    def get_client_keys(self, rgw_app_name=None):
```
ChrisMacNaughton, line 771: This seems like it may be more generally useful; however, we haven't needed it yet, so maybe leave it here and we can move it later

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -690,28 +702,162 @@ def expected_apps(self):
...
+        keys = output['keys'][0]
+        return keys['access_key'], keys['secret_key']
+
+    @tenacity.retry(
+        retry=tenacity.retry_if_result(lambda ret: ret is None),
+        wait=tenacity.wait_fixed(10),
+        stop=tenacity.stop_after_attempt(5)
+    )
+    def get_rgw_endpoint(self, unit_name:str, isHttps=False):
+        """Fetch Application endpoint for RGW unit"""
```
ChrisMacNaughton, line 804: Please document the parameters

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -690,28 +702,162 @@ def expected_apps(self):
...
+        )
+        keys = output['keys'][0]
+        return keys['access_key'], keys['secret_key']
+
+    @tenacity.retry(
+        retry=tenacity.retry_if_result(lambda ret: ret is None),
+        wait=tenacity.wait_fixed(10),
+        stop=tenacity.stop_after_attempt(5)
+    )
+    def get_rgw_endpoint(self, unit_name:str, isHttps=False):
```
ChrisMacNaughton, line 803: We generally use snake_case rather than lowerCamelCase for variable names (ref: `isHttps`).

I also wonder if we can tell if TLS is enabled by checking for Vault inside this function

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -749,138 +895,231 @@ def test_services(self):
...
+        zaza_model.block_until_unit_wl_status(
+            self.secondary_rgw_unit, "active"
+        )
+        logging.info('Waiting for Data and Metadata to Synchronize')
+        self.wait_for_sync(self.secondary_rgw_app, isPrimary=False)
+        self.wait_for_sync(self.primary_rgw_app, isPrimary=True)
+
+        # Create a new session and Add one more object on primary.
+        logging.info('Performing post migration IO tests.')
+        primary_client = boto3.resource("s3",
```
ChrisMacNaughton, line 1038: Is there some reason we can't use the existing `primary_client`?

ChrisMacNaughton (changes requested):
Overall, I like the form of the tests, have some comments regarding styling, as well as some questions about why some approaches are being taken

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: ACK

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: Your observation is correct, by itself it only checks whether site is in sync, the last line assertion and the tenacity decorator make it wait for sync. Since it "effectively" waits and can't be used to "just check" the sync status, I kept the original name of the method.

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: Boolean parameters look a little strange in snake case (isHttps vs is_https). However the better solution is to just check for vault and decide on endpoint, I have shifted to that approach.

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: Sometimes when there is a significant delay between subsequent requests it appears as though some internal timeout occurs and existing boto3 resource objects (Bucket instances and RGW Object Instances) can't be reused (have to be re-instantiated). It appears that doesnt happen for client instance, I have removed this for now. If it breaks in further testing I will place it back again.

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: I guess we most certainly can. I was personally opinionated against going that extra mile because it's a test deployment and hence i simply suppressed those warnings. What's your take ? Should we ?

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: Added docstring formatted info for all newly added methods.

lmlg (approved):
Overall, I'm OK with the test. Regarding certificates, I think it may be a bit overkill for this kind of tests, but I can understand why we could need them as well.

ChrisMacNaughton:
Overall, I like it; with one nit that the code base prefers `type` over `ptype` for documenting parameter types: [random exmaple](https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/utilities/ceph.py#L23).

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -749,138 +933,229 @@ def test_services(self):
...
+            zaza_model.block_until_unit_wl_status(self.primary_rgw_unit,
+                                                  "active")
+        # 3. Verify secondary fails migration due to existing Bucket.
+        logging.info('Checking Secondary Migration Block')
+        assert_state = {
+            self.secondary_rgw_app: {
+                "workload-status": "blocked",
+                "workload-status-message-prefix":
+                    "Site contain Buckets, kindly purge all Buckets before "
+                    "configuring this site as a secondary."
```
UtkarshBhatthere, line 995: Modify this accordingly with the new gerrit changes.

lmlg (changes requested):

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -690,28 +701,201 @@ def expected_apps(self):
...
+            app_name,
+            'promote',
+            action_params={},
+        )
+
+        # Period Update Commit new secondary.
+        cmd = self.get_rgwadmin_cmd_skeleton(new_secondary)
+        zaza_model.run_on_unit(
+            new_secondary, cmd + 'period update --commit'
+        ).get('Stdout', '')
```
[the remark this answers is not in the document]
UtkarshBhatthere: It's definitely a leftover from when I was debugging :). Will clear them out.

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -690,28 +701,201 @@ def expected_apps(self):
...
+        if unit_address is None:
+            return None
+        # Evaluate port
+        try:
+            zaza_model.get_application("vault")
+            return "https://{}:443".format(unit_address)
+        except KeyError:
+            return "http://{}:80".format(unit_address)
+
+    def configure_rgw_apps_for_multisite(self):
```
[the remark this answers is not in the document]
UtkarshBhatthere: Well, The app will settle once a. Config is done (being set by this method), b. relation is added, c. The migration magic is done. Since this is only a single piece from the total process (just configures the multisite values), we can't wait "just yet" for the app to settle.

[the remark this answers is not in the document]
lmlg: That's not what I meant. The function may return in terms of the sync wrapper that is `set_application_config`, but that doesn't mean that the configuration has been actually applied to the model. For that reason, I recommend inserting calls to `wait_for_application_states` whenever an "important" part of the configuration is modified, or when a long-running action is executed.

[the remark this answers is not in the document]
UtkarshBhatthere: Oh okay, I see what you meant.
We use this method twice and in both scenarios the overall expectation to wait for is different:
1. In the secondary block scenario we expect the secondary app to go to blocked state with certain message.
2. In Normal migration scenario we expect it to block and then subsequently be ready.

For this mix/match of expectations I have kept the respective waiting logic in the actual test. Usage in test_004 and test_005.

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -690,28 +701,201 @@ def expected_apps(self):
...
+        meta_primary = 'metadata sync no sync (zone is master)'
+        meta_secondary = 'metadata is caught up with master'
+        meta_check = meta_primary if is_primary else meta_secondary
         for unit_name, hostname in unit_hostnames.items():
             key_name = "rgw.{}".format(hostname)
-            cmd = 'radosgw-admin --id={} zone sync status'.format(key_name)
+            cmd = 'radosgw-admin --id={} sync status'.format(key_name)
             stdout = zaza_model.run_on_unit(unit_name, cmd).get('Stdout', '')
-            sync_states.append(sync_check_str in stdout)
+            sync_states.append((data_check in stdout) and
```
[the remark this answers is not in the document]
ajkavanagh: I'd actually do something slightly different, but that is in the spirit of the original change.

If you do something like:

```python
failures = []
...
for unit_name, hosthame in unit_hostnames.items():
    ...
    for check in (data_check, meta_check):
        if check not in stdout:
            failures.append("unit: {}, hostname: {}: missing '{}' in sync status".format(unit_name, hostname, check))

self.assertTrue(len(failures) == 0, "sync failures: {}".format(", ".join(failures)))
```

Now you get to check all of them, see which failed, etc.

ajkavanagh (commented):
Just as an observation, are any of the tests named "test_xxx" idempotent?  They appear to leave quite a bit of state around.  e.g. test_003 seems to add some objects but not clean up afterwards.  test_004 does a migration, which clearly changes state in the underlying system.

Generally, the test class as a whole appears to leave a lot of state behind.  Also, are the tests themselves fully isolated. e.g. can you just run test_005.. and not run test_004...?  Can you run those tests in any order?

UtkarshBhatthere:
@ajkavanagh In an overall sense the tests are NOT idempotent. We had a small discussion around this initially, the problem is We can't (yet) scale down to single site once we have migrated to a multisite system (so that's a non-reversible change). Migration is tested in 005 hence 001 to 004 in themselves can be executed again but once 005 has passed through 004 will never pass (in fact it will timeout waiting for a blocked state that will never arrive). We can make it so that it skips if migration has taken place but since tests are ordered I am not sure if we would like to go that way.

In terms of inter test dependence:
1. 006 needs 005 since we cant test multisite failover unless we have a multisite system at hand.
2. 004 needs 003 since the basic expectation is that migration will halt if there is pre-existing data on secondary site.

One way to reduce this dependency is to merge the tests and have bigger long-running test-cases however the overall test-suite "CephRGWTests" will still remain non-idempotent.

ajkavanagh:
> @ajkavanagh In an overall sense the tests are NOT idempotent. We had a small discussion around this initially, the problem is We can't (yet) scale down to single site once we have migrated to a multisite system (so that's a non-reversible change). Migration is tested in 005 hence 001 to 004 in themselves can be executed again but once 005 has passed through 004 will never pass (in fact it will timeout waiting for a blocked state that will never arrive). We can make it so that it skips if migration has taken place but since tests are ordered I am not sure if we would like to go that way.

Yes, that's awkward.  Partly the problem of using a unittest framework for functional/integration/system testing.

However, we can help ourselves by ensuring that the *input* to a test is only that done by the module/class/test setUp functions.

> 
> In terms of inter test dependence:
> 
> 1. 006 needs 005 since we cant test multisite failover unless we have a multisite system at hand.
> 2. 004 needs 003 since the basic expectation is that migration will halt if there is pre-existing data on secondary site.

It really sounds like 5 & 6 are essentially a continuation, and thus are one test, and the same for 3 and 4.

> 
> One way to reduce this dependency is to merge the tests and have bigger long-running test-cases however the overall test-suite "CephRGWTests" will still remain non-idempotent.

Bigger, sequenced, individual tests are not necessarily a problem; the openstack upgrade test is 'one' test and upgrades the entire openstack system!

I do think it would be helpful to 'mark' the tests that are not idempotent for future travellers, and also the test class as a whole.  Ideally, we'd use a decorator for this, but failing that, at least a comment in the relevant docstrings would be good.

UtkarshBhatthere:
Got it, I'll merge 003-004 and 005-006. Also adding relevant documentation for Future Travellers :D
Will look for example usage of the idempotent decorator and follow that.

lmlg (approved):
Looking at it again, I feel like this PR is in good shape. LGTM.

ajkavanagh (approved):

ChrisMacNaughton (approved):

PR #849, opened 2022-07-21
Move tests that aren't in stable to a different class

lmlg:
In order to correctly test the upgrade from stable, we have to move the tests that haven't landed yet to a different class so
that the charm can select to run them only after performing the upgrade to the locally built charm.

EDIT: This PR is being tested here: https://review.opendev.org/c/openstack/charm-ceph-mon/+/841498


ChrisMacNaughton:
I would actually like to see an independent change to the ceph-mon charm to update _only_ the calling of this test

lmlg:
> I would actually like to see an independent change to the ceph-mon charm to update _only_ the calling of this test

Such a thing can be seen here: https://review.opendev.org/c/openstack/charm-ceph-mon/+/851267

PR #851, opened 2022-08-03
Fix pep8 lint errors

swalladge:
One of the pep8 target dependencies must have updated,
causing a bunch of new lint errors in these categories:
- line length > 79 chars
- no whitespace after keyword


fnordahl (approved):
LGTM, thanks!

PR #853, opened 2022-08-05
Add glance-simplestreams-sync --set-latest-property functest

guimaluf:
When creating Openstack VMs the user has to specify the image it wants
to use. sstream-mirror-glance adds a date to the image name, so they
always have to recheck which is the current latest image.

This commit tests the usage of the `set-latest-property` configuration.

When set-latest-property is given sstream-mirror-glance will set the
recently synced image with the `latest=true` property and then remove
the `latest` property from all the os_version/architecture matching
images.

Closes-bug: LP #1933130

### DIFF
--- a/zaza/openstack/charm_tests/glance_simplestreams_sync/setup.py
+++ b/zaza/openstack/charm_tests/glance_simplestreams_sync/setup.py
@@ -21,6 +21,7 @@ import tenacity
 import pprint
 
 import zaza.model as zaza_model
+import zaza.charm_lifecycle.utils as lifecycle_utils
 import zaza.openstack.utilities.generic as generic_utils
 import zaza.openstack.utilities.openstack as openstack_utils
 
@@ -77,3 +78,17 @@ def sync_images():
         logging.info('Contents of Keystone service catalog: "{}"'
                      .format(pprint.pformat(catalog)))
         raise
+
+
+def set_latest_property_config():
+    """Enable set_latest_property config.
+
+    This config adds `latest=true` to new synced images.
+    """
+    logging.info("Change config `set_latest_property=true`")
+    zaza_model.set_application_config('glance-simplestreams-sync',
+                                      {'set_latest_property': 'true',
+                                       'snap-channel': 'edge'})
+    test_config = lifecycle_utils.get_charm_config(fatal=False)
+    zaza_model.wait_for_application_states(
+        states=test_config.get('target_deploy_status', {}))
--- a/zaza/openstack/charm_tests/glance_simplestreams_sync/tests.py
+++ b/zaza/openstack/charm_tests/glance_simplestreams_sync/tests.py
@@ -125,3 +125,23 @@ class GlanceSimpleStreamsSyncTest(test_utils.OpenStackBaseTest):
         _check_local_product_streams(expected_images)
 
         logging.debug("Local product stream successful")
+
+
+class GlanceSimpleStreamsSyncWithPropertiesTest(GlanceSimpleStreamsSyncTest):
+    """Glance Simple Streams Sync Test with Image property.
+
+    `setup.py:set_latest_property_config()` is required by this test and it is
+    called during charm-glance-simplestreams-sync/tests/tests.yaml:configure
+    phase.
+    """
+
+    # TODO(guimalufb) test if the latest property gets removed from old images
+    def test_200_check_image_latest_property(self):
+        """Verify that images had metadata property set."""
+        logging.debug("Checking images with latest=true property...")
+
+        filter_properties = {'filters': {'latest': 'true'}}
+        images = self.glance_client.images.list(**filter_properties)
+        self.assertTrue(len(list(images)) > 0,
+                        "'latest=true' property not found in glance images"
+                        " list")


zaza/openstack/charm_tests/glance_simplestreams_sync/tests.py
```diff
@@ -125,3 +125,21 @@ def _check_local_product_streams(expected_images):
...
+        """Run class setup for running glance simple streams sync tests."""
+        super(GlanceSimpleStreamsSyncWithPropertiesTest, cls).setUpClass()
+        zaza_model.set_application_config('glance-simplestreams-sync', {'set_latest_property': 'true'})
+
+    # TODO(guimalufb) test if the latest property gets removed from old images
+    def test_200_check_image_latest_property(self):
+        """Verify that a property inclusion has been done"""
+        logging.debug("Checking image latest property")
+
+        _filter_properties = {'filters': {'latest': 'true'}}
```
ajkavanagh, line 143: Not sure why you need an underscore variable here; it's already 'private' to the function?

zaza/openstack/charm_tests/glance_simplestreams_sync/tests.py
```diff
@@ -125,3 +125,21 @@ def _check_local_product_streams(expected_images):
...
+        images = self.glance_client.images.list(**_filter_properties)
+        self.assertTrue(len(list(images)) > 0)
```
ajkavanagh, line 145: The `len()` function already takes any object that has an `__iter__()` method, so there isn't a need to convert it to a list first.

It's also quite nice to add a message to the `self.assertTrue()` to indicate what the error is in the output if the assertion fails.  e.g. something like "'latest' property not found in glance images list".

zaza/openstack/charm_tests/glance_simplestreams_sync/tests.py
```diff
@@ -125,3 +125,21 @@ def _check_local_product_streams(expected_images):
...
         logging.debug("Local product stream successful")
+
+
+class GlanceSimpleStreamsSyncWithPropertiesTest(GlanceSimpleStreamsSyncTest):
+
+    @classmethod
+    def setUpClass(cls):
+        """Run class setup for running glance simple streams sync tests."""
+        super(GlanceSimpleStreamsSyncWithPropertiesTest, cls).setUpClass()
+        zaza_model.set_application_config('glance-simplestreams-sync', {'set_latest_property': 'true'})
```
ajkavanagh, line 136: Why is this part of a class set up?  Wouldn't this line be better in the test.  Plus, I think you'll need to wait for the application to settle as this function is asynchronous.  i.e. it returns before the `config-changed` hook has completed (or even started) on the model. e.g. you'll need something like:

```python
    zaza_model.set_application_config(...)
    zaza_model.wait_for_application_states(states=states)
```

You'll need to work out what `states` is above for this particular test.  Take a look at other usages of `set_application_config` in test code (not in the utilities) to see what they are doing.

ajkavanagh (changes requested):
Thanks for the patch; it looks like it is starting to get there.  A couple of minor nitpicks, but also a significant issue with the async nature of the models and the code driving them.  e.g. config changed -> "config-changed" hook, but that runs asynchronously to the code that starts it.  Please take a look at the comments.  Thanks.

zaza/openstack/charm_tests/glance_simplestreams_sync/tests.py
guimaluf: thanks @ajkavanagh for the quick review! :) 

[`glance_client.images.list` returns a generator](https://docs.openstack.org/python-glanceclient/latest/reference/api/glanceclient.v2.images.html#glanceclient.v2.images.Controller.list), that's why I need to convert it to a list

```python
(Pdb) len(images)
*** TypeError: object of type 'generator' has no len()
```

zaza/openstack/charm_tests/glance_simplestreams_sync/tests.py
guimaluf: I've use it in the test setup because I don't know how to tear down the images and rerun sync-images actions, or to force glance to sync the images on top of the existing one. 

if the images have been synced, already exist in glance and  we call sync-image action nothing will change as
 the `latest=true` property is attached to new images only. 

that's why I wrote a separated class for the test instead of a new test case. I also think the latter is better.

zaza/openstack/charm_tests/glance_simplestreams_sync/tests.py
ajkavanagh: Ah, that's right; generators don't support len.  Tedious.  As it's probably small, `len(list())` is okay.  I'd use `sum(1 for _ in images)` if the generator could produce a lot of items.

zaza/openstack/charm_tests/glance_simplestreams_sync/tests.py
ajkavanagh: Sorry, I wasn't clear: there's two issues here:

1. The setup of the property in the `setUpClass()` method; I don't think it should be there; it's actually part of the test rather than the fixture.
2. (More important): handling the async nature of `set_application_config()`;

Having the test as a separate class is fine.

guimaluf: so, the reason why the setup of the property is in the `setUpClass()` is due the fact I need that config to be done BEFORE the image is actually sync. 
If we set the config after the images already exists in glance and call the sync-images action, nothing will be done. the existing image won't get the property.

ajkavanagh: > so, the reason why the setup of the property is in the `setUpClass()` is due the fact I need that config to be done BEFORE the image is actually sync. If we set the config after the images already exists in glance and call the sync-images action, nothing will be done. the existing image won't get the property.

That doesn't make sense to me.  Image syncing is kicked off by `../setup.py:sync_images()`  How are you ensuring that the class is setup before that function is called?

guimaluf: hmm. you're right. I'm not ensuring the config is set before the sync. actually the `setup.py:sync_images()` is called in the `functest-configure` phase.

so, should I 

1. set `set_latest_property=true` in the  `functest-configure` phase, and test if there is an image with latest property?
2. delete any existing image, `set_latest_property=true`, resync, and test if there is an image with latest property?
3. something else which I'm not seeing?

Thanks for pointing that out!

guimaluf: I've done option 1. 

what you've said makes way more sense now :) 

thanks

zaza/openstack/charm_tests/glance_simplestreams_sync/tests.py
```diff
@@ -125,3 +125,24 @@ def _check_local_product_streams(expected_images):
...
+
+
+class GlanceSimpleStreamsSyncWithPropertiesTest(GlanceSimpleStreamsSyncTest):
+
+    @classmethod
+    def setUpClass(cls):
+        # `setup.py:set_latest_property_config()` is required by this test and
+        # it is called by
+        # charm-glance-simplestreams-sync/tests/tests.yaml:configure phase
+        super(GlanceSimpleStreamsSyncWithPropertiesTest, cls).setUpClass()
```
ajkavanagh, line 137: Sorry, to be a bit pedantic, but this override on `setUpClass` isn't needed any more.  The comment re: set_latest_property_config() is useful, and should be retained in the class's docstring.

zaza/openstack/charm_tests/glance_simplestreams_sync/tests.py
```diff
@@ -125,3 +125,24 @@ def _check_local_product_streams(expected_images):
...
+    # TODO(guimalufb) test if the latest property gets removed from old images
+    def test_200_check_image_latest_property(self):
+        """Verify that images had metadata property set"""
+        logging.debug("Checking images with latest=true property...")
+
+        filter_properties = {'filters': {'latest': 'true'}}
+        images = self.glance_client.images.list(**filter_properties)
+        self.assertTrue(len(list(images)) > 0,
+                        "'latest=true' property not found in glance images"
+                        " list")
```
ajkavanagh, line 148: This test assumes that all the images synced will have the latest property.  If all the images have the latest property, which image is the latest synced?

ajkavanagh (changes requested):
So I've taken a step back, and instead of just concentrating on the code in this review, I looked at the associated bug.  I'm pretty sure that the code that this might be testing doesn't do what the bug reporter wants: a named alias for the 'latest' version of a particular synced image type.

This code appears to be testing 'adding a property to every image and calling that latest' which doesn't seem that useful to be honest.

I think this solution should be revisited - perhaps email the opendev mailing list, or contact the team via a channel, and see if we can move this forward.  Thanks.

ajkavanagh:
> I've discussed with @wolsen and @freyes about this topic. it was suggested that instead of changing the name of the image I should add the `latest` attribute.

Okay, at the very least, please go back to the bug and outline the solution so that the bug submitter can take a look at agree/disagree that it solves their problem.  Otherwise, whilst it may be a solution, it simply may not work for the bug submitter and their tools.

> 
> I'm probably not doing in the best way, but if you look at [my proposal for simplestreams](https://git.launchpad.net/~guimalufb/simplestreams/diff/?h=set_latest_property_glance_image) I'm removing the `latest` property from all other images while letting only the last one synced.

That's why I couldn't find the change.  This change needs to be on gerrit (review.opendev.org), NOT on launchpad.  If you take a look at the code page (https://code.launchpad.net/charm-glance-simplestreams-sync) you'll see that it syncs the code from https://opendev.org/openstack/charm-glance-simplestreams-sync.git - any changes you attempt to make on LP will a) not be merged, and b) if they were, overwritten by a sync.  Please re-do your PR on opendev.


> 
> ```diff
> diff --git a/simplestreams/mirrors/glance.py b/simplestreams/mirrors/glance.py
> index 9181e9c..b96f8eb 100644
> --- a/simplestreams/mirrors/glance.py
> +++ b/simplestreams/mirrors/glance.py
> @@ -517,6 +523,20 @@ class GlanceMirror(mirrors.BasicMirrorWriter):
>  
>              print("created %s: %s" % (glance_image.id, full_image_name))
>  
> +            # TODO(guimalufb) use self.loaded_content or
> +            # self.load_products() instead
> +            if self.set_latest_property:
> +                # Search all images with the same target attributtes
> +                _filter_properties = {'filters': {
> +                    'latest': 'true',
> +                    'os_version': glance_props['os_version'],
> +                    'architecture': glance_props['architecture']}}
> +                images = self.gclient.images.list(**_filter_properties)
> +                for image in images:
> +                    if image.id != glance_image.id:
> +                        self.gclient.images.update(image.id,
> +                                                   remove_props=['latest'])
> +
> ```
> 
> I'd like to test having an image with the `latest` and then a new image on top. but I have no idea how to force that scenario.

Yes, that's tricky from a functional/integration test scenario.  Best approach would be to verify that functionality using unit tests and mocking out simplestreams (or even in the g-sss (glance-simplestreams-sync) charm itself).

> 
> thanks for you careful review!

No problem. :)

guimaluf:
> Okay, at the very least, please go back to the bug and outline the solution so that the bug submitter can take a look at agree/disagree that it solves their problem. Otherwise, whilst it may be a solution, it simply may not work for the bug submitter and their tools.

done!

> That's why I couldn't find the change. This change needs to be on gerrit (review.opendev.org), NOT on launchpad. If you take a look at the code page (https://code.launchpad.net/charm-glance-simplestreams-sync) you'll see that it syncs the code from https://opendev.org/openstack/charm-glance-simplestreams-sync.git - any changes you attempt to make on LP will a) not be merged, and b) if they were, overwritten by a sync. Please re-do your PR on opendev.

my bad. I ran `git review` for the simplestreams repository and it didn't work, thus I assumed it would work for g-s-s.
[done! ](https://review.opendev.org/c/openstack/charm-glance-simplestreams-sync/+/852445)

> Yes, that's tricky from a functional/integration test scenario. Best approach would be to verify that functionality using unit tests and mocking out simplestreams (or even in the g-sss (glance-simplestreams-sync) charm itself).

That has been [done on simplestreams repository](https://code.launchpad.net/~guimalufb/simplestreams/+git/simplestreams/+merge/427903)

zaza/openstack/charm_tests/glance_simplestreams_sync/setup.py
```diff
@@ -77,3 +78,15 @@ def sync_images():
         logging.info('Contents of Keystone service catalog: "{}"'
                      .format(pprint.pformat(catalog)))
         raise
+
+
+def set_latest_property_config():
```
wolsen, line 83: Any particular reason this is a setup method rather than just an option in one of the test bundles?

guimaluf: my change doesn't add the `latest=true` property to existing images, just for the newly-created.
also, I couldn't find a way to force a new sync of images on top of the existing one, and deleting the image and re-sync seemed too extreme.
thus I need to set the config before the sync actually happen. 

@ajkavanagh asked the same thing [here in this collapsed comment](https://github.com/openstack-charmers/zaza-openstack-tests/pull/853#discussion_r938714958)

guimaluf: @wolsen and @ajkavanagh 
my change has been [merged on simplestreams repo](https://git.launchpad.net/simplestreams/commit/?id=bc0d94320bb6ce56c7ec0171bcf087f4eb921e87). how can I reference this commit to the CI tests?

ChrisMacNaughton:
Landing this change but only after raising https://github.com/openstack-charmers/zaza-openstack-tests/issues/947 to resolve the strangely placed config setting

PR #850, opened 2022-07-28
Revert "Add test for online extending lvm volume in cinder"

ajkavanagh:
Reverts openstack-charmers/zaza-openstack-tests#776

Reverting due to it breaking the cinder-ceph charm's gate with an assert:

```
2022-07-25 10:43:04.607769 | focal-medium | 2022-07-25 10:43:04 [INFO] ======================================================================
2022-07-25 10:43:04.607974 | focal-medium | 2022-07-25 10:43:04 [INFO] FAIL: test_200_online_extend_volume_lvm_backend (zaza.openstack.charm_tests.cinder.tests.CinderTests)
2022-07-25 10:43:04.608035 | focal-medium | 2022-07-25 10:43:04 [INFO] Test extending an attached volume with lvm backend.
2022-07-25 10:43:04.608128 | focal-medium | 2022-07-25 10:43:04 [INFO] ----------------------------------------------------------------------
2022-07-25 10:43:04.608289 | focal-medium | 2022-07-25 10:43:04 [INFO] Traceback (most recent call last):
2022-07-25 10:43:04.608351 | focal-medium | 2022-07-25 10:43:04 [INFO]   File "/home/ubuntu/src/review.opendev.org/openstack/charm-cinder-ceph/.tox/func-target/lib/python3.8/site-packages/zaza/openstack/charm_tests/cinder/tests.py", line 263, in test_200_online_extend_volume_lvm_backend
2022-07-25 10:43:04.608443 | focal-medium | 2022-07-25 10:43:04 [INFO]     self.assertIn('@LVM', host)
2022-07-25 10:43:04.608469 | focal-medium | 2022-07-25 10:43:04 [INFO] AssertionError: '@LVM' not found in 'cinder@cinder-ceph'
```

This original PR will need to re-submitted and all the affected charms that include the `test_200_online_extend_volume_lvm_backend (zaza.openstack.charm_tests.cinder.tests.CinderTests)` test will need to use a func-test-pr to verify that the change is consistent across all the charms.

PR #854, opened 2022-08-10
Add test for online extending lvm volume in cinder

swalladge:
This was originally added in https://github.com/openstack-charmers/zaza-openstack-tests/pull/776 , but then reverted in https://github.com/openstack-charmers/zaza-openstack-tests/pull/850 since it broke cinder's gate tests.

This is an attempt to fix and re-add the test. :)

Func-test-pr gerrit patches to verify the tests pass:

- charm-cinder: https://review.opendev.org/c/openstack/charm-cinder/+/852697
- charm-cinder-ceph: https://review.opendev.org/c/openstack/charm-cinder-ceph/+/852698

### DIFF
--- a/zaza/openstack/charm_tests/cinder/tests.py
+++ b/zaza/openstack/charm_tests/cinder/tests.py
@@ -17,11 +17,15 @@
 """Encapsulate Cinder testing."""
 
 import logging
+import time
 
 import zaza.model
 import zaza.openstack.charm_tests.test_utils as test_utils
 import zaza.openstack.utilities.openstack as openstack_utils
 import zaza.openstack.charm_tests.glance.setup as glance_setup
+import zaza.openstack.charm_tests.neutron.tests as neutron_tests
+import zaza.openstack.configure.guest as guest
+import zaza.openstack.charm_tests.nova.utils as nova_utils
 
 from tenacity import (
     Retrying,
@@ -42,8 +46,9 @@ class CinderTests(test_utils.OpenStackBaseTest):
         cls.application_name = 'cinder'
         cls.lead_unit = zaza.model.get_lead_unit_name(
             "cinder", model_name=cls.model_name)
+        # version 3.42 is required for in-use (online) resizing lvm volumes
         cls.cinder_client = openstack_utils.get_cinder_session_client(
-            cls.keystone_session)
+            cls.keystone_session, version=3.42)
         cls.nova_client = openstack_utils.get_nova_session_client(
             cls.keystone_session)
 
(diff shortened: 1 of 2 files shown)


nobuto-m:
> Should this resize work here? Otherwise we can enable the test only for cinder.


Yes, I'm pretty sure the live-resizing worked with Ceph RBD volume in our testing. I've shared my notes separately with you @swalladge.


```
Jun 22 08:33:57 my-instance1 kernel: virtio_blk virtio1: [vda] new size: 62914560 512-byte logical blocks (32.2 GB/30.0 GiB)
Jun 22 08:33:57 my-instance1 kernel: vda: detected capacity change from 10737418240 to 32212254720
```

nobuto-m:
Okay, so the root cause of the functional test failure was:

[cinder-volume.log w/ charm-cinder-ceph]
```
2022-08-15 01:39:33.260 36417 ERROR cinder.compute.nova [req-7fad0978-badf-4b76-8dc0-549db525912b 9ee53f0a0f064a0fbbd8bd9b94879af6 ce58a706707f4e8c8ac4406fafaaf162 - - -] Failed to notify nova on events: [{'
name': 'volume-extended', 'server_uuid': '382457e9-e86b-46a3-94e3-5a5f65b68a0e', 'tag': '1cd8ef37-ae8e-45ca-acaf-4dba590aac22'}].: keystoneauth1.exceptions.http.BadRequest: Expecting to find domain in projec
t. The server could not comply with the request since it is either malformed or otherwise incorrect. The client is assumed to be in error. (HTTP 400) (Request-ID: req-94a8133e-13f6-446a-a30d-c86b6ad38cca)
```

Which was fixed in https://bugs.launchpad.net/charm-cinder/+bug/1939389

However, the master branch of each charm has functional tests with `yoga/edge` instead of `latest/edge` so without backporting the fix above *without* functional tests is required, which is a bit weird.

nobuto-m:
https://github.com/openstack/charm-cinder-ceph/blob/master/tests/bundles/jammy-yoga.yaml#L159-L168

nobuto-m:
The situation is trickier than expected. `cinder-ceph`'s functional tests in the master branch rely on [both yoga and xena jobs](https://opendev.org/openstack/charm-cinder-ceph/src/commit/a8cfbe21f768cd6fe14f2a86999b6949a589afac/osci.yaml) so we cannot enable this functional test until we backport the fix for [LP: #1939389](https://bugs.launchpad.net/charm-cinder/+bug/1939389) is merged into yoga and xena branches.

However, this functional test is exactly for validating the fix for [LP: #1939389](https://bugs.launchpad.net/charm-cinder/+bug/1939389)... In this case, should we backport the fix without running functional tests as an exception?

Would you mind chiming in? @ajkavanagh

ajkavanagh:
@nobuto-m @swalladge There appear to be a bunch of problems with the master branch of cinder-ceph which need fixing to help resolve this:

* The zed batch enablement will remove the xena problem: https://review.opendev.org/c/openstack/charm-cinder-ceph/+/850177
* The test bundles shouldn't refer to anything other than latest/edge (except for the mysql, hacluster, etc). <- this appears to be wrong in the zed enablement bundles too; so I'll look into this.

It shouldn't be necessary to backport this PR to yoga, etc., as the feature is only landing in the master branch for 'zed' at present (AFAICT).  We'll need to get the zed batch enablement patches through for cinder and cinder-ceph before the patches for cinder-ceph and cinder for this feature can land:

* cinder: https://review.opendev.org/c/openstack/charm-cinder/+/850174
* cinder-ceph: https://review.opendev.org/c/openstack/charm-cinder-ceph/+/850177

ajkavanagh:
> This isn't a feature. You could extend a volume prior to Victoria, you can't with Victoria onwards. While the means to achieve this was changed, its clearly a changed (broken) behavior without this fix. Including the functional test should be necessary; and we can think of it like the test is the regression test that proves its both broken and fixed.

Well, indeed, that's a regression. This means the original patches for the fix need backporting to victoria and this patch needs porting back to stable/victoria.  That will then test bundles from ussuri (which should work) all the way to master.

nobuto-m:
> Func-test-pr gerrit patches to verify the tests pass:
> 
>     charm-cinder: https://review.opendev.org/c/openstack/charm-cinder/+/852697
>     charm-cinder-ceph: https://review.opendev.org/c/openstack/charm-cinder-ceph/+/852698

Those two empty review have confirmed that the new test in this PR successfully ran.

> 2022-08-31 06:16:18.780656 | focal-medium | 2022-08-31 06:16:18 [INFO] test_200_online_extend_volume (zaza.openstack.charm_tests.cinder.tests.CinderTests)

> 2022-08-31 06:15:01.784883 | focal-medium | 2022-08-31 06:15:01 [INFO] test_200_online_extend_volume (zaza.openstack.charm_tests.cinder.tests.CinderTests)

I think this PR can land to the main branch, and we can think about the backport separately.

ajkavanagh (approved):
LGTM

PR #855, opened 2022-08-11
Remove libjuju and juju_wait dependency in zaza-openstack-tests

ChrisMacNaughton:
This dependency is available via zaza already, and more
clearly illustrates the origin of dependencies.


wolsen (approved):
Yes, let's not declare dependencies that aren't really ours

PR #859, opened 2022-08-18
Add tests for mysql prometheus exporter

jneo8:
test(zaza/openstack/charm_tests/mysql/test_prometheus_mysql_exporter.py): Add tests for mysql prometheus exporter

Here is a review using this code: https://review.opendev.org/c/openstack/charm-mysql-innodb-cluster/+/852064

### DIFF
--- a/zaza/openstack/charm_tests/mysql/test_prometheus_mysql_exporter.py
+++ b/zaza/openstack/charm_tests/mysql/test_prometheus_mysql_exporter.py
@@ -1,0 +1,110 @@
+# Copyright 2022 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""MySQL Prometheus Exporter Testing."""
+
+import json
+
+import zaza.model as zaza_model
+from zaza.openstack.charm_tests.mysql.tests import MySQLBaseTest
+
+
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/mysql/test_prometheus_mysql_exporter.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def _exporter_http_check(
+        self,
+        cmd,
+        expected,
+    ):
+        """Exec check cmd on each unit in the application.
+
+        :param str cmd: The check command run on unit.
+        :param str expected: Expected result code.
```
ajkavanagh, line 40: Although we don't strictly enforce params in docstrings on zaza-opnestack-tests, if you do include a `:param:` it would be nice to also include the `:type:` for the param, please.

ajkavanagh:
Please could you add the review on gerrit that uses this PR and add it to the description above?  (i.e. a review with a `func-test-pr:` in the commit; this is to verify that the tests do actually work prior to adding them to the library.  Thanks.

zaza/openstack/charm_tests/mysql/test_prometheus_mysql_exporter.py
jneo8: Updated

zaza/openstack/charm_tests/mysql/test_prometheus_mysql_exporter.py
```diff
@@ -0,0 +1,114 @@
...
+        if not active:
+            excepted = "inactive\n"
+        for unit in zaza_model.get_units(self.application):
+            result = zaza_model.run_on_unit(unit.name, cmd)
+            self.assertEqual(result.get("stdout"), excepted)
+
+    def test_01_exporter_http_check(self):
+        """Check exporter endpoint is working."""
+        self._exporter_http_check(
+            cmd="curl http://localhost:9104",
```
wolsen, line 67: Is there a specific reason for logging in and running a curl from localhost here? I think there's value in just querying it remotely.

jneo8: I think we can add another check `mysql_up` = 1 here. It can make sure exporter service is alive also permission is correct.

PR #864, opened 2022-08-30
Migrate CephFS tests to not use OpenStack

ChrisMacNaughton:
This change updates the expectations of the test to expect
two ubuntu units deployed in the same model as the ceph-fs
units. These ubuntu units are used to mount and then verify
the CephFS shares.

### DIFF
--- a/zaza/openstack/charm_tests/ceph/fs/tests.py
+++ b/zaza/openstack/charm_tests/ceph/fs/tests.py
@@ -15,7 +15,7 @@
 """Encapsulate CephFS testing."""
 
 import logging
-import asyncio
+import json
 import subprocess
 from tenacity import retry, Retrying, stop_after_attempt, wait_exponential
 import unittest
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/ceph/fs/tests.py
```diff
@@ -16,42 +16,101 @@
...
+                zaza.utilities.generic.run_via_ssh(
+                    unit_name='ubuntu/0',
+                    cmd='sudo fusermount -u {0} && sudo rmdir {0}'.format(
+                        self.mount_dir))
+                zaza.utilities.generic.run_via_ssh(
+                    unit_name='ubuntu/1',
+                    cmd='sudo fusermount -u {0} && sudo rmdir {0}'.format(
+                        self.mount_dir))
+            except subprocess.CalledProcessError:
+                logging.warning("Failed to cleanup mounts")
```
wolsen, line 46: This won't even attempt to cleanup the mounts on the second unit if the first unit fails (which I think we should do). Additionally, we may want to augment the logging to indicate which unit?

wolsen (changes requested):

wolsen (approved):

PR #867, opened 2022-09-01
Fix race conditions in VaultFacade

swalladge:
LP: https://bugs.launchpad.net/vault-charm/+bug/1988273

This ensures initialising and unsealing works nicely
with vault using the raft backend
(which makes race conditions evident because of the time
it takes for the raft cluster to settle).

To test this, see the linked launchpad bug.


fnordahl (approved):
LGTM, thanks!

Have confirmed this works with both single- and multi-unit operation.

PR #882, opened 2022-09-06
Fix pep8 lint errors

gnuoy:
One of the pep8 target dependencies must have updated,
causing a bunch of new lint errors in these categories:
- line length > 79 chars
- no whitespace after keyword


hemanthnakkina (approved):
LGTM!

PR #887, opened 2022-09-08
Fix Ganesha _restart_share_instance for non-ha

gnuoy:
Manila Ganesha is not always deployed in an ha configuration in functional tests. But the _restart_share_instance method shuts down manila services via systemctl and expects them to be automatically started. This change detects if the hacluster charm is present, if it is it uses the old behaviour otherwise systemctl is used to restart the services.

Test run without ha manila ganesha:
https://openstack-ci-reports.ubuntu.com/artifacts/642/856330/7/check/ganesha-focal-xena/6421186/job-output.txt

Test run with ha manila ganesha:
https://openstack-ci-reports.ubuntu.com/artifacts/bff/856459/1/check/focal-xena/bffb34c/job-output.txt


freyes (approved):
LGTM.

PR #894, opened 2022-09-08
Ensure correct snapshot support

gnuoy:
A manila charm using the wallaby branch will be testing wallaby and backwards compatability with victoria so just for the wallaby branch of z-o-t the OpenStack release needs to be detected and snapshot set accordingly.


coreycb (approved):

freyes (approved):
master has snapshort support set to True since bf6b466b55e5a9602e5b75ba8e226cbe2dcb040e

PR #899, opened 2022-09-12
Fix dashboard policy tests.

gnuoy:
The dashboard policyd  override tests assume the deloyment is http and non-ha. This is not a safe assumption so use the existing code to generate the correct domain listing url.


gnuoy:
Tested with https ha and http non-ha functional tests

fnordahl (approved):

PR #908, opened 2022-09-13
Fix keystone saml mellon tests

gnuoy:
Fix keystone saml mellon tests

When calculating the region to pass when setting the form data
the test checks the wrong OpenStack combination. In stein onwards
'default' can be passed as the region id *1
 
*1 https://opendev.org/openstack/horizon/src/branch/stable/stein/openstack_auth/forms.py#L32


ajkavanagh (approved):
LGTM; landing as stable/ussuri track hasn't yet been released and so doesn't need two reviews.

PR #912, opened 2022-09-15
Fix code for non-ha keystone saml mellon deploys

gnuoy:
When deployed in a non-ha setup the keystone ip can fail to be found due to problems with getting ips of subordinates. Also, there is duplicate code for finding the keystone ip so put this in the setup class.


ajkavanagh (approved):
This is a good approach to nail down the keystone ip.

PR #915, opened 2022-09-15
Fix code for non-ha keystone saml mellon deploys (#912)

gnuoy:
When deployed in a non-ha setup the keystone ip can fail to be found due to problems with getting ips of subordinates. Also, there is duplicate code for finding the keystone ip so put this in the setup class.

(cherry picked from commit a65ef82ef56fcb126401548fde2484fda14f9d69)


ajkavanagh (approved):
LGTM

PR #922, opened 2022-09-16
Remove workaround for Octavia Bug #1964117

coreycb:
Remove the work around for Bug #1964117 now that it is fixed in the latest ovs package versions in ubuntu and the cloud archive. This is fixed in the following versions of ovs: xena (2.16.4), jammy/yoga (2.17.2), and kinetic (3.0.0).


fnordahl (approved):
This is indeed fixed in the referenced OVS versions, LGTM

PR #927, opened 2022-09-23
ovn: Remove workaround for EOL release

fnordahl:
The workaround was specifically targeting Groovy which is now EOL.

Also remove the `current_release` class variable as it is no longer needed.

Related-Bug: #1852221


ajkavanagh (approved):
LGTM

PR #930, opened 2022-09-27
Pin pyopenssl<22.1.0

freyes:
pyopenssl depends on a newer version of cryptography since 22.1.0.

https://github.com/pyca/pyopenssl/commit/a145fc3bc6d2e943434beb2f04bbf9b18930296f


ajkavanagh (approved):
I think at this stage in the cycle it's okay to pin this dependency.  We definitely need a "review all pins" for the next cycle though.

PR #933, opened 2022-09-28
Tests for ovn-central downscaling and cluster status

mkalcok:
The test for cluster member kick might be a bit controversial as it alters the model permanently. To account for this I also proposed increasing the ovn-central unit count from 3 to 4 in testing bundles. The test logic goes like this:

- Screw with ovn sb/nb processes on selected unit to prevent it from leaving gracefully
- Remove selected unit
- Verify that it shows up as "UNKNOWN" in output of "cluster-status" action
- Request kick of the "UNKNOWN" servers
- verify that they were removed from the cluster

Alternative (non-destructive) testing scenario could be:

- Request kick of a selected cluster member servers (without removing unit)
- verify that servers from that unit no longer show up in "cluster-status" output
- run re-join commands on unit to bring it back to the cluster

I prefer the currently proposed destructive scenario because it also tests that servers from units that did not leave cluster gracefully (e.g. their machine were killed abruptly) show up as "UNKNOWN" in "cluster-status".


zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -701,3 +701,134 @@ def run_tests(self):
         self.run_package_change_test(
             'ovn-central',
             'ovn-central')
+
+
+class OVNCentralDownscaleTests(test_utils.BaseCharmTest):
+    """Tests for cluster-status and cluster-kick actions."""
+
+    SB_CMD = "ovs-appctl -t /var/run/ovn/ovnsb_db.ctl {}"
```
fnordahl, line 709: This should be `ovn-appctl`

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -701,3 +701,134 @@ def run_tests(self):
...
+            sb_id = sb_cluster_status["Stdout"].splitlines()[0]
+            expected_mapping[unit_name] = {"sb_id": sb_id, "nb_id": nb_id}
+
+        for unit_name, unit_data in expected_mapping.items():
+            sb_id = unit_data["sb_id"]
+            nb_id = unit_data["nb_id"]
+            self.assertEqual(sb_status["Servers"][sb_id]["Unit"], unit_name)
+            self.assertEqual(nb_status["Servers"][nb_id]["Unit"], unit_name)
+
+    def test_cluster_kick(self):
```
fnordahl, line 770: The approach of modifying the test bundles to have a extra unit for this test is not acceptable.

Please extend this test to add/remove units when needed [0] and make sure you leave the model in a usable state for other tests to succeed when you're done.

Please also add a test that confirms that a regular remove-unit procedure leads to successful self-removal from the cluster.

0: https://github.com/openstack-charmers/zaza/blob/87e7a9c431c89ad1577106d0eda1818561f343a9/zaza/model.py#L2680

fnordahl (changes requested):
Thank you for your work on this, I have a few comments in-line.

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -701,3 +701,223 @@ def run_tests(self):
...
+        """Add specified number of units to ovn-central application.
+
+        This function also waits unit the application reaches active state.
+        """
+        zaza.model.add_unit(
+            "ovn-central",
+            count=number_of_units,
+            wait_appear=True
+        )
+        zaza.model.wait_for_application_states()
```
fnordahl, line 751: I took this for a spin and I see the test code moving forward before everything is ready when adding and removing units, leading to spurious failures.

Normally this is due to a race where the test code sees the target state before the charm has started doing anything, and the solution is often something like this:
https://github.com/openstack-charmers/zaza-openstack-tests/blob/827d81fc12201e525d6d57d2908ee9be71f02649/zaza/openstack/charm_tests/ovn/tests.py#L625-L627

However, that trick does not work here as zaza/libjuju appears to get itself into trouble when units (dis-)appear.

Adding a call to `block_untill_all_units_idle` before `wait_for_application_states` does appear to help, what do you think?:
```python
        zaza.model.add_unit(...)
        zaza.model.block_until_all_units_idle()
        zaza.model.wait_for_application_states()
```

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -701,3 +701,223 @@ def run_tests(self):
...
+
+    @staticmethod
+    def _remove_unit(unit_name):
+        """Add specified unit from ovn-central application.
+
+        This function also waits unit the application reaches active state.
+        """
+        zaza.model.destroy_unit("ovn-central", unit_name)
+        zaza.model.wait_for_application_states()
```
fnordahl, line 760: As above, the test code sometimes progresses here before the model is ready. Try to add a `block_until_all_units_idle`?

fnordahl (changes requested):
This is looking really good!

I took this for a spin on a MAAS substrate with virtual machines, and ran into what appear to be timing/race issues, see comments in-line.

mkalcok:
Thanks for the review @fnordahl I applied your suggestions and It does not break the tests, however I don't have fast (slow?) enough environment to test if it solves the race condition (I was not experiencing it before either). Do you think that the additional `block_until_all_units_idle()` solves the issue in a "true" sense, or its execution simply adds enough time for units to start doing "something" before the `wait_for_application_states()` is executed?

fnordahl:
> Thanks for the review @fnordahl I applied your suggestions and It does not break the tests, however I don't have fast (slow?) enough environment to test if it solves the race condition (I was not experiencing it before either). Do you think that the additional `block_until_all_units_idle()` solves the issue in a "true" sense, or its execution simply adds enough time for units to start doing "something" before the `wait_for_application_states()` is executed?

That is a good question. There is a failed test result which may need some more investigation, not sure if it's an instance of this problem or a payload problem. Would you mind taking a look?

https://openstack-ci-reports.ubuntu.com/artifacts/392/859720/4/check/jammy-yoga/3922f64/job-output.txt

If it turns out to be the test code progressing before the model is ready, I guess we need to figure out how to get `wait_for_agent_status` to work in conjunction with adding/destroying units or something similar.

mkalcok:
> That is a good question. There is a failed test result which may need some more investigation, not sure if it's an instance of this problem or a payload problem. Would you mind taking a look?

This failure is caused by test not finding a unit that hosts either ovnsb or ovnnb leader. The test searches for substring [`leader:`](https://github.com/mkalcok/zaza-openstack-tests/blob/570a9f1a5719ff84e6b79a0fca46bd422449b0c2/zaza/openstack/charm_tests/ovn/tests.py#L914) in workload message. I'm pretty sure that this is just a timing issue of workload message not being updated after OVN leadership change. I think that forcing `update-status` hook before searching for leader will prevent this failure in future.

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -896,6 +896,10 @@ def test_cluster_downscale(self):
         application = zaza.model.get_application("ovn-central")
 
         logging.info("Removing unit that hosts OVN follower server.")
+        # Run `update-status` hook. This updates the workload status message,
+        # helping us to correctly identify unit that does not host OVN leader
+        # servers.
+        zaza.run(application.run("hooks/update-status"))
```
fnordahl, line 902: This does appear a bit brittle, I see one failure in the gate, and I had one failure locally after running the tests a couple of times.

One thing is that multiple charm units can have the "leader:" text in their status, the Northbound and Sothbound leader does not need to reside on the same unit.

Another thing is that juju might not be done executing when you return from the above call.

An alternative could be to collect this information with the cluster-status action you've just added?

fnordahl (changes requested):

zaza/openstack/charm_tests/ovn/tests.py
mkalcok: > One thing is that multiple charm units can have the "leader:" text in their status, the Northbound and Sothbound leader does not need to reside on the same unit.

I don't think that this is an issue, we just need a unit which hosts at least one leader. If the cluster downscales cleanly, it means that the graceful departure works even if we remove unit with leader server (it doesn't necessarily need to host both leaders).

> Another thing is that juju might not be done executing when you return from the above call.

Hmm, I assumed that [`Application.run()`](https://pythonlibjuju.readthedocs.io/en/latest/api/juju.application.html#juju.application.Application.run) waits for hook to finish given that it has also an option for timeout.

> An alternative could be to collect this information with the cluster-status action you've just added?

Definitely. This just seemed way more simple at the time.

>  I see one failure in the gate, and I had one failure locally after running the tests a couple of times.

Any chance you still have logs of the failure? I'd like to examine it and the OSCI seems to fully pass with these updated tests.

fnordahl: > > Another thing is that juju might not be done executing when you return from the above call.
> 
> Hmm, I assumed that [`Application.run()`](https://pythonlibjuju.readthedocs.io/en/latest/api/juju.application.html#juju.application.Application.run) waits for hook to finish given that it has also an option for timeout.

The difference here is that after the hook tool has finished the global view of the Juju model status, on which your code depends, will update asynchronously. So for this to work, you most likely need to wait for something to happen+wait for things to be idle again before the status is accurate.

> > I see one failure in the gate, and I had one failure locally after running the tests a couple of times.
> 
> Any chance you still have logs of the failure? I'd like to examine it and the OSCI seems to fully pass with these updated tests.

I unfortunately do not have the logs available.

mkalcok: "leader/follower detection" now uses `cluster-status` action

fnordahl (approved):
Thank you for the contribution, LGTM!

PR #938, opened 2022-09-30
Pin pyOpenSSL in setup.py [wallaby backport]

swalladge:
It was pinned in requirements.txt,
but this isn't picked up when installing this via pip.
So, it needs to be pinned in the setup.py install_require list too.

(cherry picked from commit a6a540aebbce24bf5348d291115302dacfd2028a)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,11 @@ install_require = [
     'async_generator',
     'boto3',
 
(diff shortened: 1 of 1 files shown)


freyes (approved):
LGTM

PR #943, opened 2022-09-30
Add jammy image to octavia ensure_lts_images()

coreycb:


wolsen:
I'm only concerned about whether we really need 3 full LTS images loaded. I don't think this should be a problem, but most of the instances really aren't that large. Let's go for it - can always remove older LTS releases to keep at most 2.

wolsen (approved):

PR #948, opened 2022-10-11
k8s: use action to retrieve admin password

javacruft:
The Keystone K8S operator now uses generated passwords; make use of the helper action to retrieve the admin password.


javacruft:
Testing here: https://review.opendev.org/c/openstack/charm-keystone-k8s/+/860826

PR #953, opened 2022-10-14
Add method to check OpenStack endpoints

gnuoy:
Add method to check OpenStack endpoints are returning acceptable http codes. This should be used with caution as a charm whould indicate if its payload is not ready via workload status and workload status messages

### DIFF
--- a/zaza/openstack/charm_tests/keystone/setup.py
+++ b/zaza/openstack/charm_tests/keystone/setup.py
@@ -15,6 +15,8 @@
 """Code for setting up keystone."""
 
 import logging
+import requests
+import tenacity
 
 import keystoneauth1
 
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/keystone/setup.py
```diff
@@ -167,3 +169,43 @@ def add_tempest_roles():
...
     _add_additional_roles(TEMPEST_ROLES)
+
+
+def wait_for_url(url, ok_codes=None):
+    """Wait for url to return acceptable return code.
+
+    :param url: url to test
+    :type url: str
+    :param ok_codes: HTTP codes that are acceptable
+    :type ok_codes: List[int]
```
ajkavanagh, line 180: I think this is `Optional[List[int]]`?

ajkavanagh (commented):
LGTM; one tiny nitpick.

ajkavanagh (approved):
LGTM

PR #954, opened 2022-10-18
Move add-disk/remove-disk tests to a different class

lmlg:
We are seeing intermittent failures with disk removal, and the most likely cause for this is that the cluster we are using is too small. However, we can't simply increase the number of units in the bundles since there are clients that also run these tests, and having them all update their bundles can be cumbersome. On the other hand, there's no reason for non-ceph charms to actually run these tests, so we are moving them to a separate class, where we can try out the bigger cluster solution without bothering anyone else.

ETA: Being tested on the following patchsets:
https://review.opendev.org/c/openstack/charm-ceph-osd/+/861677
https://review.opendev.org/c/openstack/charm-ceph-mon/+/861678

### DIFF
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -149,16 +149,6 @@ class CephTest(test_utils.OpenStackBaseTest):
     def setUpClass(cls):
         """Run the ceph's common class setup."""
         super(CephTest, cls).setUpClass()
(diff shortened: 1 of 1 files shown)


coreycb:
Hi @lmlg , is this change still needed?

lmlg:
> Hi @lmlg , is this change still needed?

Not anymore. I'm going to close this PR.

PR #956, opened 2022-10-20
Allow tempest to run in minimal mode

gnuoy:
Add test class that runs tempest in miminal mode. This is useful for testing that the OpenStack apis are responding but there are components not configured like an external network. In addtion some missing doc strings were added and support for neutron tempest configuration pre ussuri removed as there is a branch for that.


ajkavanagh (approved):
LGTM

PR #957, opened 2022-10-20
Pin pyOpenSSL to be compatible with cryptography

ajkavanagh:
pyOpenSSl has been pinned < 22.1.0 but this allows pyOpenSSL 22.0.0 to be installed that requires cryptograph 35.0+.  Unfortunately, cryptography is pinned < 3.4 which creates a conflict leading to strange openssl errors.

The solution is to pin pyOpenSSL < 22.0.0, which will currently select 21.0.0, and that has a requirement for cryptography 3.3+. This combination will select pyOpenSSL 21.0.0 and cryptography 3.3.2 which are compatible.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,10 +5,10 @@ lxml<4.6.3
 pyparsing<3.0.0  # pin for aodhclient which is held for py35
 aiounittest
 async_generator
-# pyopenssl depends on a newer version of cryptography since 22.1.0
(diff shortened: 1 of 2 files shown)


jadonn:
Do we still need to change the pin for pyOpenSSL from <22.1.0 to <22.0.0? I see that we still pin pyOpenSSL to <22.1.0 today.

ajkavanagh:
I think this is OBE.

PR #958, opened 2022-10-21
Don't test nrpe bits on rabbitmq-server. (#793) [yoga backport]

freyes:
When the nrpe application is not deployed in the model skip the test that checks for the correct configuration of the nagios checks, this is because at the moment nrpe has no support for jammy, hence is not being deployed in the jammy-yoga bundles.

(cherry picked from commit 5a125a4f9ea52bcd532768d69139e59247824db0)


freyes:
Associated charm patch - https://review.opendev.org/c/openstack/charm-rabbitmq-server/+/862396

freyes:
yoga fixed by https://github.com/openstack-charmers/zaza-openstack-tests/pull/995

PR #959, opened 2022-10-25
Align py3.x tox targets with jammy/kinetic versions

coreycb:


freyes (approved):
py3.10 is needed for jammy support but requires switching away from nose.

dropping 3.5 (xenial), 3.6 (bionic) and 3.7 makes sense for master.

PR #968, opened 2022-10-25
Test all supported LTS releases with octavia-diskimage-retrofit

freyes:


freyes:
this PR is exercised by this change - https://review.opendev.org/c/openstack/charm-octavia-diskimage-retrofit/+/862643

jadonn:
Do we still need this PR? If we do, what can we do to move it forward?

PR #970, opened 2022-10-26
Retry check of service IP configured

freyes:
There are scenarios where the config-changed hook can complete, yet the service IP get configured many seconds after, because a relation-changed hook execution needs to be triggered on the hacluster side of the relation.

This change adds a retry to the check (10 times with a 2 seconds wait time).

This issue was found at the gate https://review.opendev.org/c/openstack/charm-designate-bind/+/861417

### DIFF
--- a/zaza/openstack/charm_tests/designate_bind/tests.py
+++ b/zaza/openstack/charm_tests/designate_bind/tests.py
@@ -16,6 +16,13 @@
 import logging
 import os
 
+from tenacity import (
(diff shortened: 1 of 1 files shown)


freyes:
This PR is being exercised by this patch - https://review.opendev.org/c/openstack/charm-designate-bind/+/862735

ajkavanagh (approved):
LGTM

PR #972, opened 2022-10-27
Add retry decorator to masakari segment creation

coreycb:
The create_segment() function is often failing in the gate due to being unable to establish a connection to the masakari endpoint. This will allow some more time for the endpoint to become available when this error occurs.


coreycb:
Error can be seen here: https://review.opendev.org/c/openstack/charm-masakari/+/850201 (note that I ran jammy-zed successfully locally)

ChrisMacNaughton (approved):

PR #974, opened 2022-11-07
Test get-csr with existing CA

gnuoy:
Check that if an existing CA is present then get-csr requires the force flag. Test using new action name regenerate-intermediate-ca


ajkavanagh (approved):
LGTM

PR #978, opened 2022-12-23
Don't test nrpe bits on rabbitmq-server. (#793)

ajkavanagh:
When the nrpe application is not deployed in the model skip the test that checks for the correct configuration of the nagios checks, this is because at the moment nrpe has no support for jammy, hence is not being deployed in the jammy-yoga bundles.

(cherry-picked from master at 5a125a4f9ea52bcd532768d69139e59247824db0)

### DIFF
--- a/.github/workflows/tox.yaml
+++ b/.github/workflows/tox.yaml
@@ -20,7 +20,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install tox tox-gh-actions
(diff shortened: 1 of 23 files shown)


coreycb:
Hi Alex, is this change still needed?

ajkavanagh:
> Hi Alex, is this change still needed?

I still think the skip is useful; I'll merge the updates into it and sort the conflicts.

ajkavanagh:
After the merge, it appears the main functionality has already made it to stable/yoga.  Therefore, closing this.

PR #979, opened 2023-01-06
Fix race in test 408 for rabbitmq

ajkavanagh:
There is a race in the 408 test for rabbitmq where the config-change to enable ssl causes a leader-settings-changed hook in the non-leader units which results in a rabbitmq service restart.  This can happen at exactly the same time as the test attempts to establish a connection with the that unit. This patch retries the connection attempt.

Note that this may only be a partial fix as it's possible that a restart will happen just after the connection is made, which would then result in a test failure.

Related-Bug: LP#2002156

### DIFF
--- a/.github/workflows/tox.yaml
+++ b/.github/workflows/tox.yaml
@@ -19,7 +19,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        sudo apt-get install -q --yes libxml2-dev libxslt1-dev libyaml-dev
+        sudo apt-get install -q --yes libxml2-dev libxslt1-dev
         python -m pip install --upgrade pip
         pip install tox tox-gh-actions
     - name: Lint with tox
--- a/.github/workflows/tox.yaml
+++ b/.github/workflows/tox.yaml
@@ -19,7 +19,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        sudo apt-get install -q --yes libxml2-dev libxslt1-dev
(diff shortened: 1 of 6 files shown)


freyes (changes requested):
The change LGTM, although it's failing on the CI, this probably needs to get libyaml-dev installed.

```
      /opt/hostedtoolcache/Python/3.9.16/x64/include/python3.9/cpython/unicodeobject.h:446:26: note: declared here
        446 | static inline Py_ssize_t _PyUnicode_get_wstr_length(PyObject *op) {
            |                          ^~~~~~~~~~~~~~~~~~~~~~~~~~
      error: command '/usr/bin/gcc' failed with exit code 1
```

ajkavanagh:
> The change LGTM, although it's failing on the CI, this probably needs to get libyaml-dev installed.
> 
> ```
>       /opt/hostedtoolcache/Python/3.9.16/x64/include/python3.9/cpython/unicodeobject.h:446:26: note: declared here
>         446 | static inline Py_ssize_t _PyUnicode_get_wstr_length(PyObject *op) {
>             |                          ^~~~~~~~~~~~~~~~~~~~~~~~~~
>       error: command '/usr/bin/gcc' failed with exit code 1
> ```

Nope, it wasn't that.  It turns out the error is:

```
ext/_yaml.c:25698:25: error: ‘PyTypeObject’ {aka ‘struct _typeobject’} has no member named ‘tp_print’
```

And my guess is because PyYAML is really old and won't build against Python 3.9.  I'll look at the pins and see what's happening.

freyes (approved):
LGTM

PR #982, opened 2023-01-16
Revert "Test get-csr with existing CA"

gnuoy:
The original change that added this action is blocked by tox4 issues and this change had broken vault in the gate. Let's revert it and land it when things have calmed down.

Reverts openstack-charmers/zaza-openstack-tests#974


freyes (approved):
Makes sense to me, this needs to land.

PR #983, opened 2023-01-20
Migrate several Ceph test cases to not use the openstack base

ChrisMacNaughton:


ChrisMacNaughton:
This change is verified in https://review.opendev.org/c/openstack/charm-ceph-mon/+/871295

PR #984, opened 2023-01-25
Update ceph tests to deal with non-OpenStack substrates

ChrisMacNaughton:


coreycb:
@UtkarshBhatthere @sabaini Do you know if we still need this change?

PR #986, opened 2023-02-01
Fix handling of libjuju calls

hemanthnakkina:
python-libjuju 3.x has changes related to
output variables for classes like Actions.
This patch handles those cases.


hemanthnakkina:
This fix is not required.

PR #987, opened 2023-02-01
juju3

gnuoy:


jadonn:
@gnuoy Do you still want to move forward with this pull request? We're pruning old pull requests from Zaza OpenStack Tests, and it seems like this draft has not received any activity for several months. We can keep this PR if you would like, or we can close it.

PR #989, opened 2023-02-08
Modified test to use additionally provided block device from bundle rather than temporarily creating and using loopback devices.

UtkarshBhatthere:
The following change modifies the add/remove device test to use an already present osd block device rather than loop-devices. To maintain the idempotency of the test, the order of add-device/ remove-device operation has been flipped.

Test being verified at: https://review.opendev.org/c/openstack/charm-ceph-osd/+/872856

### DIFF
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -150,16 +150,6 @@ class CephTest(test_utils.BaseCharmTest):
     def setUpClass(cls):
         """Run the ceph's common class setup."""
         super(CephTest, cls).setUpClass()
-        cls.loop_devs = {}   # Maps osd -> loop device
-        for osd in (x.entity_id for x in zaza_model.get_units('ceph-osd')):
-            loop_dev = zaza_utils.add_loop_device(osd, size=10).get('Stdout')
-            cls.loop_devs[osd] = loop_dev
-
-    @classmethod
-    def tearDownClass(cls):
-        """Run the ceph's common class teardown."""
-        for osd, loop_dev in cls.loop_devs.items():
-            zaza_utils.remove_loop_device(osd, loop_dev)
 
     def osd_out_in(self, services):
         """Run OSD out and OSD in tests.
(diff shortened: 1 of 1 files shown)


lmlg:
As part of this PR, we should remove all uses of loopback creation / tear down as well (present in the constructor of the CephTest class).

sabaini:
Some minor comments inline

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -522,68 +517,122 @@ def get_local_osd_id(self, unit):
...
+            )
+            # Save for later use.
+            osd_info[unit] = device_info
+        logging.debug('Removed OSD Info: {}'.format(osd_info))
+        zaza_model.wait_for_application_states()
+
+        logging.info('Recycling previously removed disks')
+        for unit in list(osd_info):
+            block_dev = osd_info[unit]['block-device']
+            logging.info("Found device {} on unit {}".format(block_dev, unit))
```
ajkavanagh, line 593: It's very minor, but it is interesting to note that logging calls (i.e. to python's logger) are okay to use the more legacy `%s` formatting option.  That's because, the string is only interpolated (evaluated) if the logger actually writes the log somewhere.  `.format()` is evaluated regardless.

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -522,68 +517,122 @@ def get_local_osd_id(self, unit):
...
-        for param in params:
-            osd_id = self.get_local_osd_id(param['unit'])
-            param.update({'osd-id': osd_id})
-            action_obj = zaza_model.run_action(
-                unit_name=param['unit'],
-                action_name='remove-disk',
-                action_params={'osd-ids': osd_id, 'timeout': 5,
-                               'format': 'json', 'purge': False}
+        logging.info('Purge/Removing added OSDs')
+        for unit in list(osd_info):
```
ajkavanagh, line 606: as above.

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -522,68 +517,122 @@ def get_local_osd_id(self, unit):
...
+        return osd_devices
+
+    def remove_disk_from_osd_unit(self, unit, osd_id, is_purge: bool = False):
+        """Remove provided osd_id from unit."""
+        action_obj = zaza_model.run_action(
+            unit_name=unit,
+            action_name='remove-disk',
+            action_params={
+                'osd-ids': osd_id, 'timeout': 5,
+                'format': 'json', 'purge': is_purge
```
ajkavanagh, line 546: From a code-reading perspective, this might be nicer as 4 lines.  i.e. then it's obvious that there are 4 items in the dictionary, etc.

ajkavanagh (changes requested):
Generally, the utility functions ought to have their parameters described (with types) and the return values described, unless they are trivial, please.  (e.g. :param:, :type:, :returns:, :rtype: as indicated).  It's perfectly okay for the `test_` functions to not have this as it's known that they take nothing and return nothing, and are only executed for their side-effects.

It's really important that test code is kept to the same standards as the rest of the charm code, please.

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -150,16 +150,11 @@ class CephTest(test_utils.BaseCharmTest):
...
-        for osd in (x.entity_id for x in zaza_model.get_units('ceph-osd')):
-            loop_dev = zaza_utils.add_loop_device(osd, size=10).get('Stdout')
-            cls.loop_devs[osd] = loop_dev
 
     @classmethod
     def tearDownClass(cls):
         """Run the ceph's common class teardown."""
-        for osd, loop_dev in cls.loop_devs.items():
-            zaza_utils.remove_loop_device(osd, loop_dev)
+        pass
```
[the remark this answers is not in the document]
UtkarshBhatthere: Thanks, I was a bit confused around what to do in this case.

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: ack

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: Thanks I have changed instances of .format() based string formatting to %s based string formatting.

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: Thanks for your comments @ajkavanagh. I have resolved them. Currently waiting for [PR#872856](https://review.opendev.org/c/openstack/charm-ceph-osd/+/872856) for verification.

ajkavanagh:
@UtkarshBhatthere is this being exercised in a gerrit review?  If so, it would be good to add it to the description so we can see that it is working.  Thanks!

ChrisMacNaughton:
This is functionally verified in https://review.opendev.org/c/openstack/charm-ceph-osd/+/872856

PR #991, opened 2023-02-13
Pin PyYAML based on python version

coreycb:
This is a partial cherry-pick from commit
9ad7e55a5f559790041271d2e5964dfb5f8bcc86


coreycb:
Starting over with more patches batched

PR #992, opened 2023-02-14
Try to get tests to pass with assertions

coreycb:


coreycb:
Starting over with more patches batched

PR #1005, opened 2023-02-21
Test the keystone rotate-service-user-password action

ajkavanagh:
This test, added as part of the default keystone tests, tests that the service user password can be rotated for the glance application.

### DIFF
--- a/zaza/openstack/charm_tests/keystone/tests.py
+++ b/zaza/openstack/charm_tests/keystone/tests.py
@@ -203,7 +203,8 @@ class CharmOperationTest(BaseKeystoneTest):
                 '{} is not deployed, so not doing password change'
                 .format(GLANCE_APP))
         # keep the old password to verify it is changed.
-        old_passwd = juju_utils.leader_get(GLANCE_APP, GLANCE_PASSWD_KEY)
+        old_passwd = juju_utils.leader_get(
+            self.application_name, GLANCE_PASSWD_KEY)
 
         # verify that images can be listed.
         glance_client = openstack_utils.get_glance_session_client(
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/keystone/tests.py
```diff
@@ -191,6 +192,56 @@ def test_rotate_admin_password(self):
...
+            self.application_name, GLANCE_PASSWD_KEY)
+        new_passwd_conf = _get_password_from_leader()
+        self.assertNotEqual(old_passwd_leader_storage,
+                            new_passwd_leader_storage)
+        self.assertNotEqual(old_passwd_conf,
+                            new_passwd_conf)
+        self.assertEqual(new_passwd_leader_storage, new_passwd_conf)
+
+        # verify that the images can still be listed.
+        glance_client = openstack_utils.get_glance_session_client(
```
freyes, line 241: shouldn't you get a new token?, I'm not sure what kind of caching API service do for token validation.

freyes (changes requested):
Hi Alex, I added an inline comment, also do you have an idea how long it takes to run this test?

zaza/openstack/charm_tests/keystone/tests.py
ajkavanagh: No, I'm pretty sure that a new token is needed.  The password that has changed is the one between the glance service and keystone for glance to do it's things rather than the authentication to glance via the API.

ajkavanagh:
> Hi Alex, I added an inline comment, also do you have an idea how long it takes to run this test?

About a minute roughly, with this model.  i.e. it's quite quick and not really going to impact the gate at all.

freyes (approved):
LGTM

PR #1012, opened 2023-02-23
This fixes flakyness with iscsi tests. There was a previous fix that

lmlg:

https://review.opendev.org/c/openstack/charm-ceph-iscsi/+/871154


freyes (approved):
LGTM

coreycb (changes requested):
I don't see the refresh_partitions function in stable/xena and stable/yoga. Do we need it there first?

lmlg:
> I don't see the refresh_partitions function in stable/xena and stable/yoga. Do we need it there first?

For yoga, there's this PR: https://github.com/openstack-charmers/zaza-openstack-tests/pull/1019
I'll make one for Xena as well. We don't strictly need this one first, but it would be nice to have, as it's blocking some patchsets.

PR #1013, opened 2023-02-23
Fix iscsi tests (ussuri)

lmlg:

https://review.opendev.org/c/openstack/charm-ceph-iscsi/+/870871

### DIFF
--- a/zaza/openstack/charm_tests/ceph/iscsi/tests.py
+++ b/zaza/openstack/charm_tests/ceph/iscsi/tests.py
@@ -238,6 +238,10 @@ class CephISCSIGatewayTest(test_utils.BaseCharmTest):
             action_params={
                 'name': self.EC_METADATA_POOL}))
 
+    def refresh_partitions(self, ctxt):
+        """Refresh kernel partition tables in client."""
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/ceph/iscsi/tests.py
```diff
@@ -250,9 +254,11 @@ def run_client_checks(self, test_ctxt):
         """
         self.create_iscsi_target(test_ctxt)
         self.login_iscsi_target(test_ctxt)
+        self.refresh_partitions(test_ctxt)
         self.check_client_device(test_ctxt, init_client=True)
         self.logout_iscsi_targets(test_ctxt)
         self.login_iscsi_target(test_ctxt)
+        self.refresh_partitions(test_ctxt)
         self.check_client_device(test_ctxt, init_client=False)
```
coreycb, line 262: Does this need another refresh_partitions call like stable/zed has?

coreycb (changes requested):

coreycb:
One more thing, we probably need to land this in stable/victoria and stable/xena as well. They're EOL but we still have to backport through them and users have to upgrade through them.

PR #1014, opened 2023-02-27
Skip test_930_scaleback on crmsh-4.4.0-1ubuntu1

freyes:
The crmsh package available in kinetic fails to put a cluster node in maintenance mode, this is part of the scaleback process, more details on the failure available at the related bug.

This change introduces a new decorator skipVersion() that allows to provide a list of package versions with an operation flag on how dpkg should compare the version(s).

The test HaclusterTest.test_930_scaleback() is disabled when running crmsh-4.4.0-1ubuntu1, at the moment this would be kinetic, although if any new package gets released this test will be re-enabled automatically allowing us catch early if the test got fixed or not.

Related-Bug: http://pad.lv/1972730

### DIFF
--- a/unit_tests/charm_tests/test_utils.py
+++ b/unit_tests/charm_tests/test_utils.py
@@ -185,6 +185,25 @@ class TestBaseCharmTest(ut_utils.BaseTestCase):
 
         self.config_current.assert_called_once_with(None, None)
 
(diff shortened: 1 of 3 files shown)


freyes:
This change is exercised by https://review.opendev.org/c/openstack/charm-hacluster/+/874686

PR #1016, opened 2023-03-03
This fixes flakyness with iscsi tests. There was a previous fix that

lmlg:

https://review.opendev.org/c/openstack/charm-ceph-iscsi/+/870117

### DIFF
--- a/zaza/openstack/charm_tests/ceph/iscsi/tests.py
+++ b/zaza/openstack/charm_tests/ceph/iscsi/tests.py
@@ -238,6 +238,10 @@ class CephISCSIGatewayTest(test_utils.BaseCharmTest):
             action_params={
                 'name': self.EC_METADATA_POOL}))
 
+    def refresh_partitions(self, ctxt):
+        """Refresh kernel partition tables in client."""
+        self.run_commands(ctxt['client_entity_id'], ('partprobe', ), ctxt)
+
     def run_client_checks(self, test_ctxt):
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/ceph/iscsi/tests.py
```diff
@@ -250,9 +254,11 @@ def run_client_checks(self, test_ctxt):
         """
         self.create_iscsi_target(test_ctxt)
         self.login_iscsi_target(test_ctxt)
+        self.refresh_partitions(test_ctxt)
         self.check_client_device(test_ctxt, init_client=True)
         self.logout_iscsi_targets(test_ctxt)
         self.login_iscsi_target(test_ctxt)
+        self.refresh_partitions(test_ctxt)
         self.check_client_device(test_ctxt, init_client=False)
```
coreycb, line 262: Does this need another call to refresh_partitions?

coreycb (changes requested):

zaza/openstack/charm_tests/ceph/iscsi/tests.py
lmlg: Yes, it's needed after every login call.

coreycb: Ok I noticed in later branches there's a refresh_partitions call after the final check_client_device. I wasn't sure if that was needed here.

PR #1017, opened 2023-03-07
Add test to scale application out and in

gnuoy:
Add test that scales an application out and in and checks each stage with tempest.


freyes (changes requested):
Hi Liam,

Some inline comments.

freyes (approved):
LGTM

PR #1020, opened 2023-03-09
Fix ceph-iscsi tests (zed)

lmlg:

https://review.opendev.org/c/openstack/charm-ceph-iscsi/+/871317


coreycb:
LGTM and is already in antelope+

PR #1023, opened 2023-03-14
Ensure keystone endpoint is available

coreycb:
The keystone-ldap tests are failing often due to accessing the keystone endpoint before it is available. Adding tenacity retries here to deal with this.


freyes (approved):
LGTM. Waiting for the CI results to return. https://review.opendev.org/c/openstack/charm-keystone-ldap/+/873802

PR #1024, opened 2023-03-14
Ensure keystone endpoint is available

coreycb:
The keystone-kerberos test setup is failing often due to accessing the keystone endpoint before it is available. Adding tenacity retries here to deal with this.


freyes (approved):
LGTM

PR #1025, opened 2023-03-16
Unblock after pristine deconfiguration

sabaini:
When a charm has a non-pristine disk deconfigured, it should leave the previously-set blocked state (LP #1988088).

### DIFF
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -413,6 +413,29 @@ class CephTest(test_utils.BaseCharmTest):
 
         zaza_model.wait_for_application_states()
 
+    def test_active_after_pristine_block(self):
+        """Test if we can get back to active state after pristine block.
+
+        Set a non-pristine status, then trigger update-status to see if it
+        clears.
+        """
+        logging.info('Setting Non-pristine status')
+        zaza_model.run_on_leader(
+            "ceph-osd",
+            "status-set blocked 'Non-pristine'"
+        )
+        ceph_osd_states = {
+            'ceph-osd': {
+                'workload-status': 'blocked',
+                'workload-status-message-prefix': 'Non-pristine'
+            }
+        }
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -413,6 +413,44 @@ def test_blocked_when_non_pristine_disk_appears(self):
...
+        This is to test the case of LP #1988088 where a charm would not come
+        out of blocked state after a non-pristine disk was deconfigured
+
+        We are using the same ephemeral disk as in the previous test, see
+        above `test_blocked_when_non_pristine_disk_appears` for details.
+        """
+        logging.info('Ensuring default configuration...')
+        juju_service = 'ceph-osd'
+        set_default = {
+            'ephemeral-unmount': '',
```
ChrisMacNaughton, line 428: It feels like this set_default should look at whatever's currently set in the model rather than just using charm defaults

sabaini: I've now added two calls to store app config away and re-store it at the end of the test method so we're going back to where we started -- does this address your concerns?

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -413,6 +413,29 @@ def test_blocked_when_non_pristine_disk_appears(self):
...
+        )
+        ceph_osd_states = {
+            'ceph-osd': {
+                'workload-status': 'blocked',
+                'workload-status-message-prefix': 'Non-pristine'
+            }
+        }
+        zaza_model.wait_for_application_states(states=ceph_osd_states)
+        logging.info('Running update-status action')
+        zaza_model.run_on_leader('ceph-osd', 'update-status')
```
ChrisMacNaughton, line 435: Does this not need to call `hooks/update-status`?

ChrisMacNaughton (approved):
LGTM - holding off on landing just until the review lands as well

PR #1027, opened 2023-03-22
Add test to rotate mysql service user password

ajkavanagh:
This new test verifies that keystone can have its password rotated and then still operate afterwards.  It verifies that the on-disk password is changed in the keystone application and that the user list can be performed.

### DIFF
--- a/zaza/openstack/charm_tests/mysql/tests.py
+++ b/zaza/openstack/charm_tests/mysql/tests.py
@@ -615,7 +615,7 @@ class MySQLInnoDBClusterRotatePasswordTests(MySQLCommonTests):
             'list-service-usernames',
             action_params={}
         )
-        usernames = action.data['results']['usernames']
+        usernames = action.data['results']['usernames'].split(',')
         self.assertIn('keystone', usernames)
         logging.info("... usernames: %s", ', '.join(usernames))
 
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/mysql/tests.py
```diff
@@ -567,6 +568,101 @@ def test_911_restart_after_group_replication_cache_set(self):
...
+        # now rotate the password for keystone
+        # run the action to rotate the password.
+        logging.info("Rotating password for keystone in mysql.")
+        zaza.model.run_action_on_leader(
+            self.application_name,
+            'rotate-service-user-password',
+            action_params={'service-user': 'keystone'},
+        )
+
+        # let everything settle.
```
freyes, line 643: something I've found necessary to avoid race conditions is to have a `zaza.model.wait_for_agent_status()` between running an action/config change and the `block_until_all_units_idle()`, because if the unit takes an extra few seconds to go into a running/executing state the block_until_all_units_idle() will find effectively all units in idle state and move on.

freyes (changes requested):
Hi Alex, in general looks good, just one detail which I think is valuable to avoid race conditions, see inline comment.

freyes (approved):
LGTM, thanks Alex.

freyes:
For the record, this is the patch that exercises this change - https://review.opendev.org/c/openstack/charm-mysql-innodb-cluster/+/878222 - they need to be merged in tandem.

freyes (approved):
LGTM

PR #1032, opened 2023-03-30
Move pristine test to ceph-mon tests

sabaini:
Pristine status test was applied to all ceph related charms, but only really is relevant for the ceph-mon charm.


sabaini:
Test is exercised in https://review.opendev.org/c/openstack/charm-ceph-osd/+/879099

lmlg:
lgtm

PR #1034, opened 2023-04-06
Make Ceph tests more generic

lmlg:
This PR removes some hardcoded values in Ceph tests (device names), and removes another test that doesn't make sense to run (The relation test, which compares unpredictable public IP addresses).

### DIFF
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -116,33 +116,6 @@ class CephLowLevelTest(test_utils.BaseCharmTest):
             self.assertEqual(pool['pg_autoscale_mode'], 'on')
 
 
-class CephRelationTest(test_utils.BaseCharmTest):
-    """Ceph's relations test class."""
-
-    @classmethod
-    def setUpClass(cls):
-        """Run the ceph's relations class setup."""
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -467,10 +447,15 @@ def test_blacklist(self):
...
         # Attempt to add device with existent path should succeed
+        unused_dev = self.get_unused_device(unit_name)
+        if unused_dev is None:
+            raise unittest.SkipTest(
+                "Skipping rest of test because there are no unused devices")
+
         action_obj = zaza_model.run_action(
             unit_name=unit_name,
             action_name='blacklist-add-disk',
-            action_params={'osd-devices': '/dev/vda'}
```
ChrisMacNaughton, line 473: I don't believe that this device is expected to be unused - on serverstack, vda will be mounted at /!

lmlg: right. it's "unused" from ceph's perspective. I guess "pristine" would be a better name?

ChrisMacNaughton: generally speaking, we want it to be an in-use device that's not used by Ceph, the version using /dev/vda is using the operating system's root disk

PR #1037, opened 2023-04-19
Update application name for skipVersion

coreycb:
Update application name for skipVersion() decorator used on test_930_scaleback() to 'keystone-hacluster'. In this manner, you would have to add a second skipVersion() decorator for other application names, such as 'nova-hacluster', if you wanted to check that application for the package version.

Hard-coding the application name doesn't feel perfect, but I've tried a few alternative approaches that weren't as good.

Alternative approaches:
  1) Find the application name based on the charm name. This
     isn't ideal because there may be multiple hacluster
     applications in the model, such as keystone-hacluster
     and nova-hacluster.
  2) Pass 'charm' as the first argument to skipVersion instead
     of 'application', and then generate the application name
     inside the _skipVersion_inner_2() decorator code based on:
     f"{args[0]._principle_app_name}-{charm}". For example:
     'keystone-hacluster'. This isn't ideal because you'd need
     prior knowledge of inner workings of the decorator to
     understand how it works.


freyes:
> Find the application name based on the charm name. This
> isn't ideal because there may be multiple hacluster
> applications in the model, such as keystone-hacluster
> and nova-hacluster.

I think this would be better, make the skipVersion() decorator capable of receiving an application name or a charm name. Then how that's used in test_930_scaleback() is using `@skipVersion(charm='hacluster', ...)`, because for our use cases that holds true, we do know that any environment running that test method with that deb package version will fail hence should be skipped.

coreycb:
On Wed, Apr 19, 2023 at 9:37 AM Felipe Reyes ***@***.***>
wrote:

> Find the application name based on the charm name. This
> isn't ideal because there may be multiple hacluster
> applications in the model, such as keystone-hacluster
> and nova-hacluster.
>
> I think this would be better, make the skipVersion() decorator capable of
> receiving an application name or a charm name. Then how that's used in
> test_930_scaleback() is using @skipVersion(charm='hacluster', ...),
> because for our use cases that holds true, we do know that any environment
> running that test method with that deb package version will fail hence
> should be skipped.
>
> The reason I didn't take that approach is because it's possible to have
multiple hacluster charms in the model. Would you suggest picking the first
application returned if a charm is specified?

freyes (approved):
LGTM

PR #1039, opened 2023-04-20
Increase tenacity delay for policyd tests

coreycb:
Double the tenacity stop_after_delay for the
get_client_and_attempt_operation methods. This fixes the "Unable to establish connection" test failures seen in: https://review.opendev.org/c/openstack/charm-cinder-ceph/+/878958


ajkavanagh (approved):
LGTM; does this require a backport to stable/antelope?

coreycb:
Thanks for the quick merge! I haven't seen it there but I'm guessing that it will be needed in stable/antelope too.

PR #1040, opened 2023-04-22
(s/antelope) Couple of pins to ensure stability on antelope

ajkavanagh:
These are a couple of pins on aodhclient for stability on antelope. This would be better as upper constraints, but that still a WIP item.


coreycb:
Hi Alex, is this still needed?

PR #1041, opened 2023-04-22
(s/antelope) Retry assert after service_ips removal

ajkavanagh:
Similar to when an IP has been added to the service_ips earlier in test_configure_ips(), this adds tenacity Retrying after an IP has been removed from the service_ips.

cherry-pick from 080a715130b76dc546ed61c6c20af84a88d7dfba that was PR#1036


coreycb:
Do we still need this?

ajkavanagh:
> Do we still need this?

Yes, I believe so.

PR #1043, opened 2023-04-22
(s/antelope) Change skipVersion decorator to a function

ajkavanagh:
The skipVersion decorator was proving to be tricky to use because it took an application name which we didn't want to hard-code, since it is in hacluster/tests.py. One alternative was to pass a charm name instead of application name, but if there were mutiple hacluster applications in the model (e.g. keystone-hacluster and nova-hacluster) we would have to pick just one to test the package version on. After further discussion we decided to change the decorator to a function so that we could use existing class variables to generate the desired hacluster application name.

cherry-pick 95fcad2a0a5c9b4960e8ab5e920729cd49c44aa3 From PR#1037


coreycb:
Alex, is this still needed?

PR #1049, opened 2023-05-15
Add some tenacity to ceph pool type test

sabaini:
Retry ceph pool type test to deal with races in setting up pools vs. running functests


coreycb:
Peter, is this still needed? It's in draft mode.

sabaini:
Thanks for the cleanup Corey. This was thought to make testing more robust but it looks like we don't need it after all. I'll close this PR; if we have issues with racey tests again I can dig it up

PR #1052, opened 2023-05-17
Feature/ussuri/test rotate passwords rabbitmq server (cherry-pick from master)

ajkavanagh:
This new test verifies that cinder can have its password rotated and
then still operate afterwards. It verifies that the on-disk password is
changed in the cinder application and that the user list can be
performed.

This is two cherry picks from master to test the rabbitmq-server rotate service user password feature:

https://github.com/openstack-charmers/zaza-openstack-tests/commit/6516a340453f525ea565b95e120c3e13979e63ab [2023-05-05] Update rabbitmq rotate password test to be more robust/correct
https://github.com/openstack-charmers/zaza-openstack-tests/commit/b29c11720e7504d72535c8598baaa33071445c27 [2023-03-23] Add test to rotate rabbitmq-server service user password

### DIFF
--- a/zaza/openstack/charm_tests/rabbitmq_server/tests.py
+++ b/zaza/openstack/charm_tests/rabbitmq_server/tests.py
@@ -21,6 +21,7 @@ import re
 import time
 import uuid
 import unittest
+import yaml
 
 import juju
 import tenacity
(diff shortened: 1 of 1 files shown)


freyes (approved):
LGTM, we need one more review to merge stable backports.

coreycb (approved):

PR #1059, opened 2023-05-23
Update ceph user creation test

sabaini:
Upstream cli changed output, see https://tracker.ceph.com/issues/61362


coreycb (approved):
Looks good to me and good idea to open the issue for it upstream. This gets the job done and the user is actually getting deleted, so that's good.

PR #1060, opened 2023-05-23
Switch from Quagga to FRR

dshcherb:
A charm that uses FRR instead of Quagga is now published under: https://charmhub.io/osci-frr

For our purposes FRR is a drop-in replacement of Quagga but the point of a change is to remove Quagga references for clarity.


fnordahl (approved):
LGTM, thanks!

PR #1063, opened 2023-05-26
Fix: upstream tooling output change

sabaini:
Ceph rbd mirroring fixes:

- Upstream ceph mirror tooling modernized its output, need to update test to reflect that.

- Avoid spamming run-action in a loop

- Don't consider system pools for failover

- Remove obsolete workaround for LP: #1820976


sabaini:
This is being exercised in https://review.opendev.org/c/openstack/charm-ceph-rbd-mirror/+/884510

lmlg (approved):
LGTM

PR #1064, opened 2023-05-26
Improve logging of _resource_removed()

freyes:
On failures there is no information of the state of the objects that couldn't be removed, maybe there are on their way to be removed and this is a failure visible on slow clouds and not an actual issue, this will help with troubleshooting.

(cherry picked from commit b453a850139f6e92af8dc3cf99aafca4ba5e38f4)


freyes:
https://review.opendev.org/c/openstack/charm-cinder-ceph/+/881045 - exercised by this patch. 

Marking this PR as draft to avoid getting it merged before the same patch makes it into master.

freyes:
PR for master -

coreycb (approved):

PR #515, opened 2021-03-04
Add Magnum tests

oprinmarius:
- Add Heat domain-setup action as a setup configuration option which can be used in other charms
   - test_100_domain_setup is being deprecated if the new configuration option is being used
- Add Magnum tests
  - test create magnum cluster template
  - test create magnum cluster
  - test config changes
  - test pause and resume
- TEST_MAGNUM_QCOW2_IMAGE_URL environment variable is required in the CI
   - The value of this variable should be a link to a fedora-coreos QCOW2 image

PR #1069, opened 2023-05-30
Magnum support [yoga]

freyes:
List of backported commits:
- a59b60d Add magnum section to tempest.conf template
- 9f19d81 Add support for Magnum to use different images per release.
- 21a5209 Add new required dns.nameservers config
- dd9e9ca Add Magnum tests (#515)


coreycb:
Felipe, is this still needed?

PR #1071, opened 2023-06-01
Fix ceph-iscsi tests (xena)

lmlg:
Another did mostly the same, but this PR makes it more similar to what we have pending for the stable branches.

https://review.opendev.org/c/openstack/charm-ceph-iscsi/+/871154


sabaini:
lgtm, cheers!

PR #1074, opened 2023-06-14
Fix: retry rbd pool status command

sabaini:
The rbd mirror status command has spurious faults on occasion. Retry after some waiting.


lmlg (approved):
LGTM

PR #1078, opened 2023-07-03
Convert images to raw if ceph image backend

rodrigogansobarbieri:
We are currently uploading qcow2 images, and
Nova is converting them to raw when running
the tests, sometimes timing out the tests.

With this change we are pre-converting the
images and uploading them as raw, so Nova
does not have to convert them.

### DIFF
--- a/zaza/openstack/charm_tests/glance/tests.py
+++ b/zaza/openstack/charm_tests/glance/tests.py
@@ -92,7 +92,8 @@ class GlanceTest(test_utils.OpenStackBaseTest):
                 self.glance_client,
                 image_url,
                 'cirros-test-import',
-                force_import=True)
+                force_import=True,
+                convert_if_ceph=False)
 
             disk_format = self.glance_client.images.get(image.id).disk_format
             self.assertEqual('raw', disk_format)
--- a/zaza/openstack/charm_tests/glance/tests.py
+++ b/zaza/openstack/charm_tests/glance/tests.py
@@ -93,7 +93,7 @@ class GlanceTest(test_utils.OpenStackBaseTest):
                 image_url,
                 'cirros-test-import',
                 force_import=True,
-                convert_if_ceph=False)
+                convert_image_to_raw_if_ceph_used=False)
 
             disk_format = self.glance_client.images.get(image.id).disk_format
             self.assertEqual('raw', disk_format)
--- a/zaza/openstack/utilities/openstack.py
+++ b/zaza/openstack/utilities/openstack.py
@@ -22,10 +22,12 @@ import datetime
 import enum
 import io
 import itertools
+import json
 import juju_wait
 import logging
 import os
 import paramiko
+import pathlib
 import re
 import shutil
 import six
@@ -2612,7 +2614,7 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
     :type local_path: str
     :param image_name: The label to give the image in glance
     :type image_name: str
-    :param disk_format: The of the underlying disk image.
+    :param disk_format: The format of the underlying disk image.
     :type disk_format: str
     :param visibility: Who can access image
     :type visibility: str (public, private, shared or community)
@@ -2651,10 +2653,65 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
     return image
 
 
+def is_ceph_image_backend(model_name=None):
+    """Check if glance is related to ceph, therefore using ceph as backend.
+
+    :returns: True if glance is related to Ceph, otherwise False
+    :rtype: bool
+    """
+    status = juju_utils.get_full_juju_status(model_name=model_name)
+    result = False
+    try:
+        result = 'ceph-mon' in (
+            status['applications']['glance']['relations']['ceph'])
+    except KeyError:
+        pass
+    logging.debug("Detected Ceph related to Glance?: {}".format(result))
+    return result
+
+
+def convert_image_format_to_raw_if_qcow2(local_path):
+    """Convert the image format to raw if the detected format is qcow2.
+
+    :param local_path: The path to the original image file.
+    :type local_path: str
+    :returns: The path to the final image file
+    :rtype: str
+    """
+    try:
+        output = subprocess.check_output([
+            "qemu-img", "info", "--output=json", local_path])
+    except subprocess.CalledProcessError:
+        logging.error("Image conversion: Failed to detect image format "
+                      "of file {}".format(local_path))
+        return local_path
+    result = json.loads(output)
+    if result['format'] == 'qcow2':
+        logging.info("Image conversion: Detected qcow2 vs desired raw format"
+                     " of file {}".format(local_path))
+        converted_path = pathlib.Path(local_path).resolve().with_suffix('.raw')
+        converted_path_str = str(converted_path)
+        if converted_path.exists():
+            logging.info("Image conversion: raw converted file already"
+                         " exists: {}".format(converted_path_str))
+            return converted_path_str
+        logging.info("Image conversion: Converting image {} to raw".format(
+            local_path))
+        try:
+            output = subprocess.check_output([
+                "qemu-img", "convert", local_path, converted_path_str])
+        except subprocess.CalledProcessError:
+            logging.error("Image conversion: Failed to convert image"
+                          " {} to raw".format(local_path))
+            return local_path
+        return converted_path_str
+    return local_path
+
+
 def create_image(glance, image_url, image_name, image_cache_dir=None, tags=[],
                  properties=None, backend=None, disk_format='qcow2',
                  visibility='public', container_format='bare',
-                 force_import=False):
+                 force_import=False, convert_image_to_raw_if_ceph_used=True):
     """Download the image and upload it to glance.
 
     Download an image from image_url and upload it to glance labelling
@@ -2676,6 +2733,10 @@ def create_image(glance, image_url, image_name, image_cache_dir=None, tags=[],
     :param force_import: Force the use of glance image import
         instead of direct upload
     :type force_import: boolean
+    :param convert_image_to_raw_if_ceph_used: force conversion of requested
+        image to raw upon download if Ceph is present in the model and
+        has a relation to Glance
+    :type convert_image_to_raw_if_ceph_used: boolean
     :returns: glance image pointer
     :rtype: glanceclient.common.utils.RequestIdProxy
     """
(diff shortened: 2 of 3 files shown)


zaza/openstack/utilities/openstack.py
```diff
@@ -2651,6 +2652,50 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
...
+        result = 'ceph-mon' in (
+            status['applications']['glance']['relations']['ceph'])
+    except KeyError:
+        pass
+    if result is False:
+        logging.info("Did not detect Ceph related to Glance")
+    return result
+
+
+def _convert_image_if_needed(local_path, disk_format):
```
ajkavanagh, line 2673: I don't think this function name conveys what the function is doing very well.  It seems like it is more of a `_ensure_image_is_raw_format()` ; I'm not set on *that* name, but it does describe the function more accurately.  Also, why is it nominally private? (i.e. with a leading `_`)?

zaza/openstack/utilities/openstack.py
```diff
@@ -2695,6 +2740,8 @@ def create_image(glance, image_url, image_name, image_cache_dir=None, tags=[],
         logging.info('Cached image found at {} - Skipping download'.format(
             local_path))
 
+    local_path = _convert_image_if_needed(local_path, disk_format)
+
```
ajkavanagh, line 2744: Unfortunately, this breaks the contract of the function's signature.  The `disk_format='qcow2'` as a default is no longer true, and this will silently nullify (for example), `test_412_image_conversion()` in zaza-openstack-tests that tests whether server-side image conversion works.

By extension it also breaks `add_image()` as that also has a `disk_format='qcow2'` default parameter, and even if the caller asks specifically for `qcow2`, they can't get it now.

So unfortunately, this is the wrong place to do this image conversion.

ajkavanagh (changes requested):
Firstly, thank you for taking the time to look at this issue.  I've put some comments in as prompts for (in my opinion!) improvements to the code, which I hope are useful.

However, there's a much more tricky issue that where the image conversion is done actually breaks the contract that `create_image()` *offers* via its function signature.  i.e. a qcow2 image can never be added to ceph under any circumstances, and this nullifies at least one test (see the comments).

I think what is needed is a more fine grained approach, and to use the `is_ceph_image_backend()` elsewhere in the code, and pass use the `image_format` parameter to get the `raw` image (even if `qcow2` is only available) by converting it explicitly.

Other ideas that may work better are adding a new default parameter called something like `upload_format='raw'` to request client side conversion prior to the upload.

zaza/openstack/utilities/openstack.py
```diff
@@ -2651,6 +2652,50 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
     return image
 
 
+def is_ceph_image_backend():
+    """Check if glance is related to ceph, therefore using ceph as backend.
+
+    :returns: True if glance is related to Ceph, otherwise False
+    :rtype: bool
+    """
+    status = juju_utils.get_full_juju_status(model.get_juju_model())
```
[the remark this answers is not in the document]
rodrigogansobarbieri: done (available in next patchset)

zaza/openstack/utilities/openstack.py
```diff
@@ -2651,6 +2652,50 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
...
+            status['applications']['glance']['relations']['ceph'])
+    except KeyError:
+        pass
+    if result is False:
+        logging.info("Did not detect Ceph related to Glance")
+    return result
+
+
+def _convert_image_if_needed(local_path, disk_format):
+    try:
```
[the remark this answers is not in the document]
rodrigogansobarbieri: done (available in next patchset)

zaza/openstack/utilities/openstack.py
```diff
@@ -2651,6 +2652,50 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
...
+        output = subprocess.check_output([
+            "qemu-img", "info", "--output=json", local_path])
+    except subprocess.CalledProcessError:
+        logging.error("Failed to detect image format of file {}".format(
+            local_path))
+        return local_path
+    result = json.loads(output)
+    if disk_format == 'raw' and result['format'] == 'qcow2':
+        converted_path = local_path.rsplit('.', 1)[0] + ".raw"
```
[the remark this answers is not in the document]
rodrigogansobarbieri: done (available in next patchset)

zaza/openstack/utilities/openstack.py
rodrigogansobarbieri: The problem I see now is that we have a contradiction on the reason for using uploading qcow2 images or raw images.

If we are using ceph, we should be using raw images to prevent major impact to the testing time, which is the motivation of this PR.
On the other hand, if we need to test image conversion, then we need at least a qcow2 image uploaded instead of only raw images.

I understand the contract violation, I am still thinking about how to adjust it, but I would like to propose starting from a different perspective. I believe the test "test_412_image_conversion" should upload a specific qcow2 image and test its conversion to raw, given that its what the test is really testing. By doing this, it could use a lighter image like cirros, and therefore not impact all other tests that are ALSO converting qcow2 to raw when they shouldn't need to.

Now, back to the contradiction, we remove the need to upload qcow2 if the backend is ceph (unless explicitely specified, such as the test 412 would be doing). We just need to figure out a clean function declaration parameters given it is a shared function. My priority is that the default behavior is the convertion if it is a ceph backend, so I am inclined along the lines of "force_specified_image_format=False" (needs a better name) parameter to be used only for test 412 or other situation where the image conversion needs to be overridden.

nobuto-m:
This is actually a similar topic to:
https://review.opendev.org/c/openstack/charm-glance-simplestreams-sync/+/809392

It would be nice if it can be completed with a functional test (we, the field team haven't been able to complete it so far since we are distracted by other duties)

zaza/openstack/utilities/openstack.py
rodrigogansobarbieri: Alex and I discussed in a meeting today and agreed to follow my proposal, where the new function parameter will be something like "convert_if_ceph=True", and only test 412 will specify convert_if_ceph=False to override.

zaza/openstack/utilities/openstack.py
```diff
@@ -2651,10 +2653,68 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
...
+            status['applications']['glance']['relations']['ceph'])
+    except KeyError:
+        pass
+    if result is False:
+        logging.info("Did not detect Ceph related to Glance")
+    return result
+
+
+def ensure_image_format(local_path, disk_format):
+    """Convert the image if its format differs from the desired format.
```
ajkavanagh, line 2675: This isn't quite right. What the function does is "if the disk_format param is 'raw' and the image format is 'qcow2' then convert the image to 'raw', otherwise don't convert it.  And if there are any errors, don't convert it, and return the original path."

So, for example, the function would be more efficient if it just bailed out on checking for the disk_format being "raw".  But then, the question becomes, "why even call the function if the `disk_format` isn't 'raw'"?  i.e. the caller can decide not to call the function if the required disk_format isn't 'raw'.

So taking all this into account, I think:

* the function should be called "convert_image_format_to_raw_if_qcow2()".  That's what it *does*.
* Don't call the function if disk_format isn't going to 'raw'.
* Don't pass 'disk_format' to the function; it's only purpose is to check if it's 'raw' or do nothing.
* At the call site, only call the function if the disk_format is 'raw'.

zaza/openstack/utilities/openstack.py
```diff
@@ -2695,6 +2755,12 @@ def create_image(glance, image_url, image_name, image_cache_dir=None, tags=[],
         logging.info('Cached image found at {} - Skipping download'.format(
             local_path))
 
+    if convert_if_ceph and is_ceph_image_backend():
+        logging.info("Image conversion: Detected ceph backend, forcing"
+                     " use of raw image format")
+        disk_format = 'raw'
+        local_path = ensure_image_format(local_path, disk_format)
```
ajkavanagh, line 2762: Convert this to:

```python
local_path = convert_image_format_to_raw_if_qcow2(local_path)
```

zaza/openstack/utilities/openstack.py
```diff
@@ -2651,10 +2653,68 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
...
+            return local_path
+        return converted_path_str
+    return local_path
+
+
 def create_image(glance, image_url, image_name, image_cache_dir=None, tags=[],
                  properties=None, backend=None, disk_format='qcow2',
                  visibility='public', container_format='bare',
-                 force_import=False):
+                 force_import=False, convert_if_ceph=True):
```
ajkavanagh, line 2717: I don't think this is descriptive enough.  It needs to be something like `convert_image_to_raw_if_ceph_used=True`; i.e. explain the purpose in the name of the variable.

Plus this new parameter needs to be explained in the docstring and why it is being done; please refer to the bug being 'fixed' by this for future readers of the code.

ajkavanagh (changes requested):
This is getting much closer.  The review comments now are about tightening up the code, and explaining what is going on for future readers.  i.e. we're avoiding a major refactor to deal, but we are trying to deal with a performance issue.  We must, please, reference the bug in the code and explain the default behaviour (with the default values in the params) as it *is* slightly counter-intuitive and isn't immediately obvious from calling the function.  When the non-default action is being performed it's important to say *why*.

Thanks!

zaza/openstack/utilities/openstack.py
rodrigogansobarbieri: done, indeed I had forgotten the docstring. I need to log a bug for this so it can be referenced

zaza/openstack/utilities/openstack.py
rodrigogansobarbieri: done

zaza/openstack/utilities/openstack.py
rodrigogansobarbieri: done

ajkavanagh:
@rodrigogansobarbieri please, please, **don't** force push once review comments have been added to a PR; it just erases the history and reviewers can't see what is changing.  When the review is merged it will generally be squashed down to a single commit.  Essentially, by force-pushing you are causing more work for reviewers.

zaza/openstack/charm_tests/glance/tests.py
```diff
@@ -92,7 +92,8 @@ def test_412_image_conversion(self):
                 self.glance_client,
                 image_url,
                 'cirros-test-import',
-                force_import=True)
+                force_import=True,
+                convert_if_ceph=False)
```
ajkavanagh, line 96: I don't think this has been updated?

PR #1082, opened 2023-07-19
Adding bootstrap mysqlrouter action test to mysqlrouter

xtrusia:
Review that tests this: https://review.opendev.org/c/openstack/charm-mysql-router/+/886810

### DIFF
--- a/zaza/openstack/charm_tests/mysql/tests.py
+++ b/zaza/openstack/charm_tests/mysql/tests.py
@@ -1123,7 +1123,7 @@ class MySQLRouterTests(test_utils.OpenStackBaseTest):
         cls.services = ["mysqlrouter"]
         # Config file affected by juju set config change
         cls.conf_file = (
-            "/var/lib/mysql/{}/mysqlrouter.conf"
+            "/var/lib/mysql/{}-mysql-router/mysqlrouter.conf"
             .format(application_name))
 
     def test_910_restart_on_config_change(self):
@@ -1146,11 +1146,19 @@ class MySQLRouterTests(test_utils.OpenStackBaseTest):
             self.services)
         logging.info("Passed restart on changed test.")
 
-    def test_action_bootstrap_mysqlrouter(self):
+    def test_920_bootstrap_mysqlrouter_action(self):
         """Checking bootstrap-mysqlrouter action.
 
         Run the bootstrap-mysqlrouter action
         """
+        # application_name on test is keystone-mysql-router
+        # using self.conf_file introduces error.
+        # instead of changing current self.conf_file,
+        # define one (it could introduce another issue)
+        config_file = (
+            "/var/lib/mysql/{}/mysqlrouter.conf"
+            .format(self.application_name))
+
         logging.info("Starting bootstrap-mysqlrouter test")
 
         # put empty string to conf_file and make it wrong
@@ -1159,7 +1167,7 @@ class MySQLRouterTests(test_utils.OpenStackBaseTest):
                                  "echo '[DEFAULT]\n \
                                         [metadata_cache:[\\w$]+$] \
                                         ' > {}".format(
-                                     self.conf_file))
+                                     config_file))
 
         # verify it is in error state
         for attempt in tenacity.Retrying(
@@ -1206,7 +1214,7 @@ class MySQLRouterTests(test_utils.OpenStackBaseTest):
         logging.info("Getting configuration file")
         recovered = zaza.model.run_on_leader(self.application,
                                              "cat {}".format(
-                                                 self.conf_file))['Stdout']
+                                                 config_file))['Stdout']
 
         # Checking bootstrap action result
         assert "Success" in action.data["results"]["outcome"], (
--- a/zaza/openstack/charm_tests/mysql/tests.py
+++ b/zaza/openstack/charm_tests/mysql/tests.py
@@ -1123,7 +1123,7 @@ class MySQLRouterTests(test_utils.OpenStackBaseTest):
         cls.services = ["mysqlrouter"]
         # Config file affected by juju set config change
         cls.conf_file = (
-            "/var/lib/mysql/{}-mysql-router/mysqlrouter.conf"
+            "/var/lib/mysql/{}/mysqlrouter.conf"
             .format(application_name))
 
     def test_910_restart_on_config_change(self):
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/mysql/tests.py
```diff
@@ -1145,3 +1145,17 @@ def test_910_restart_on_config_change(self):
...
+        Run the bootstrap-mysqlrouter action
+        """
+        logging.info("Running bootstrap-mysqlrouter action")
+        action = zaza.model.run_action_on_leader(
+            self.application,
+            "bootstrap-mysqlrouter",
+            action_params={})
+        assert "Success" in action.data["results"]["outcome"], (
+            "Bootstrap mysqlrouter action failed.")
+        logging.info("Passed bootstrap-mysqlrouter action test.")
```
ajkavanagh, line 1161: This is a good start.  It is a "happy path" test to just verify that the bootstrap action works.

A better test will be to break the mysql-router instance (e.g. overwrite a critical file, ensure that the router is broken), and then run the action and then verify that the mysql-router is actually working.  i.e. this test doesn't actually verify that the action works.

ajkavanagh (changes requested):
Please add a verification test for the bootstrap action that checks that the action does (re-)bootstrap the mysql-router.

zaza/openstack/charm_tests/mysql/tests.py
xtrusia: Thanks I'm going to do this in  this week.

xtrusia:
Please review this case if you have time, thanks a lot!

zaza/openstack/charm_tests/mysql/tests.py
```diff
@@ -1145,3 +1146,64 @@ def test_910_restart_on_config_change(self):
...
+                                        [metadata_cache:[\\w$]+$] \
+                                        ' > {}".format(
+                                     self.conf_file))
+
+        # update status to make the status error
+        logging.info("Run update-status")
+        self.run_update_status_hooks(['keystone-mysql-router/0'])
+
+        while True:
+            time.sleep(5)
```
ajkavanagh, line 1170: This is a pattern that we try to avoid in zaza-openstack-tests.  You can create the same effect using the`tenacity` module.  Take a look at https://github.com/openstack-charmers/zaza-openstack-tests/blob/a07fb4017a2c7a941d64fc6589bfc4ea6c1203b8/zaza/openstack/charm_tests/designate/utils.py#L135 for inspriration.  Essentially, you want something along the lines of an exponential back-off waiting say 2 minutes before bailing out and raising an error to stop the test.  Otherwise, this could hold here for the full 120 minutes (the maximum a test run can last) holding up other tests in the CI system.

zaza/openstack/charm_tests/mysql/tests.py
```diff
@@ -1145,3 +1146,64 @@ def test_910_restart_on_config_change(self):
...
+
+        # run bootstrap-mysqlrouter
+        logging.info("Running bootstrap-mysqlrouter action")
+        action = zaza.model.run_action_on_leader(
+            self.application,
+            "bootstrap-mysqlrouter",
+            action_params={})
+
+        while True:
+            logging.info("Running resolved unit")
```
ajkavanagh, line 1186: As above, please use `tenacity`.

ajkavanagh (changes requested):
This looks good; just an issue with potentially very long test times due to endless `while` loops.

ajkavanagh (dismissed):

zaza/openstack/charm_tests/mysql/tests.py
```diff
@@ -1123,7 +1123,7 @@ def setUpClass(cls, application_name="keystone-mysql-router"):
         cls.services = ["mysqlrouter"]
         # Config file affected by juju set config change
         cls.conf_file = (
-            "/var/lib/mysql/{}-mysql-router/mysqlrouter.conf"
+            "/var/lib/mysql/{}/mysqlrouter.conf"
```
ajkavanagh, line 1126: This change looks curious?  i.e. the router directory is changing from `/var/lib/mysql/keystone-mysql-router/mysqlrouter.con` to `/var/lib/mysql/keystone/mysqlrouter.conf` but I can't find a corresponding change in https://review.opendev.org/c/openstack/charm-mysql-router/+/886810

Please could you help me out, and explain this change?

ajkavanagh (changes requested):

zaza/openstack/charm_tests/mysql/tests.py
xtrusia: oh sorry about this, 
I meant to ask this.. 
While I'm testing zaza test, self.conf_file returned keystone-mysql-router-mysql-router as application name was keystone-mysql-router.

I think it is wrong in the beginning. but I'm not sure 100%.

xtrusia: > This change looks curious? i.e. the router directory is changing from `/var/lib/mysql/keystone-mysql-router/mysqlrouter.con` to `/var/lib/mysql/keystone/mysqlrouter.conf` but I can't find a corresponding change in https://review.opendev.org/c/openstack/charm-mysql-router/+/886810
> 
> Please could you help me out, and explain this change?

Could you please check my comment if it makes sense of I need to change it to original?

Thanks in advance.

ajkavanagh: I'm not clear on *why* you've made this change.  Did the test break?  Also, the associated gerrit review also needs some changes please.  Thanks.

xtrusia: > I'm not clear on _why_ you've made this change. Did the test break? Also, the associated gerrit review also needs some changes please. Thanks.

Yes. my test was failed because self.conf_file returned keystone-mysql-router-mysql-router instead of keystone-mysql-router as I mentioned earlier. Thanks!

xtrusia: fixing tests against to jammy as well

xtrusia: > I'm not clear on _why_ you've made this change. Did the test break? Also, the associated gerrit review also needs some changes please. Thanks.

I mentioned why I changed this one. Could you please review this again? Thanks in advance!

xtrusia:
Could somebody review this one?

Thanks in advance

zaza/openstack/charm_tests/mysql/tests.py
```diff
@@ -1145,3 +1145,75 @@ def test_910_restart_on_config_change(self):
             {}, {},
             self.services)
         logging.info("Passed restart on changed test.")
+
+    def test_action_bootstrap_mysqlrouter(self):
```
ajkavanagh, line 1149: I just noticed that this test name needs a number to fit in with the other numbered tests.  I suggest `test_920_bootstrap_mysqlrouter_action()` to fit in with the other numbered tests.

zaza/openstack/charm_tests/mysql/tests.py
ajkavanagh: I'll quote myself:

> This change looks curious? i.e. the router directory is changing from /var/lib/mysql/keystone-mysql-router/mysqlrouter.con to /var/lib/mysql/keystone/mysqlrouter.conf but I can't find a corresponding change in https://review.opendev.org/c/openstack/charm-mysql-router/+/886810

> Please could you help me out, and explain this change?

I'm confused.  either `test_910_restart_on_config_change` used to break silently or something else has changed, but I can't work out what is different.  If you can't explain *why* this is different then I can't land the change without personally setting it up and running it to see *why* this is different for this patch.  i.e. before this patch the test 'passed' (but may not have been working??), and it also passes with this change, which makes no sense, as one version should break as there are no changes in the associated mysql-router charm patch on gerrit to affect this.

ajkavanagh (changes requested):
I'm still not clear on why the `cls.conf_file` has been changed as without a bit more explanation it's not obvious whether the `test_910_restart_on_config_change` actually does nothing (i.e. isn't a valid test) as the change to the `cls.conf_file` ought to affect the test in some way.

zaza/openstack/charm_tests/mysql/tests.py
xtrusia: Thanks I changed it

zaza/openstack/charm_tests/mysql/tests.py
xtrusia: I removed the change and I put seperated one into specific test
I hope this is proper change.
Thanks.

xtrusia:
> I'm still not clear on why the `cls.conf_file` has been changed as without a bit more explanation it's not obvious whether the `test_910_restart_on_config_change` actually does nothing (i.e. isn't a valid test) as the change to the `cls.conf_file` ought to affect the test in some way.

I changed it back to original and I put another variable for specific test. could you please review it? Thanks!

xtrusia:
@ajkavanagh any chance to review this case? I get it back to original and added new variable for only new test.
Thanks in advance!

PR #1084, opened 2023-07-20
Add vault cachine of secrets on relation test

ajkavanagh:
This specific test is for the certificates relation to ensure that the
data presented to units related to vault have a consistent set of data.

Review that tests this change: https://review.opendev.org/c/openstack/charm-vault/+/883947


freyes (approved):
LGTM

PR #1088, opened 2023-07-26
Add the mapping for Ceph's Reef release (antelope)

lmlg:


coreycb:
@lmlg This might need a commit message to explain why the backport is needed. Will reef be backported prior to mantic and the bobcat cloud archive?

PR #1089, opened 2023-07-26
Add the mapping for Ceph's Reef release (master)

lmlg:


zaza/openstack/utilities/os_versions.py
```diff
@@ -278,6 +278,7 @@
         ('15', 'ussuri'),    # octopus
         ('16', 'victoria'),  # pacific
         ('17', 'yoga'),      # quincy
+        ('18', 'zed'),       # reef
```
freyes, line 281: my understanding is that Reef will be available in ubuntu 23.10 which maps to OpenStack 2023.2 (bobcat)

freyes (changes requested):

zaza/openstack/utilities/os_versions.py
lmlg: yeah, that's true. To be honest, the mapping doesn't matter as much, only that the key ('18') is present. But let's do it right :)

freyes (approved):
LGTM

PR #1090, opened 2023-08-02
Adds ssl config switch test to ceph-dashboard

UtkarshBhatthere:


zaza/openstack/charm_tests/ceph/dashboard/tests.py
```diff
@@ -20,6 +20,8 @@
 import requests
 import tenacity
 import uuid
+import trustme
+from base64 import b64encode
```
sabaini, line 24: Nit: I realize this wasn't very tidy before but it would be nice to get import ordering conform to PEP8, i.e.:
- Standard library imports.
- Related third party imports.
- Local application/library specific imports.

cf. https://peps.python.org/pep-0008/#imports

zaza/openstack/charm_tests/ceph/dashboard/tests.py
```diff
@@ -309,3 +311,115 @@ def test_saml(self):
...
+            {'ssl_cert': "", 'ssl_key': "", 'ssl_ca': ""}
+        )
+
+        # Wait for status to clear
+        assert_state = self._get_wait_for_dashboard_assert_state(
+            "active", "Unit is ready"
+        )
+        zaza.model.wait_for_application_states(
+            states=assert_state, timeout=500
+        )
```
sabaini, line 425: I'm wondering if it would make sense to run another round of self._run_request_get() here in order to verify SSL works correctly again? (e.g. put the certificate verification section above from L383-L397 into a helper method and call it two times)

sabaini (changes requested):
Excellent, lgtm in general, some small nits/questions inline -- thanks!

zaza/openstack/charm_tests/ceph/dashboard/tests.py
UtkarshBhatthere: ack

zaza/openstack/charm_tests/ceph/dashboard/tests.py
UtkarshBhatthere: ack

zaza/openstack/charm_tests/ceilometer/tests.py
```diff
@@ -126,7 +126,7 @@ def test_900_restart_on_config_change(self):
         current_value = openstack_utils.get_application_config_option(
             self.application_name, config_name
         )
-        assert type(current_value) == bool
+        assert type(current_value) is bool
```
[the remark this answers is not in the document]
UtkarshBhatthere: ack, that has been removed from this test.

sabaini (approved):
Thanks, lgtm!

ajkavanagh (changes requested):
Just putting a hold on this; there's some code strangeness that I'd like to comment on.

zaza/openstack/charm_tests/ceph/dashboard/tests.py
```diff
@@ -309,3 +302,133 @@ def test_saml(self):
...
+
+    def _get_dashboard_hostnames_sans(self):
+        """Get a tuple of hostnames for the units of the application."""
+        ret_val = ['ceph-dashboard']  # Include hostname in san as well.
+        for unit in zaza.model.get_units('ceph-mon'):
+            addr = zaza.model.get_unit_public_address(unit)
+            if addr:
+                ret_val.append(addr)
+        return tuple(ret_val)
+
```
ajkavanagh, line 356: Okay; so this function's return value is *very* odd to me.  It appears to be returning a variable-length `tuple()`, and I'm really struggling to work out the justification for this.  AFAICT, this function.

1. Creates a variable length list via `ret_val`
2. Consumes the variable length list int a `tuple` (which essentially takes an iterator), thus returning the tuple and *throwing away* the list structure.
3. Is immediately used by `test_006_charm_config_ssl` which using `*var` which consumes an iterator, which obviously the `tuple` provides, and then throws the tuple away.

So why create the `tuple`?

If you are concerned about immutability, then just make `_get_dashboard_hosnames_sans` a generator (although that can only be used on, or just return the list.

UtkarshBhatthere: Thanks @ajkavanagh, As per the suggestion I have switched to using a generator.

ajkavanagh (approved):

PR #1091, opened 2023-08-08
Fix pep8 error in ceilometer

xtrusia:
Fix below pep8 error message
zaza/openstack/charm_tests/ceilometer/tests.py:129:16: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinst ance()`


hemanthnakkina (approved):
LGTM

PR #1096, opened 2023-08-22
Convert images to raw if ceph image backend (#1078)

rodrigogansobarbieri:
Convert images to raw if ceph image backend

We are currently uploading qcow2 images, and
Nova is converting them to raw when running
the tests, sometimes timing out the tests.

With this change we are pre-converting the
images and uploading them as raw, so Nova
does not have to convert them.

(cherry picked from commit e0498d6a1691fa1ef661ab753d4936b7f138e6ce)


freyes (approved):
LGTM

PR #1100, opened 2023-08-24
Pin zaza to stable/antelope for the unit tests.

ajkavanagh:
Also fix a (newish) pep8 error.


freyes (approved):
LGTM.

this being a change only to testing infrastructure I will merge it with a single +2.

PR #1118, opened 2023-09-06
Gate quorum queue tests on whether the charm supports them

gnuoy:


gnuoy:
Testing here: https://github.com/openstack-charmers/charm-rabbitmq-k8s/pull/9

PR #1138, opened 2023-09-27
swift-proxy: wait for model to settle after running action [Antelope backport]

freyes:
Let the units run all the hooks before attempting to check if the ring is synced, because the action set-weight triggers the hook swift-storage-relation-changed in the swift-storage units which it's the one in charge of getting the new ring data.

(cherry picked from commit 0ca305f6c85a8a55091007041c2df3767027e73e)


freyes:
keeping it in draft until I get confirmation from this gerrit patch ci run:  https://review.opendev.org/c/openstack/charm-swift-proxy/+/891549

freyes:
Backport of https://github.com/openstack-charmers/zaza-openstack-tests/pull/1139

PR #1139, opened 2023-09-27
swift-proxy: wait for model to settle after running action

freyes:
Let the units run all the hooks before attempting to check if the ring is synced, because the action set-weight triggers the hook swift-storage-relation-changed in the swift-storage units which it's the one in charge of getting the new ring data.


freyes:
this PR is being exercised by https://review.opendev.org/c/openstack/charm-swift-proxy/+/896723

ajkavanagh (approved):

PR #1141, opened 2023-09-28
Remove hardcoded nova-compute app name and add ceph keys test

rodrigogansobarbieri:
WIP

### DIFF
--- a/zaza/openstack/charm_tests/nova/tests.py
+++ b/zaza/openstack/charm_tests/nova/tests.py
@@ -19,6 +19,7 @@
 import json
 import logging
 import os
+import re
 import tempfile
 import tenacity
 import unittest
@@ -127,7 +128,9 @@ class VTPMGuestCreateTest(test_utils.OpenStackBaseTest):
                                     password=password, privkey=privkey,
                                     verify=check_tpm)
 
-    @test_utils.skipUntilVersion('nova-compute', 'nova-common', '3:23.0.0')
+    @test_utils.skipUntilVersion(
+        openstack_utils.get_app_names_for_charm('nova-compute')[0],
+        'nova-common', '3:23.0.0')
     def test_launch_vtpm_1_2_instance(self):
         """Launch an instance using TPM 1.2."""
         self.RESOURCE_PREFIX = 'zaza-nova'
@@ -138,7 +141,9 @@ class VTPMGuestCreateTest(test_utils.OpenStackBaseTest):
         # Note: TPM 1.2 presents tpm0 as a device
         self._check_tpm_device(instance, '/dev/tpm0')
 
-    @test_utils.skipUntilVersion('nova-compute', 'nova-common', '3:23.0.0')
+    @test_utils.skipUntilVersion(
+        openstack_utils.get_app_names_for_charm('nova-compute')[0],
+        'nova-common', '3:23.0.0')
     def test_launch_vtpm_2_instance(self):
         """Launch an instance using TPM 2.0."""
         self.RESOURCE_PREFIX = 'zaza-nova'
@@ -265,8 +270,9 @@ class CloudActions(test_utils.OpenStackBaseTest):
 
     def test_940_enable_disable_actions(self):
         """Test disable/enable actions on nova-compute units."""
-        nova_units = zaza.model.get_units('nova-compute',
-                                          model_name=self.model_name)
+        nova_units = zaza.model.get_units(
+            self.application_name,
+            model_name=self.model_name)
 
         # Check that nova-compute services are enabled before testing
         for service in self.nova_client.services.list(binary='nova-compute'):
@@ -307,8 +313,9 @@ class CloudActions(test_utils.OpenStackBaseTest):
             instances = result.data.get('results', {}).get('instance-count')
             self.assertEqual(instances, str(expect_count))
 
-        nova_unit = zaza.model.get_units('nova-compute',
-                                         model_name=self.model_name)[0]
+        nova_unit = zaza.model.get_units(
+            self.application_name,
+            model_name=self.model_name)[0]
 
         check_instance_count(0, nova_unit.entity_id)
 
@@ -345,8 +352,9 @@ class CloudActions(test_utils.OpenStackBaseTest):
                 sleep_timeout = 10
             return False
 
-        all_units = zaza.model.get_units('nova-compute',
-                                         model_name=self.model_name)
+        all_units = zaza.model.get_units(
+            self.application_name,
+            model_name=self.model_name)
 
         unit_to_remove = all_units[0]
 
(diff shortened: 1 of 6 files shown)


zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -470,10 +481,50 @@ def test_901_pause_resume(self):
         with self.pause_resume(['nova-compute']):
             logging.info("Testing pause resume")
 
+    def test_904_test_ceph_keys(self):
+        """Run pause and resume tests.
```
rodrigogansobarbieri, line 485: docstring needs to be a proper one instead of this copy-pasted crap

zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -470,10 +481,50 @@ def test_901_pause_resume(self):
...
+                            self.assertEqual(2, len(result))
+                            self.assertEqual(result[0], key_app_name)
+                            for k, v in key_dict.items():
+                                if k == result[0]:
+                                    self.assertEqual(v, result[1])
+                                else:
+                                    self.assertNotEqual(v, result[1])
+                            key_dict[result[0]] = result[1]
+
+        check_keyring([{self.application_name: [self.application_name]}])
```
rodrigogansobarbieri, line 521: this was coded for stable/yoga where it does not have the partial fix that is on master. This needs to be adjust to run against the code on master.

rodrigogansobarbieri (commented):
needs updates

zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -470,10 +481,50 @@ def test_901_pause_resume(self):
...
+                            app_name, model_name=self.model_name):
+                        for key_app_name in key_list:
+                            keyring_file = (
+                                '/etc/ceph/ceph.client.{}.keyring'.format(
+                                    key_app_name))
+                            data = str(generic_utils.get_file_contents(
+                                unit, keyring_file))
+
+                            result = regex.findall(data)
+                            self.assertEqual(2, len(result))
```
brianphaley, line 512: nit: might be more obvious if you pull the key/value out here:

app_name, app_value = result

But just a nit

brianphaley (commented):
Just had a question on the test you added, otherwise looks good.

zaza/openstack/charm_tests/vault/setup.py
```diff
@@ -187,12 +188,13 @@ def auto_initialize(cacert=None, validation_application='keystone', wait=True,
...
         # bug #1826382
+        nova_compute_app_name = get_app_names_for_charm('nova-compute')[0]
         cmd_map = {
             'nova-cloud-controller': ('systemctl restart '
                                       'nova-scheduler nova-conductor'),
-            'nova-compute': 'systemctl restart nova-compute',
+            nova_compute_app_name: 'systemctl restart nova-compute',
         }
-        for app in ('nova-compute', 'nova-cloud-controller',):
+        for app in (nova_compute_app_name, 'nova-cloud-controller',):
```
nicolasbock, line 197: I might be misunderstanding the intent here, but if you anticipate that there
might be several apps that use the `nova-compute` charm, wouldn't you want to
restart services on all of those? However, you only store the first application
in line 191.

zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -372,7 +380,7 @@ def wait_for_nova_compute_count(expected_count):
 
         # Wait for nova-compute service to be removed from the
         # nova-cloud-controller
-        if not wait_for_nova_compute_count(0):
+        if not wait_for_nova_compute_count(service_count-1):
```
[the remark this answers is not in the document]
dosaboy: also need whitespace around the '-'

zaza/openstack/utilities/openstack.py
```diff
@@ -2693,6 +2694,28 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
...
+    """
+    status = juju_utils.get_full_juju_status(model_name=model_name)
+    result = []
+    try:
+        for app_name, app_data in status['applications'].items():
+            try:
+                if charm_name == app_data['charm-name']:
+                    result.append(app_name)
+            except KeyError:
+                # Older juju versions may not have the field "charm-name"
```
dosaboy, line 2711: I see references to "charm-name" as far back as juju 1.25 and it is definitely in 2.3, maybe this is overkill?

zaza/openstack/charm_tests/nova/tests.py
rodrigogansobarbieri: I rewrote that logic in the new version I am going to push soon

zaza/openstack/charm_tests/nova/tests.py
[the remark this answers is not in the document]
rodrigogansobarbieri: that is "future code", removing hardcoded values without changing the behavior, and keeping it ready to support more than 1 unit

zaza/openstack/charm_tests/nova/tests.py
rodrigogansobarbieri: It would raise value error if the length is different. The assert validation is less imprecise as an error message.

zaza/openstack/utilities/openstack.py
```diff
@@ -2693,6 +2694,28 @@ def upload_image_to_glance(glance, local_path, image_name, disk_format='qcow2',
...
+                if charm_name in app_data['charm']:
```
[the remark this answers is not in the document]
rodrigogansobarbieri: the value here could be 'nova-compute-kvm' for example, so we are checking if the 'nova-compute' string  is contained in app_data['charm'] string

zaza/openstack/charm_tests/vault/setup.py
rodrigogansobarbieri: yes, that was partial future-proofing code, it is easy to make it fully future proof, making those changes now. I was initially just fixing the error where it was hardcoded.

zaza/openstack/utilities/openstack.py
rodrigogansobarbieri: I have seen juju status json without this field while coding probelius, then a later juju update introduced this field consistently. A bit overkill probably, but harmless I suppose.

brianphaley (approved):
lgtm, will see what others think

zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -492,8 +570,9 @@ class NovaComputeActionTest(test_utils.OpenStackBaseTest):
 
     def test_virsh_audit_action(self):
         """Test virsh-audit action."""
-        for unit in zaza.model.get_units('nova-compute',
-                                         model_name=self.model_name):
+        for unit in zaza.model.get_units(
+                self.application_name,
```
dosaboy, line 574: are you sure that self.application_name is being set to the name of the application in the same way as get_app_names_for_charm()?

rodrigogansobarbieri: it is somewhat similar. The base class code that assigns self.application_name checks if each app starts with the charm_name declared in tests.yaml, and it breaks when it finds the first one. We could remove the break and go on creating a "self.application_nameS" variable and we ultimately have the same as get_app_names_for_charm. Ultimately I'd prefer if it uses "in (contains)" instead of startswith though.

rodrigogansobarbieri:
Closed the PR as we don't intend to refactor this at the moment

PR #1144, opened 2023-10-01
tempest: Make concurrency configurable with a sane default.

fnordahl:
Tempest will by default run with a concurrency matching the number of cores on the test runner.

When running on a workstation, it is likely that the default concurrency will be too high for the scale of deployed workload.


ajkavanagh (approved):

coreycb:
Thanks Frode! That might solve some intermittent issues I've seen.

PR #1148, opened 2023-10-05
Allow 401s when checking endpoints

gnuoy:
Some endpoints (heat) return 401s if you try and access them without authenticating. Given the ep has been able to verify whether you are authenticated a 401 is ok for a rudementary health check


gnuoy:
Testing here https://review.opendev.org/c/openstack/charm-heat-k8s/+/897385

hemanthnakkina (approved):
LGTM

PR #1153, opened 2023-10-13
Add vault caching of secrets on relation test

xtrusia:
This specific test is for the certificates relation to ensure that the data presented to units related to vault have a consistent set of data.

(cherry picked from commit 752e643c3345ef61eca458968aa14549dd618d15)


ajkavanagh (changes requested):
This patch isn't in the stable/antelope branch (yet), so can't be merged here until it is merged in stable/antelope.

PR #1163, opened 2023-10-31
Add new heat_plugin section to tempest.conf

coreycb:
The heat tepest tests have moved out of tempest proper into a plugin. This updates the tempest.conf accordingly.

### DIFF
--- a/zaza/openstack/charm_tests/tempest/templates/tempest_v3.j2
+++ b/zaza/openstack/charm_tests/tempest/templates/tempest_v3.j2
@@ -87,19 +87,16 @@ project_domain_name = demoDomain
 admin_username = {{ admin_username }}
 admin_password = {{ admin_password }}
 admin_project_name = {{ admin_project_name }}
-admin_user_domain_name = admin_domain
-admin_project_domain_name = admin_domain
+admin_domain_name = {{ admin_domain_name }}
 auth_url = {{ proto }}://{{ keystone }}:5000//v3
 auth_version = 3
 disable_ssl_certificate_validation = true
-ca_file=/home/ubuntu/.cache/openstack-root-ca.crt
 region = RegionOne
 instance_type = m1.small
 minimal_instance_type = m1.small
 image_ref = {{ image_id }}
 minimal_image_ref = {{ image_id }}
-skip_functional_test_list = LoadBalancerv2Test,NotificationTest,UpdateStackTest.test_stack_update_with_replacing_userdata,CancelUpdateTest.test_cancel_update_server_with_port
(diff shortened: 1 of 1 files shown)


coreycb:
The following makes use of this: https://github.com/openstack-charmers/charmed-openstack-tester/pull/126

zaza/openstack/charm_tests/tempest/templates/tempest_v3.j2
```diff
@@ -77,10 +77,29 @@ port_security = true
...
+user_domain_name = demoDomain
+project_domain_name = demoDomain
+admin_username = {{ admin_username }}
+admin_password = {{ admin_password }}
+admin_project_name = {{ admin_project_name }}
+admin_user_domain_name = admin_domain
+admin_project_domain_name = admin_domain
+auth_url = {{ proto }}://{{ keystone }}:5000//v3
+auth_version = 3
+disable_ssl_certificate_validation = true
```
freyes, line 94: if we are disabling cert validation, we can probably get away with not setting the `ca_file` config key, it will allow us to not rely where the root CA is.

freyes: we probably need to set this to true too - https://github.com/openstack/heat-tempest-plugin/blob/master/heat_tempest_plugin/config.py#L170

freyes (changes requested):

zaza/openstack/charm_tests/tempest/templates/tempest_v3.j2
```diff
@@ -77,10 +77,29 @@ port_security = true
...
+username = demo
+password = password
+project_name = demoProject
+user_domain_name = demoDomain
+project_domain_name = demoDomain
+admin_username = {{ admin_username }}
+admin_password = {{ admin_password }}
+admin_project_name = {{ admin_project_name }}
+admin_user_domain_name = admin_domain
+admin_project_domain_name = admin_domain
```
freyes, line 91: shouldn't we use `{{ admin_domain_name }}`  here instead?

freyes (changes requested):

zaza/openstack/charm_tests/tempest/templates/tempest_v3.j2
coreycb: Thanks for taking a look. I've got a new version with cleanup incoming. I'd like to enable the certs eventually once the old bundles that don't use vault are no longer used.

zaza/openstack/charm_tests/tempest/templates/tempest_v3.j2
coreycb: yep

freyes (approved):
LGTM

PR #1167, opened 2023-11-06
Add tempest confirmation for octavia tests

coreycb:
This sets new tempest config options for octavia, including the setting of log_user_roles=False, which allows tests to avoid the optional list_role_assignments call that was resulting in 403 Forbidden for all octavia tests.


freyes (approved):
This is a great fix, Corey, appreciated.

PR #1172, opened 2023-12-01
Make wait_for_endpoints more robust.

gnuoy:
As seen in Bug #2045206 wait_for_endpoints can fails if keystone is in a transient state. This change puts tenacity around the keystone client setup to make it more resilient.

Closes-Bug: 2045206

### DIFF
--- a/zaza/openstack/charm_tests/keystone/setup.py
+++ b/zaza/openstack/charm_tests/keystone/setup.py
@@ -192,6 +192,26 @@ def wait_for_url(url, ok_codes=None):
             assert r.status_code in ok_codes
 
 
+def wait_for_client():
+    """Wait for client to be returned successfully.
+
+    If keystone is still in a transient state then it may take a few retries
+    before a client is returned
+    """
+    for attempt in tenacity.Retrying(
+            stop=tenacity.stop_after_attempt(10),
+            wait=tenacity.wait_exponential(
+                multiplier=1, min=2, max=60)):
+        with attempt:
+            overcloud_auth = openstack_utils.get_overcloud_auth()
+            wait_for_url(overcloud_auth['OS_AUTH_URL'])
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/keystone/setup.py
```diff
@@ -192,17 +192,34 @@ def wait_for_url(url, ok_codes=None):
             assert r.status_code in ok_codes
 
 
+def wait_for_client():
```
freyes, line 195: nitpick: the function's name is a bit strange ... it's waiting, but also returning a value ... the "waiting" seems to be irrelevant from the caller's perspective, why not call it "get_keystone_client"?, or maybe re-use get_keystone_client()[0] and wrap that one with a retry() decorator?

[0] https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/utilities/openstack.py#L616

freyes (changes requested):

zaza/openstack/charm_tests/keystone/setup.py
ajkavanagh: I agree that the function's name is wrong.  I don't think that `get_keystone_client()` can be used as it seems more basic that this function; i.e. it's using overcloud auth rather than openrc.

I think this function ought to be in `z.o.utilities.openstack.py` though along with the other keystone client functions which would also open a space for refactoring at a future point.

Finally, the  `wait_for_url()` function is also really generic and probably could also live in a utility module; the `wait_for_all_endpoints()` a bit more arguable as to where it should go.  It's specific enother to be testing keystone ... but also generic enough that other tests might also want to wait on the endpoints being available as a utility.

ajkavanagh (approved):
Thanks for those changes; this definitely makes it more generic and enables more of a chance of refactoring.  thanks!

PR #1174, opened 2023-12-07
utilities/openstack: Fix handling of unit.run for CA check

fnordahl:
The `async_block_until_ca_exists` function makes direct use of the libjuju APIs which changed in Juju 3.x.

Use the action normalise helper from the Zaza model module to reconcile.


zaza/openstack/charm_tests/vault/utils.py
```diff
@@ -474,18 +474,23 @@ def run_charm_authorize(token):
...
     """Retrieve CSR from vault.
 
     Run vault charm action to retrieve CSR from vault.
 
+    :param force: Force regeneration of intermediate ca.
+    :type force: Optional[bool]
     :returns: Action object
     :rtype: juju.action.Action
     """
+    action_params={}
```
freyes, line 487: zaza/openstack/charm_tests/vault/utils.py:487:18: E225 missing whitespace around operator

freyes (changes requested):

PR #1176, opened 2023-12-22
Add Ceph RGW Cloud Sync integration test

ionutbalutoiu:
Add integration test for the Cloud Sync feature proposed to the upstream `ceph-radosgw` Juju charm:
https://review.opendev.org/c/openstack/charm-ceph-radosgw/+/899139


zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -1038,7 +1095,100 @@ def test_003_object_storage_and_secondary_block(self):
...
+        self.wait_for_status(self.cloud_sync_rgw_app, is_primary=False)
+
+        # Validate that synced data was removed from the S3 targets.
+        logging.info('Verifying that data was deleted on the S3 targets')
+        with self.assertRaises(botocore.exceptions.ClientError):
+            default_s3_client.Object(default_container_name, obj_name).get()
+        with self.assertRaises(botocore.exceptions.ClientError):
+            dev_s3_client.Object(dev_container_name, obj_name).get()
+
+    def test_100_migration_and_multisite_failover(self):
```
UtkarshBhatthere, line 1191: Is there any particular reason we renamed the 004 test ? These numbers define the order in which the tests are executed hence I would prefer that the existing order is not changed unless really needed.

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -657,6 +657,9 @@ class CephRGWTest(test_utils.BaseCharmTest):
     primary_rgw_unit = 'ceph-radosgw/0'
     secondary_rgw_app = 'secondary-ceph-radosgw'
     secondary_rgw_unit = 'secondary-ceph-radosgw/0'
+    cloud_sync_rgw_app = 'cloud-sync-ceph-radosgw'
+    cloud_sync_default_s3_app = 's3-default'
+    cloud_sync_dev_s3_app = 's3-dev'
```
UtkarshBhatthere, line 662: Need some comments around why two S3 apps are required for this test.

UtkarshBhatthere (changes requested):
The test structure lgtm, few questions.

zaza/openstack/charm_tests/ceph/tests.py
ionutbalutoiu: We want the `migration_and_multisite_failover` test case to be executed last, that's why I renamed it.

I noticed the following comment:
https://github.com/openstack-charmers/zaza-openstack-tests/blob/b0d0012a9559cf1abc8a470eba9379f3e4f6c3d1/zaza/openstack/charm_tests/ceph/tests.py#L649-L652

Therefore, I assumed that due to the unsupported scenario, we want that test to be executed at the end.

Renaming it the prefix from `test_004` to `test_100` would allow us to add more tests between `test_003` and `test_100`. This is for future-proofing, as another test case is added in the PR https://github.com/openstack-charmers/zaza-openstack-tests/pull/1193.

zaza/openstack/charm_tests/ceph/tests.py
ionutbalutoiu: I will add more comments here.

zaza/openstack/charm_tests/ceph/tests.py
UtkarshBhatthere: Ack, that does sound OK to me.

zaza/openstack/charm_tests/ceph/tests.py
ionutbalutoiu: Added comments to clarify the purpose of these S3 apps in the Zaza cloud-sync tests bundle.

See the added comments [here](https://github.com/openstack-charmers/zaza-openstack-tests/compare/4582566842a449dd0204df9af5e39fe02fb3ef79..4c2fea973bd94ba223df21ba396235cb599b38a4).

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -685,6 +691,15 @@ def multisite(self):
         except KeyError:
             return False
 
+    @property
+    def cloud_sync(self):
+        """Determine whether Ceph cloud sync is used."""
```
UtkarshBhatthere, line 696: NIT: I think this doc string is not 100% accurate. Maybe something like, `Determine whether ceph cloud sync application is present` is clearer ?

ionutbalutoiu: Yep, I agree. I [changed this](https://github.com/openstack-charmers/zaza-openstack-tests/compare/a474f61d4d3ec30b73091276033482e0e684ef8b..59876270cfe372bdf744b6eed9557d1763a9a251).

PR #1177, opened 2024-01-02
Make Ceph tests self-contained.

lmlg:
This PR implements the needed changes so that the Ceph charms can be run without a full Openstack environment. In particular, the RBD mirror charm needed tweaking so as to not depend on Cinder, using pools of its own creating, and skipping some tests when needed.

### DIFF
--- a/zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
+++ b/zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
@@ -156,14 +156,15 @@ def create_cinder_volume(cinder, name='zaza', image_id=None, type_id=None):
 
 
 def setup_rbd_mirror():
-    """Setup an RBD pool in case Cinder isn't present."""
+    """Set up an RBD pool in case Cinder isn't present."""
     zaza.model.run_action_on_leader(
         'ceph-mon',
         'create-pool',
         action_params={
             'name': 'zaza-boot',
             'app-name': 'rbd',
-    })
+        }
+    )
     zaza.model.run_action_on_leader(
         'ceph-rbd-mirror',
         'refresh-pools',
@@ -191,6 +192,7 @@ class CephRBDMirrorBase(test_utils.BaseCharmTest):
             cls.with_cinder = False
 
     def check_cinder_present(self, caller):
+        """Skip a test if Cinder isn't present."""
         if not self.with_cinder:
             raise unittest.SkipTest('Skipping %s due to lack of Cinder'
                                     % caller)
@@ -244,7 +246,8 @@ class CephRBDMirrorBase(test_utils.BaseCharmTest):
         """
         site_a_pools, site_b_pools = self.get_pools()
         if (self.with_cinder and
-            get_cinder_rbd_mirroring_mode(self.cinder_ceph_app_name) == 'image'):
+                get_cinder_rbd_mirroring_mode(self.cinder_ceph_app_name) ==
+                'image'):
             site_a_pools.remove(self.cinder_ceph_app_name)
             site_b_pools.remove(self.cinder_ceph_app_name)
 
@@ -798,8 +801,8 @@ class CephRBDMirrorDisasterFailoverTest(CephRBDMirrorBase):
             })
         self.assertEqual(int(result.results['Code']), 0)
 
-        # The action may not show up as 'failed' if there are no pools that needed
-        # to be promoted.
+        # The action may not show up as 'failed' if there are no pools that
+        # needed to be promoted.
         # self.assertEqual(result.status, 'failed')
 
         # Retry to promote site-b using the 'force' Juju action parameter.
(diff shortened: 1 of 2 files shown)


zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -166,6 +184,18 @@ def setUpClass(cls):
...
         cls.site_b_app_suffix = '-b'
+        # Test if we have the needed Openstack applications.
+        try:
+            zaza.model.get_application(cls.cinder_ceph_app_name)
+            cls.with_cinder = True
+        except Exception:
+            cls.with_cinder = False
+
+    def check_cinder_present(self, caller):
+        """Skip a test if Cinder isn't present."""
```
ajkavanagh, line 195: I think that this would be better named something like `skip_test_if_cinder_is_not_present()` as that is what it actually does.  At the moment, it looks like a predicate that has an unexpected side-effect.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -166,6 +184,18 @@ def setUpClass(cls):
         # get ready for multi-model Zaza
         cls.site_a_model = cls.site_b_model = zaza.model.get_juju_model()
         cls.site_b_app_suffix = '-b'
+        # Test if we have the needed Openstack applications.
```
ajkavanagh, line 187: This comment isn't strictly accurate.  What the following lines do is to check for specifically the 'cinder-ceph' zaza application name.  Please consider dropping the comment.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -166,6 +184,18 @@ def setUpClass(cls):
...
+        try:
+            zaza.model.get_application(cls.cinder_ceph_app_name)
+            cls.with_cinder = True
+        except Exception:
```
ajkavanagh, line 191: Please tighten up the exception. Which exception are you expecting when the application name is not present (`KeyError` or a custom exception).  This will silently set `cls.with_cinder` to `False` on a network error to the controller for example, and not fail in the test with a suitable error message.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
```diff
@@ -166,6 +184,18 @@ def setUpClass(cls):
         # get ready for multi-model Zaza
         cls.site_a_model = cls.site_b_model = zaza.model.get_juju_model()
         cls.site_b_app_suffix = '-b'
+        # Test if we have the needed Openstack applications.
+        try:
+            zaza.model.get_application(cls.cinder_ceph_app_name)
+            cls.with_cinder = True
```
ajkavanagh, line 190: This is caching whether the `cinder-ceph` application is deployed in the model at the beginning of the tests.  As caching is one of the hard problems, perhaps what's really needed is a predicate function that can be called at each test.

e.g. if some future test decided to test what happens when cinder-ceph is removed from the model (I don't why!) then this caching would break.  I think we should all develop more a mindset around avoiding caching where possible unless it really is a performance or memory issue.

ajkavanagh (changes requested):
Please see my inline comments.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
lmlg: Done.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
lmlg: Modified the comment to fit what it does.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
lmlg: Done.

zaza/openstack/charm_tests/ceph/rbd_mirror/tests.py
lmlg: I understand where you're coming from. In the context of this PR, this caching is only looked at the beginning of a couple of tests to skip them in an application is not present. It's pretty harmless in my opinion, but I can accommodate further changes. For example, I can document how it's supposed to be used or replace the caching for a call that looks for the application on the model on each run. What do you think?

ajkavanagh: Yes, I think documenting it in the code would provide enough sign-posting if you want to keep the cached value; however, if you can avoid caching altogether then it would make the code more robust from a maintenance perspective.  If there are performance issues, or other undesirable side-effects, then caching is acceptable.

lmlg: Done, there's no longer caching.

ajkavanagh (approved):
LGTM; thanks for the changes and simplifying the code.

PR #1179, opened 2024-01-04
Watcher yoga

wolsen:


wolsen:
Exercised in https://review.opendev.org/c/openstack/charm-watcher/+/905265

PR #1185, opened 2024-02-19
Set tempdir to a directory under ~/

freyes:
When juju 3.x is used (installed from a snap in strict confinement) the temporary directory of the system (e.g. /tmp) can't be read by 'juju', this is used by zaza-openstack-tests to write testing policy files among other transitory files that are passed to juju's CLI.

This change moves Python's default temporary file to ~/tmp unless overriden by the environment variable TEST_TMPDIR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,42 +29,43 @@ git+https://github.com/openstack-charmers/zaza-openstack-tests.git#egg=zaza.open
 
 Zaza-openstack-test uses environment variables to configure the tests:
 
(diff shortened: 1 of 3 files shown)


freyes:
Testing this change at the gate with this gerrit change - https://review.opendev.org/c/openstack/charm-keystone/+/908183

freyes:
Keeping this as draft until the Keystone patch passes the CI

ajkavanagh (approved):

PR #1187, opened 2024-02-28
add test for virtual hosted bucket

shundezhang:
Add integration test for the virtual hosted bucket feature proposed to the upstream ceph-radosgw Juju charm:
https://review.opendev.org/c/openstack/charm-ceph-radosgw/+/904554

### DIFF
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -900,6 +900,40 @@ class CephRGWTest(test_utils.BaseCharmTest):
             '--rgw-zonegroup=default'
         )
 
+    def enable_virtual_hosted_bucket(self):
+        """Enable virtual hosted bucket on primary rgw app."""
+
+        zaza_model.set_application_config(
+            self.primary_rgw_app,
+            {
+                'virtual-hosted-bucket-enabled': "true"
+            }
+        )
+
+    def set_os_public_hostname(self):
+        """Set os-public-hostname on primary rgw app."""
+
+        zaza_model.set_application_config(
+            self.primary_rgw_app,
+            {
+                'os-public-hostname': "rgw.example.com",
+            }
+        )
+
+    def clean_virtual_hosted_bucket(self):
+        """Clear virtual hosted bucket on primary app.
+
+        :param app_name: App for which config values are to be cleared
+        :type app_name: str
+        """
+        zaza_model.set_application_config(
+            self.primary_rgw_app,
+            {
+                'os-public-hostname': "",
+                'virtual-hosted-bucket-enabled': "false"
+            }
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -967,6 +975,17 @@ def test_003_object_storage_and_secondary_block(self):
         content = primary_object_one.get()['Body'].read().decode('UTF-8')
         self.assertEqual(content, obj_data)
 
+        if self.virtual_hosted_bucket_enabled:
+            # test virtual hosted bucket
+            logging.info('Testing virtual hosted bucket')
+            primary_client.Bucket(container_name).Acl().put(ACL='public-read')
+            primary_client.Object(container_name,obj_name).Acl().put(ACL='public-read')
```
sabaini, line 982: One doubt around this. The above makes the test container publically available. Later on the test checks access for the test container using test credentials. Given that the container is now publically available, would the test credentials even be considered then?

I feel it would be preferable to isolate this test into a test method of it's own (and not share a container name with the other tests), also given that this test method is awfully long already

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -967,6 +975,17 @@ def test_003_object_storage_and_secondary_block(self):
...
+            public_hostname = zaza_model.get_application_config(self.primary_rgw_app)["os-public-hostname"]["value"]
+            url = primary_endpoint + "/" + obj_name
+            headers = {'host': container_name + "." + public_hostname}
```
sabaini, line 985: Minor style nit: suggest to use f-strings for readability

```
url = f"{primary_endpoint}/{obj_name}"
headers = {'host': f"{container_name}.{public_hostname}"}
```

sabaini (changes requested):
Thank you for this test! Two minor comments inline

zaza/openstack/charm_tests/ceph/tests.py
shundezhang: this is changed

zaza/openstack/charm_tests/ceph/tests.py
shundezhang: This test is moved to test_005_virtual_hosted_bucket function

sabaini (approved):

ajkavanagh:
@shundezhang hi; it would be really useful if you **didn't** force push, please.  Each time you do, you erase the previous commits which makes it really hard to see what the differences are between your commits, which also slows down the review process.  We can squash small edits down as part of the merge.  Thanks.

shundezhang:
@ajkavanagh sure will do. Actually it is easier for me to not force push.

PR #1190, opened 2024-03-06
[yoga] Add test to check ceph keys

rodrigogansobarbieri:
For https://review.opendev.org/897549

(cherry picked from commit 912f33e0712034648d393874f14a27cf168bf68b) (cherry picked from commit e242121880ffa1535ad5ed89036a09cc3a6806af) (cherry picked from commit d1a5ab463e3645ddf49818e23ccdea961721c187) (cherry picked from commit e23e14b3607101ade5dec81216d032579c5282ca)

### DIFF
--- a/zaza/openstack/charm_tests/nova/tests.py
+++ b/zaza/openstack/charm_tests/nova/tests.py
@@ -19,6 +19,7 @@
 import json
 import logging
 import os
+import re
 import tempfile
 import unittest
 import urllib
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/nova/tests.py
```diff
@@ -468,6 +469,63 @@ def test_901_pause_resume(self):
...
+
+        for unit in zaza.model.get_units(
+                self.application_name, model_name=self.model_name):
+
+            # old key
+            check_keyring(self.application_name)
+            # new key
+            check_keyring('nova-compute-ceph-auth-c91ce26f')
+
     def test_930_check_virsh_default_network(self):
```
dosaboy, line 529: check_keyring is referring to variable "unit" that is from global space. It should be passed into check_keyring() as an argument.

rodrigogansobarbieri: omg yes, that slipped by, we will need to fix it in master and bobcat as well

rodrigogansobarbieri: actually, I just tested the logic and that coding problem does not affect the result at all

dosaboy: yep no worries, can just fix on master if we get time

PR #1191, opened 2024-03-12
Fix 'NoneType' object has no attribute 'get'

freyes:
There are situations where the action object returned by libjuju may not have the 'results' attribute which leads to using the default value when running ".get('results')" which it's None.

This change sets the default value to an empty dictionary to allow the following ".get('Stdout', '')" succeed.

### DIFF
--- a/zaza/openstack/utilities/openstack.py
+++ b/zaza/openstack/utilities/openstack.py
@@ -238,7 +238,7 @@ async def async_block_until_ca_exists(application_name, ca_cert,
             for unit in units:
                 try:
                     output = await unit.run('cat {}'.format(ca_file))
-                    contents = output.data.get('results').get('Stdout', '')
+                    contents = output.data.get('results', {}).get('Stdout', '')
                     if ca_cert not in contents:
                         break
                 # libjuju throws a generic error for connection failure. So we


freyes:
This PR was exercised successfully by this gerrit change - https://review.opendev.org/c/openstack/charm-mysql-innodb-cluster/+/912487

sabaini (approved):
+1 as a tactical fix

The results attribute for `unit.run()` missing sometime smells a bit like a bug in libjuju though?

cheers!

zaza/openstack/utilities/openstack.py
```diff
@@ -238,7 +238,7 @@ async def _check_ca_present(model, ca_files):
             for unit in units:
                 try:
                     output = await unit.run('cat {}'.format(ca_file))
-                    contents = output.data.get('results').get('Stdout', '')
+                    contents = output.data.get('results', {}).get('Stdout', '')
```
wolsen, line 241: Hmm, I think the problem here is that this is invoking direct juju invocations rather than the zaza model. The zaza model has logic to deal with the async result differences between libjuju 2.x and 3.x, but this code doesn't have that.

wolsen (changes requested):

wolsen:
I'm closing this pull request as we used PR #1192 instead, which uses the underlying zaza.model changes to not use the unit.run directly.

PR #1192, opened 2024-03-18
Use zaza.model.run_on_unit for ca checks

wolsen:
Existing code uses the python libjuju unit.run in order to execute a wait check for ca readiness across the units. The behavior of libjuju changed between 2.x and 3.x and causes this functionality to break. This is abstracted and handled in the zaza library, so use that code instead as it properly handles the differences.


wolsen:
Changes are being exercised in https://review.opendev.org/c/openstack/charm-mysql-innodb-cluster/+/912487

ajkavanagh (approved):

PR #1195, opened 2024-04-05
[WIP] Implement tests for ceph-mon's rotate-key action

lmlg:
This PR implements tests for the ceph-mon charm's action 'rotate-key'. For now, the charm only supports key rotation for managers, but more entities will be added. This PR will be used to test those new entities as well.

### DIFF
--- a/zaza/openstack/charm_tests/ceph/tests.py
+++ b/zaza/openstack/charm_tests/ceph/tests.py
@@ -1748,3 +1748,39 @@ class CephMonJujuPersistent(test_utils.BaseCharmTest):
             )
             data = json.loads(result['Stdout'])
             assert data['loglevel'] == 2
+
+
+class CephMonKeyRotationTests(test_utils.BaseCharmTest):
+    """Tests for the rotate-key action."""
+
+    def _get_key_for_mgr(self, unit, mgr_id):
+        cmd = 'sudo ceph auth ls'
+        result = zaza_model.run_on_unit(unit, cmd)
+        # Don't use json formatting, as it's buggy upstream.
+        data = result['Stdout'].split()
+
+        for ix, line in enumerate(data):
+            # Structure:
+            # mgr.XXXX
+            # key:
+            # key contents
+            # That's why we need to move 2 positions ahead.
+            if line.startswith('mgr.') and line.endswith(mgr_id):
+                return data[ix + 2]
+
+    def test_mgr_key_rotate(self):
+        """Test that rotating the manager key actually changes it."""
+        unit = 'ceph-mon/0'
+        mgr_id = unit[-1]
+        old_key = self._get_key_for_mgr(unit, mgr_id)
+        self.assertIsNotNone(old_key)
+
+        action_obj = zaza_model.run_action(
+            unit_name=unit,
+            action_name='rotate-key',
+            action_params={'entity': 'mgr'}
+        )
+        zaza_utils.assertActionRanOk(action_obj)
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -1748,3 +1748,38 @@ def test_persistent_config(self):
...
+        old_keys = self._get_mgrs_keys(unit)
+
+        action_obj = zaza_model.run_action(
+            unit_name=unit,
+            action_name='rotate-key',
+            action_params={'entity': 'mgr'}
+        )
+        zaza_utils.assertActionRanOK(action_obj)
+        new_keys = self._get_mgrs_keys(unit)
+        self.assertNotEqual(old_keys, new_keys)
```
sabaini, line 1785: Believe this would succeed if just one of the keys were changed -- suggest to check that actually each key is changed

lmlg: This only rotates the key for the manager on the ceph-mon unit specified. Getting the manager for a mon is rather complicated, since managers don't have stable names (they change on upgrade, for instance).
I can harden it a bit by asserting that only one key changed, though.

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -1748,3 +1748,72 @@ def test_persistent_config(self):
...
+        first = next(iter(diff))
+        # Check that the entity matches. The 'entity_filter'
+        # callable will return a true-like value if it
+        # matches the type of entity we're after (i.e: 'mgr')
+        self.assertTrue(entity_filter(first[0]))
+
+    def _get_rgw_client(self, unit):
+        cmd = 'sudo ceph auth ls | grep client.rgw'
+        result = zaza_model.run_on_unit(unit, cmd)
+        return result['Stdout'].strip()
```
sabaini, line 1798: Or could also run the _get_all_keys() helper from above I think?

zaza/openstack/charm_tests/ceph/tests.py
```diff
@@ -1748,3 +1748,72 @@ def test_persistent_config(self):
...
+
+        try:
+            zaza_model.get_application('ceph-radosgw')
+            rgw_client = self._get_rgw_client('ceph-radosgw/0')
+            if rgw_client:
+                self._check_key_rotation(rgw_client, unit)
+            else:
+                logging.info('ceph-radosgw units present, but no RGW service')
+        except KeyError:
+            pass
```
sabaini, line 1813: Hm, what would cause a KeyError here? Should we add some logging around this?
Similar in the below try/except

sabaini (commented):
Quick notes the checking

zaza/openstack/charm_tests/ceph/tests.py
lmlg: `KeyError` is meant to catch the case when an application (`ceph-radosgw` in this case) isn't present. It's a common pattern throughout z-o-t

lmlg:
The PR is ready for merge now. The mgr and RGW pieces have been run with this PR already.

zaza/openstack/charm_tests/ceph/tests.py
lmlg: Done.

lmlg:
This PR has been successfully tested on these MR's:
https://review.opendev.org/c/openstack/charm-ceph-mon/+/915087
https://review.opendev.org/c/openstack/charm-ceph-radosgw/+/915339

PR #1197, opened 2024-04-10
Fix race in setting up certs

sabaini:
Closes bug #2060785


sabaini:
Test is being exercised in https://review.opendev.org/c/openstack/charm-ceph-radosgw/+/915433

lmlg (approved):
test went well, lgtm!

PR #1199, opened 2024-04-17
Fix: ceph-radosgw test should only check radosgw app

sabaini:
Don't block on global app state waiting but only check the SUT app state.


sabaini:
Exercised in https://review.opendev.org/c/openstack/charm-ceph-radosgw/+/916118

lmlg (approved):
lgtm

PR #1201, opened 2024-04-22
Implement key rotation tests for OSD units

lmlg:
This PR adds OSD tests to the key rotation class for the ceph-mon charm. It will be tested in the following merge requests:
https://review.opendev.org/c/openstack/charm-ceph-mon/+/915925
https://review.opendev.org/c/openstack/charm-ceph-osd/+/915926


sabaini (approved):
lgtm, cheers!

PR #1202, opened 2024-04-24
Implement ceph-fs key rotation tests

lmlg:
This PR implements tests for key rotation in ceph-fs units. It works similarly to how RGW tests do.


sabaini (approved):
lgtm, cheers

PR #1208, opened 2024-05-13
Fix test for ceph-fs key rotation

lmlg:
The ceph-fs charm sometimes creates an entity called 'mds.None'. This is probably a bug, but in the meantime, we filter it out in the key rotation test class.


lmlg:
LP Bug: https://bugs.launchpad.net/charm-ceph-fs/+bug/2065900

sabaini:
lgtm as a tactical fix

PR #1212, opened 2024-05-24
Add class for keystone audit middleware testing

MylesJP:
Added general class for testing keystone audit middleware functionality in charms. Tests correct rendering of api-paste.ini file.

Example usage with charm Heat:

tests/tests.yaml
```
tests:
  - zaza.openstack.charm_tests.audit.tests.KeystoneAuditMiddlewareTest

tests_options:
  audit-middleware:
    service: heat
```

### DIFF
--- a/zaza/openstack/charm_tests/audit/tests.py
+++ b/zaza/openstack/charm_tests/audit/tests.py
@@ -17,65 +17,77 @@
 """
 Keystone audit middleware API logging testing.
 
-These methods test the
-rendering of the charm's api-paste.ini file to ensure the appropriate sections
-are rendered or not rendered depending on the state of the audit-middleware
-configuration option.
+These methods test the rendering of the charm api-paste.ini file to
+ensure the appropriate sections are rendered or not rendered depending
+on the state of the audit-middleware configuration option.
 """
 
+import textwrap
 import logging
 import zaza.model
 import zaza.openstack.charm_tests.test_utils as test_utils
 
 
-class KeystoneAuditMiddlewareBaseTest(test_utils.OpenStackBaseTest):
-    """Base class for Keystone audit middleware tests."""
+class KeystoneAuditMiddlewareTest(test_utils.OpenStackBaseTest):
+    """Keystone audit middleware test class."""
 
     @classmethod
     def setUpClass(cls):
         """Run class setup for Keystone audit middleware tests."""
-        super(KeystoneAuditMiddlewareBaseTest, cls).setUpClass()
+        super(KeystoneAuditMiddlewareTest, cls).setUpClass()
         test_config = cls.test_config['tests_options']['audit-middleware']
         cls.service_name = test_config['service']
 
-        # For cases when service name and application name differ
-        if not test_config["application"]:
-            cls.application_name = cls.service_name
-        else:
+        if test_config['application']:
             cls.application_name = test_config['application']
-        print(test_config)
-        print(cls.model_name)
-        cls.lead_unit = zaza.model.get_lead_unit_name(
+        else:
+            cls.application_name = cls.service_name
+            logging.info('Application name not set, using service name:\
+                         %s', cls.service_name)
+
+        cls.initial_audit_middleware = zaza.model.get_application_config(
+            cls.application_name)['audit-middleware']['value']
+
+    @classmethod
+    def tearDownClass(cls):
+        """Restore the audit-middleware configuration to its original state."""
+        super(KeystoneAuditMiddlewareTest, cls).tearDownClass()
+        logging.info("Running teardown on %s" % cls.application_name)
+        zaza.model.set_application_config(
             cls.application_name,
+            {'audit-middleware': str(cls.initial_audit_middleware)},
             model_name=cls.model_name
         )
-        print(cls.lead_unit)
-        logging.info('Leader unit is %s', cls.lead_unit)
-        logging.info('Service name is %s', cls.service_name)
+        zaza.model.wait_for_application_states(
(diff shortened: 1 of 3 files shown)


MylesJP:
Added a fix for the case where the service name and the application name differ. In most cases they are the same but in nova-cloud-controller, service name is 'nova' in file path while application_name is nova-cloud-controller.

zaza/openstack/charm_tests/audit/tests.py
```diff
@@ -0,0 +1,98 @@
...
+        test_config = cls.test_config['tests_options']['audit-middleware']
+        cls.service_name = test_config['service']
+
+        # For cases when service name and application name differ
+        if not test_config["application"]:
+            cls.application_name = cls.service_name
+        else:
+            cls.application_name = test_config['application']
+        print(test_config)
+        print(cls.model_name)
```
freyes, line 47: drop this, if you need this information, replace the print() with logging.debug()

zaza/openstack/charm_tests/audit/tests.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        """Run class setup for Keystone audit middleware tests."""
+        super(KeystoneAuditMiddlewareBaseTest, cls).setUpClass()
+        test_config = cls.test_config['tests_options']['audit-middleware']
+        cls.service_name = test_config['service']
+
+        # For cases when service name and application name differ
+        if not test_config["application"]:
```
freyes, line 42: please, swap the if-else statment, it's easier to reason `if test_config["application"]` than the `if not...`

zaza/openstack/charm_tests/audit/tests.py
```diff
@@ -0,0 +1,98 @@
...
+            cls.application_name = cls.service_name
```
freyes, line 43: add a logging.info() statement to inform you are re-using the application name as service name, so this way you leave a breadcrumb for users just reading at the CI logs, something like `logging.info('Application name not set, re-using service name %s', cls.service_name)`

zaza/openstack/charm_tests/audit/tests.py
```diff
@@ -0,0 +1,98 @@
...
+        else:
+            cls.application_name = test_config['application']
+        print(test_config)
+        print(cls.model_name)
+        cls.lead_unit = zaza.model.get_lead_unit_name(
```
freyes, line 48: this could change over time, storing the leader unit during the `setUpClass` may lead to inaccurate results, it's better you get the lead unit closer to where it will be used (e.g. the test method).

zaza/openstack/charm_tests/audit/tests.py
```diff
@@ -0,0 +1,98 @@
...
+        try:
+            return zaza.model.file_contents(
+                self.lead_unit,
+                api_paste_ini_path,
+            )
+        except zaza.model.CommandRunFailed as e:
+            self.fail("Error fetching api-paste.ini: {}".format(str(e)))
+
+
+class KeystoneAuditMiddlewareTest(KeystoneAuditMiddlewareBaseTest):
```
freyes, line 68: I believe there is no need to have 2 separate classes for this, you can merge  `KeystoneAuditMiddlewareBaseTest` into this class, unless you are anticipating a future need to have this separation, although if that's the case I will argue that it's better we split the class when the need comes in a future PR.

zaza/openstack/charm_tests/audit/tests.py
```diff
@@ -0,0 +1,98 @@
...
+
+        set_default = {'audit-middleware': False}
+        set_alternate = {'audit-middleware': True}
+
+        with self.config_change(default_config=set_default,
+                                alternate_config=set_alternate,
+                                application_name=self.application_name):
+            api_paste_content = self.fetch_api_paste_content()
+            for line in expected_content:
+                self.assertIn(line, api_paste_content)
```
freyes, line 88: the function `file_contents()` returns a multiline string, it's better you do something like this:

```python
expected_content = f"""[filter:audit]
paste.filter_factory = keystonemiddleware.audit:filter_factory
audit_map_file = /etc/{self.service_name}/api_audit_map.conf
service_name = {self.service_name}
""""
self.assertIn(expected_content, api_paste_content)
```

this way you don't have to iterate the string.

zaza/openstack/charm_tests/audit/tests.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def test_102_apipaste_excludes_audit_section(self):
+        """Test api_paste.ini does not render audit section when disabled."""
+        section_heading = '[filter:audit]'
+
+        if not self.config_current(self.application_name)['audit-middleware']:
+            api_paste_content = self.fetch_api_paste_content()
+            self.assertNotIn(section_heading, api_paste_content)
+        else:
+            self.fail("Config option audit-middleware incorrectly set to true")
```
freyes, line 98: if it's set to true, why don't you set to false using the `with self.config_change...`?, this way you would be able to test the content of the api-paste.ini file no matter how the testing bundle looks like, you make your testing less dependent on the setup of the environment, it pays off over time , because it drops the requirement of keeping two pieces in sync (the testing bundle and the expectations of this test).

freyes (changes requested):
Hi Myles, thanks for proposing this change, I left some inline comments that need to be addressed.

freyes:
This PR is being exercised by this gerrit change - https://review.opendev.org/c/openstack/charm-nova-cloud-controller/+/918017

PR #1213, opened 2024-05-30
Remove designate-agent for caracal+ testing

ajkavanagh:
The designate-agent service was removed at caracal. This patch removes
the designate-agent from the list of services that designate supports so
that the pause/resume testing works appropriately.


wolsen (approved):
lgtm

PR #1214, opened 2024-05-30
Remove designate-agent for caracal+ testing

wolsen:
Remove designate-agent for caracal+ testing
    
The designate-agent service was removed at caracal. This patch removes
the designate-agent from the list of services that designate supports so
that the pause/resume testing works appropriately.

(cherry picked from commit d83e01f86270dd7005c0fab26a4e387977726701)


freyes (approved):
LGTM.

PR #1223, opened 2024-06-17
Add audit middleware test class for ironic-api

MylesJP:
New test class for extension of Keystone Audit Middleware tests for charm Ironic-API. This extension is required because Ironic-API does not use an api-paste.ini file and the existing tests are not relevant.

### DIFF
--- a/zaza/openstack/charm_tests/audit/__init__.py
+++ b/zaza/openstack/charm_tests/audit/__init__.py
@@ -1,0 +1,20 @@
+# Copyright 2024 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


freyes (changes requested):
What's the associate review.opendev.org patch that's exercising this PR?, the one that has the "Func-Test-Pr: https://..." header.

freyes (changes requested):
Since you will need to backport this PR in the future, it's better if you squash these two commits.

Ref: https://gist.github.com/lpranam/4ae996b0a4bc37448dc80356efbca7fa

MylesJP:
> What's the associate review.opendev.org patch that's exercising this PR?, the one that has the "Func-Test-Pr: https://..." header.

https://review.opendev.org/c/openstack/charm-ironic-api/+/922144

The CI is currently being hung up on the jammy-caracal build with ironic-api but passes the jammy-bobcat build.

PR #1227, opened 2024-06-20
Add a constraints file specifically for noble

wolsen:
Add a constraints file to pin to a version of juju which supports noble. Adding a new file so as not to impact current master branch targets for all charms until they are ready.


hemanthnakkina (approved):
LGTM 
I am guessing the file name is constraints-noble instead of constraints-<openstack release> as this is specifically used for s390x

PR #1235, opened 2024-06-25
Added required config option idp_id to Keystone OpenIDC charm

jadonn:
This change makes z-o-t pass in the value of idp_id when configuring the Keystone OpenIDC charm during testing.

The idp_id cannot be set to a default value. It has to match the name of the identity provider you set up in Keystone for federation. If you do not do this, the URLs in the template do not render properly.


jadonn:
This PR is related to this change https://review.opendev.org/c/openstack/charm-keystone-openidc/+/922049.

PR #1244, opened 2024-07-03
[caracal] Enable SSL support on manila client

ajkavanagh:
The python-manilaclient requires passing the cacert file for TLS
endpoints. This commit enables that on the `get_manila_session_client`
and also, for the manila tests, wraps the client in a retrier that
retries on connection failures (i.e. if the server is not running).


ajkavanagh:
Exercised by: https://review.opendev.org/c/openstack/charm-manila-ganesha/+/923383

PR #1257, opened 2024-08-22
Unpin cinderclient

gboutry:
cinderclient was pinned to train release, this version prevented zaza to be run with python 3.12

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,7 @@ install_require = [
     'python-neutronclient<7.0.0',
     'python-octaviaclient<1.11.0',
     'python-ceilometerclient',
-    'python-cinderclient<6.0.0',
+    'python-cinderclient',
     'python-swiftclient<3.9.0',
     'python-watcherclient',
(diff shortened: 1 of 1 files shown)


hemanthnakkina (commented):
Shall we unpin the rest of clients as well or use versions from https://opendev.org/openstack/requirements/src/branch/master/upper-constraints.txt?

gboutry:
I think, except from an incompatibility in a specific client, they should be unpinned, and choice of versions given to users of the lib, where they can provide their own constraints.

setup.py
```diff
@@ -63,7 +63,7 @@
     'python-neutronclient<7.0.0',
     'python-octaviaclient<1.11.0',
     'python-ceilometerclient',
-    'python-cinderclient<6.0.0',
+    'python-cinderclient',
```
freyes, line 66: LGTM. We were just talking about this an hour ago, thanks for the fix!

freyes (approved):

PR #1261, opened 2024-09-02
COS integration: make prom-api configurable

sabaini:
Refactor integration tests and make retrieving prom-api configurable

### DIFF
--- a/zaza/openstack/charm_tests/ceph/mon/integration.py
+++ b/zaza/openstack/charm_tests/ceph/mon/integration.py
@@ -13,19 +13,13 @@
 # limitations under the License.
 
 """Integration tests for ceph-mon."""
-
+import requests
+import tenacity
+import yaml
 import zaza.model
 from juju import juju
 from zaza import sync_wrapper
 from zaza.openstack.charm_tests import test_utils as test_utils
-
-from zaza.openstack.charm_tests.ceph.mon.tests import (
-    get_prom_api_url,
-    get_up_osd_count,
-    extract_pool_names,
-    get_alert_rules,
-    get_dashboards,
-)
 
 
 async def async_find_cos_model():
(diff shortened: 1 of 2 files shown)


zaza/openstack/charm_tests/ceph/mon/integration.py
```diff
@@ -58,6 +52,90 @@ async def async_find_cos_model():
...
+
+
+def get_alert_rules(prometheus_url):
+    """Get the alert rules from prometheus."""
+    response = requests.get(f'{prometheus_url}/rules')
+    data = response.json()
+    if data['status'] != 'success':
+        raise Exception(f"Query failed: {data.get('error', 'Unknown error')}")
+
+    alert_names = []
```
lmlg, line 103: why not just use a `set` and then call `add` instead of accumulating into a list and then converting?

zaza/openstack/charm_tests/ceph/mon/integration.py
```diff
@@ -58,6 +52,90 @@ async def async_find_cos_model():
...
+@tenacity.retry(wait=tenacity.wait_fixed(5),
+                stop=tenacity.stop_after_delay(180))
+def get_prom_api_url(grafana_agent):
+    """Get the prometheus API URL from the grafana-agent config."""
+    ga_yaml = zaza.model.file_contents(
+        f"{grafana_agent}/leader", "/etc/grafana-agent.yaml"
+    )
+    ga = yaml.safe_load(ga_yaml)
+    url = ga['integrations']['prometheus_remote_write'][0]['url']
+    return url[:-6]  # lob off the /write
```
lmlg, line 122: I recommend calling `url.endswith('/write')` before chopping off the string so that future versions that change the result don't end up biting us.

lmlg (approved):
lgtm, just 2 nits

PR #1266, opened 2024-09-27
Skip noble image sync for octavia tests

dosaboy:
Noble was recently added but is not needed and is tipping glance/ceph storage usage over the limit in release tests.


freyes (approved):
LGTM

PR #1267, opened 2024-09-30
Try cached tempest octavia test_server.bin

dosaboy:


dosaboy:
I have tested this with https://github.com/openstack-charmers/charmed-openstack-tester and it works well. It is needed because the environment I am testing it does not not have these resources in swift.

PR #1274, opened 2024-10-07
Skip unsupported vhost config on older releases

sabaini:
Quincy and older do not support virtual hosted bucket


sabaini:
functests have been run locally

PR #1282, opened 2024-10-18
Fallback to 0 minimum compute when no nova-compute units

gboutry:
Tempests can be used on deployments without nova-compute, testing only control plane APIs. Fallback to a value when no nova-compute units is found.


fnordahl (approved):
Apparently, I introduced this bug, and thank you for fixing it. LGTM!

gboutry:
Thank you!

PR #1309, opened 2025-04-16
ovn: Add test for prometheus exporter.

mkalcok:
This ensures that prometheus exporter on ovn-chassis and ovn-central units is reachable and responds to requests after the unit is deployed.

I will trigger CI for both chassis and central charms to test this change and I'll report back.

### DIFF
--- a/zaza/openstack/charm_tests/ovn/tests.py
+++ b/zaza/openstack/charm_tests/ovn/tests.py
@@ -18,6 +18,7 @@ import logging
 
 import juju
 
+import requests
 import tenacity
 import yaml
 import zaza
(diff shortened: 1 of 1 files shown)


mkalcok:
`charm-ovn-chassis` [passed CI with this change](https://github.com/canonical/charm-ovn-chassis/pull/6), however `charm-ovn-central` [didn't get picked up by ZOSCI](https://review.opendev.org/c/x/charm-ovn-central/+/947515). I did try to run locally and it passed, but I think it'd be better to have official CI run before we merge this test in.

fnordahl (approved):

PR #1310, opened 2025-04-29
Add tempest target with new RBAC

gboutry:
With Epoxy, new rbacs becomes the default under oslo.policy. Add a new test target TempestTestWithKeystoneMinimalNewRBAC to ensure proper configuration for tempest under new RBAC environments.

### DIFF
--- a/zaza/openstack/charm_tests/tempest/templates/tempest_v3.j2
+++ b/zaza/openstack/charm_tests/tempest/templates/tempest_v3.j2
@@ -47,7 +47,11 @@ auth_version = v3
 admin_role = Admin
 region = RegionOne
 default_domain_id = {{ default_domain_id }}
+{% if admin_domain_scope is defined -%}
+admin_domain_scope = {{ admin_domain_scope }}
+{% else -%}
 admin_domain_scope = true
+{% endif -%}
 disable_ssl_certificate_validation = true
 
 [identity-feature-enabled]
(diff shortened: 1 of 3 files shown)


zaza/openstack/charm_tests/tempest/templates/tempest_v3.j2
```diff
@@ -164,3 +168,34 @@ nameservers = {{ test_name_server }}
 [service-clients]
 # Default is 60s which is not enough for resource constrained environments.
 http_timeout = 120
+
+{% if enforce_scope is defined and enforce_scope -%}
+[enforce_scope]
+{% if "keystone" in enabled_services -%}
```
freyes, line 174: This would be more compact if we use the same approach used by the `[service_available]` section

```
{% for svc in enabled_services -%}
{{ svc }} = true
{% endfor -%}
{% for svc in disabled_services -%}
{{ svc }} = false
{% endfor %}
```

gboutry: Indeed but heat tempest plugin, for example, does support this. 

Should I create a new context setting with all the compatible plugins and service enabled?

freyes: yeah, I think that would be better, that way is easier to add others as they grow support

hemanthnakkina (approved):

PR #1313, opened 2025-05-05
Pin osc-lib<4.0.0

freyes:
osc-lib-4.0.0 introduced typing hints making it incompatible with py38.

Ref: https://github.com/openstack/osc-lib/commit/3d221e5992a74d6bb7d1b725e43015b8c87b9718


freyes:
We need this pinning on master, because the CI runs on Focal (py38).

MylesJP:
LGTM

PR #1332, opened 2025-12-15
Add zaza tests for cert renewals for ovn-central

arif-ali:
Tests for the charm changes as suggested from https://github.com/canonical/charm-ovn-central/pull/11

### DIFF
--- a/zaza/openstack/charm_tests/ovn/tests.py
+++ b/zaza/openstack/charm_tests/ovn/tests.py
@@ -15,6 +15,7 @@
 """Encapsulate OVN testing."""
 
 import logging
+import unittest
 
 import juju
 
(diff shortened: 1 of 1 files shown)


mkalcok (commented):
Thank you for the improvement @arif-ali. LGTM.
OVN central tests completed successfully with this: https://github.com/canonical/charm-ovn-central/pull/11

One question, is this test ovn-central specific? Or could it be called just `OVNSSLExpireTests`. And more importantly, does it make sense to introduce it to OVN chassis and dedicated chassis as well?

arif-ali:
This is `ovn-central` specific. The issue is not apparent on the ovn-chassis charms, at least we didn't face this as an issue. 

The certs for ovn-chassis get updated without any problems. In-fact I don't think the `ovn-chassis` charm checks for the ovn-chassis service liveness in the same way that ovn-central does.

mkalcok:
Ack. Thanks. I wasn't sure of the top off my head about how client-side certs are handled in chassis charms.

PR #1335, opened 2026-01-14
undercloud_and_charm_setup: Add support for LXD.

mkalcok:
This adds basic LXD support that allows adding second NIC to instances hosting networking charms as well as populating said charms configuration with the second NIC MAC address.

(cherry picked from commit a5ad00acf1b54d0461f6a288484e69b4d890a91c)


mkalcok:
We should probably wait for actual OVN functional tests passing before merging. But given that this is a smooth backport (already succesfully applied to stable/caracal), I'm marking it as "ready for review"

PR #1336, opened 2026-01-22
ovn: Make "Prometheus" and "HW Offload" tests skippable.

mkalcok:
In order to use GH runners for functional tests of OVN chassis charms, we need the support for LXD VMs in Juju. This feature is supported since 3.1, but unfortunately, `zaza` and `zaza-openstack-tests` in `stable/yoga` branch don't support Juju3.
A solution for `22.03` releases of `ovn-chassis` and `ovn-dedicated-chassis`, which until now used zaza `stable/yoga`, is to upgrade zaza to `stable/caracal`. This works fine in general and brings us a lot of fixes that were already backported to this branch, but we need to make couple tests optional, because the functionality is not present in `22.03` version of charms. Optional tests are controlled via environment variables.

* Exporting `TEST_SKIP_PROMETHEUS` skips test for reachability of prometheus exporter
* Exporting `TEST_SKIP_HW_OFFLOAD` skips test for "Hardware offload"

### DIFF
--- a/zaza/openstack/charm_tests/ovn/tests.py
+++ b/zaza/openstack/charm_tests/ovn/tests.py
@@ -15,6 +15,7 @@
 """Encapsulate OVN testing."""
 
 import logging
+import os
 import unittest
 
 import juju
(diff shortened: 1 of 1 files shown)


mkalcok:
The change was tested and passed in this PR: https://github.com/canonical/charm-ovn-chassis/pull/18

fnordahl (approved):
Seems like a reasonable compromise, LGTM.

PR #1338, opened 2026-02-10
Add zaza tests for cert renewals for ovn-central

arif-ali:
This also adds some dependent stuff that came through the cherry-pick


crypticC0der (approved):
Thanks for your work on this, the added tests tie nicely into https://github.com/canonical/charm-ovn-central/pull/16

Issue #1339, opened 2026-02-11
zaza-openstack-tests fails when setuptool 82.0.0+ is installed


gboutry:
https://github.com/openstack-charmers/zaza-openstack-tests/blob/42f49f6f05aa8f6053afa02520f59d04ae96ce1c/setup.py#L60

Keystoneclient 4.5.0 drop usage of `package_resources`, unfortunately, this library pins it earlier.

```
2026-02-11 10:31:46.014520 | charms-ubuntu-noble-32GB |   File "/home/zuul/src/opendev.org/openstack/sunbeam-charms/.tox/func/bin/functest-run-suite", line 7, in <module>
2026-02-11 10:31:46.014897 | charms-ubuntu-noble-32GB |     sys.exit(main())
2026-02-11 10:31:46.014910 | charms-ubuntu-noble-32GB |              ^^^^^^
2026-02-11 10:31:46.014916 | charms-ubuntu-noble-32GB |   File "/home/zuul/src/opendev.org/openstack/sunbeam-charms/.tox/func/lib/python3.12/site-packages/zaza/charm_lifecycle/func_test_runner.py", line 405, in main
2026-02-11 10:31:46.017393 | charms-ubuntu-noble-32GB |     func_test_runner(
2026-02-11 10:31:46.017514 | charms-ubuntu-noble-32GB |   File "/home/zuul/src/opendev.org/openstack/sunbeam-charms/.tox/func/lib/python3.12/site-packages/zaza/charm_lifecycle/func_test_runner.py", line 313, in func_test_runner
2026-02-11 10:31:46.017524 | charms-ubuntu-noble-32GB |     run_env_deployment(env_deployment, keep_model=preserve_model,
2026-02-11 10:31:46.017671 | charms-ubuntu-noble-32GB |   File "/home/zuul/src/opendev.org/openstack/sunbeam-charms/.tox/func/lib/python3.12/site-packages/zaza/charm_lifecycle/func_test_runner.py", line 174, in run_env_deployment
2026-02-11 10:31:46.017679 | charms-ubuntu-noble-32GB |     configure.configure(
2026-02-11 10:31:46.017684 | charms-ubuntu-noble-32GB |   File "/home/zuul/src/opendev.org/openstack/sunbeam-charms/.tox/func/lib/python3.12/site-packages/zaza/charm_lifecycle/configure.py", line 58, in configure
[...]

PR #1340, opened 2026-03-04
Pin setuptools<82 in constraints and requirements

xtrusia:
setuptools>=82 removes the deprecated pkg_resources.declare_namespace() which breaks several OpenStack client libraries that still rely on it.

Refs https://github.com/openstack-charmers/zaza-openstack-tests/issues/1339


freyes (commented):
should we add the pinning to the setup.py install requires too? - https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/setup.py#L27

freyes (changes requested):
can you add a "Refs #1339" header too?.

freyes (approved):

PR #1341, opened 2026-03-16
constraints: Constrain setuptools.

fnordahl:
There are multiple direct dependencies pulled in from `setup.py` that do not work with a setuptools that does not offer `pkg_resources`.


freyes:
Already fixed by https://github.com/openstack-charmers/zaza-openstack-tests/pull/1340

PR #1342, opened 2026-04-20
dpdk-bond-mapping tests for DPDKTest bundle

PanosKostopoulos:
DPDK tests for dpdk-bond-mappings ovn-chassis config option. This first iteration includes only PCI addresses, but should include also MAC addresses (the difference is there should be an extra lookup to get the PCI addresses from the mac).

The tests have extended the test_enable_dpdk function, in this first iteration, to avoid a second setup for dpdk. This could be moved (and maybe should) in a seperate test function.

### DIFF
--- a/zaza/openstack/charm_tests/ovn/tests.py
+++ b/zaza/openstack/charm_tests/ovn/tests.py
@@ -16,6 +16,7 @@
 
 import logging
 import unittest
+import json
 
 import juju
 
@@ -26,8 +27,10 @@ import zaza
 
 import zaza.model
 import zaza.openstack.charm_tests.ceph.mon.integration as cos_integration
+import zaza.openstack.charm_tests.neutron.setup as neutron_setup
 import zaza.openstack.charm_tests.test_utils as test_utils
 import zaza.openstack.utilities.generic as generic_utils
+import zaza.openstack.utilities.openstack as openstack_utils
 import zaza.utilities.juju
 
 from zaza.openstack.charm_tests.cos.setup import GRAFANA_OFFER_ALIAS
@@ -355,6 +358,12 @@ class ChassisCharmOperationTest(BaseCharmOperationTest):
 class DPDKTest(test_utils.BaseCharmTest):
     """DPDK-related tests."""
 
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        obj = cls()
+        obj.enable_hugepages_vfio_on_hvs_in_vms(4)
+
     def _openvswitch_switch_dpdk_installed(self):
         """Assert that the openvswitch-switch-dpdk package is installed.
 
@@ -364,6 +373,20 @@ class DPDKTest(test_utils.BaseCharmTest):
         for unit in zaza.model.get_units(self.application_name):
             zaza.utilities.juju.remote_run(
                 unit.name, cmd, model_name=self.model_name, fatal=True)
+
+    def _get_first_macs(self):
+        """Retrieve macs related to interface eth1 in networking charm units.
+
+        :returns: list of mac addresses(represented as strings).
+        """
+
+        networking_data = openstack_utils.get_charm_networking_data()
+        ifname="eth1"
+        macs=[]
+        for instance in networking_data.unit_machine_ids:
+            macs.append(str(openstack_utils.lxd_get_nic_hwaddr(instance, ifname)))
+
+        return macs
 
     def _ovs_dpdk_init_configured(self):
         """Assert that DPDK is configured.
@@ -452,13 +475,28 @@ class DPDKTest(test_utils.BaseCharmTest):
                 msg='OVS unexpectedly has DPDK initialized'):
             self._ovs_dpdk_initialized()
 
+    def _get_br_ex_dpdk_pci_address(self, unit):
+        """Get the PCI addresses of the DPDK interfaces bound to br-ex.
+
+        :param unit: Juju unit to inspect.
+        :type unit: juju.unit.Unit
+        :returns: list of PCI address string (e.g. ``'0000:04:00.0'``).
+        :rtype: list
+        :raises: zaza.model.CommandRunFailed
+        """
+        cmd = "lspci | awk '/Ethernet controller/ {print $1}' | tail -n 2 | tr '\n' ' '"
+        
+        result = zaza.utilities.juju.remote_run(
+            unit.name, cmd, model_name=self.model_name, fatal=True)
+        
+        return result.strip().split(' ')
+    
     def test_enable_dpdk(self):
         """Confirm that transitioning to/from DPDK works."""
         logging.info('Pre-flight check')
         self._dpdk_pre_post_flight_check()
         self._ovs_br_ex_port_is_system_interface()
 
-        self.enable_hugepages_vfio_on_hvs_in_vms(4)
         with self.config_change(
                 {
                     'enable-dpdk': False,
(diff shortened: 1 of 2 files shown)


zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -492,6 +542,49 @@ def test_enable_dpdk(self):
...
+            'Setting dpdk-bond-mappings to: {}'.format(bond_config))
+
+        with self.config_change(
+                {'dpdk-bond-mappings': ''},
+                {'dpdk-bond-mappings': bond_config},
+                application_name='ovn-chassis'):
+            logging.info(
+                'Verifying bond PCI addresses are whitelisted in dpdk-extra')
+            self._assert_dpdk_bond_pci_in_eal_args(bond_pci_addresses)
+
```
fnordahl: The PCI address is put in EAL args regardless of the use of `dpdk-bond-mappings` configuration option.

How is this a valid assertion?

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -492,6 +542,49 @@ def test_enable_dpdk(self):
         # self.disable_hugepages_vfio_on_hvs_in_vms()
         # self._ovs_br_ex_port_is_system_interface()
 
+    def _test_dpdk_bond_mappings_pci_format(self):
```
fnordahl: Make this a separate `def test_XXX` method instead of this covert piggy-backing on existing tests.

The charm code will only actually populate the OVS database with bond configuration when there are more than one interface [0], so you likely need to enable the test harness to add multiple ports to the test VM as well as interrogate the environment and ensure there are enough ports available to run the test.

0: https://github.com/openstack-charmers/charm-layer-ovn/blob/b0cee1d3b1bf7e7fc4236d9b2057419d3411771b/lib/charms/ovn_charm.py#L1324

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -492,6 +542,49 @@ def test_enable_dpdk(self):
...
+        (``other_config:dpdk-extra`` in OVS).
+
+        Must be called while DPDK is already enabled (i.e. inside a
+        ``config_change`` context that has ``enable-dpdk: true``).
+
+        :raises: AssertionError, zaza.model.CommandRunFailed
+        """
+        bond_pci_addresses = []
+        for unit in zaza.model.get_units(self.application_name):
+            pci = self._get_br_ex_dpdk_pci_address(unit)
```
fnordahl: Basing input for your test assertion on data provided by the subject under test is not the safe thing to do, you should have the test code independently discover PCI addresses and/or MACs to use as input for the test so that you can accurately measure the output of the program.

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -492,6 +542,49 @@ def test_enable_dpdk(self):
...
+                bond_pci_addresses.append(pci)
+                logging.info(
+                    'Collected br-ex PCI address {} from {}'.format(
+                        pci, unit.name))
+
+        if not bond_pci_addresses:
+            logging.warning(
+                'Could not retrieve any PCI addresses for bond mapping test, '
+                'skipping')
+            return
```
fnordahl: This does not look safe, we'd likely want to know if this failed by having the test fail.

fnordahl (changes requested):

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -469,28 +597,57 @@ def test_enable_dpdk(self):
...
-            logging.info('Checking DPDK is configured in OVS')
-            self._ovs_dpdk_init_configured()
-            logging.info('Checking DPDK is successfully initialized in OVS')
-            self._ovs_dpdk_initialized()
-            logging.info('Checking that br-ex configed with DPDK interface...')
-            self._ovs_br_ex_port_is_dpdk_interface()
-            logging.info('and is not in error.')
-            self._ovs_br_ex_interface_not_in_error()
+            instance_mac_map = self._add_second_interface()
+            self.bind_second_interface_to_dpdk(instance_mac_map)
```
fnordahl: Ensuring devices mentioned in the charm configuration are bound to a driver appropriate for use by DPDK is the responsibility of the charm.

The charm does this by updating /etc/dpdk/interfaces and restarting the `dpdk` service:
* https://github.com/openstack-charmers/charm-layer-ovn/blob/b0cee1d3b1bf7e7fc4236d9b2057419d3411771b/lib/charms/ovn_charm.py#L617-L627
* https://github.com/openstack-charmers/charm-layer-ovn/blob/master/templates/etc_dpdk_interfaces

If this is not happening for you, I'd consider if configuration change is happening in the correct order or whether you have found an actual charm bug.

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -354,6 +356,13 @@ def test_enable_hardware_offload(self):
 
 class DPDKTest(test_utils.BaseCharmTest):
     """DPDK-related tests."""
+    _hugepages_configured = False
+
+    def setUp(self):
+        super().setUp()
+        if not DPDKTest._hugepages_configured:
+            self.enable_hugepages_vfio_on_hvs_in_vms(4)
+            DPDKTest._hugepages_configured = True
```
fnordahl: Use [`setUpClass`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass) / [`tearDownClass`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDownClass) instead, which is designed for execution of once per class activities.

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -365,6 +374,39 @@ def _openvswitch_switch_dpdk_installed(self):
...
+        macs=[]
+        for instance in networking_data.unit_machine_ids:
+            openstack_utils.lxd_maybe_add_nic(instance, ifname,
+                              "third")
+            macs.append(str(openstack_utils.lxd_get_nic_hwaddr(instance, ifname)))
+
+        if macs:
+            openstack_utils.configure_networking_charms(
+                networking_data, macs, use_juju_wait=False)
+        return macs
```
fnordahl: The act of adding an interface to test instances should be a deliberate action through configure / setup steps.

Either make `zaza.openstack.charm_tests.neutron.setup.undercloud_and_charm_setup` do what you want on the back of configuration influenced by environment variables, or add a new configure function for this purpose.

The use of the word second is also misleading as the instances likely already have two interfaces.

zaza/openstack/charm_tests/ovn/tests.py
```diff
@@ -492,6 +549,93 @@ def test_enable_dpdk(self):
...
+                port_name = row[0]
+                interfaces = row[1]
+                if port_name == "dpdk-bond0":
+                    if len(interfaces) == 2:
+                        logging.info("DPDK bond was successfully created with 2 interfaces")
+                        return
+                    else:
+                        logging.info("BOND FAILED")
+                        raise Exception('Could not create DPDK bond with 2 interfaces')
+                        return
```
fnordahl: Replace this is one of the unittest assert methods: https://docs.python.org/3/library/unittest.html#assert-methods

fnordahl (changes requested):

zaza/openstack/charm_tests/ovn/tests.py
PanosKostopoulos: I tried using setUpClass in the beginning since it was what we discussed. The issue is that enable_hugepages_vfio_on_hvs_in_vms is a self method, so in the setUpClass(cls) I do not have access to it. I could create a temp instance but I do not know if that is correct. Hence the use of SetUp(self). Am I missing something here.

fnordahl: Looking at `enable_hugepages_vfio_on_hvs_in_vms`, the only variables prefixed with `self` is `model_name` and `test_config`.

Both of these are already defined as class level variables in the parent `setUpClass` so they should be easily accessible.

Would passing `cls` as instance argument to the instance methods work directly perhaps?

PR #1344, opened 2026-04-21
Fix issues blocking rerun of config phase

### LINKED ISSUE #1343: Errors when re-running configure stage
When re-running the configure stage I hit the following errors:

First one happens if we try to unseal an already unsealed vault - https://pastebin.ubuntu.com/p/3gkMkbbDqb/

Second one happens if we do octavia-diskimage-retrofit when image already exists - https://pastebin.ubuntu.com/p/rd3QyYPBYk/

dosaboy:
Vault get_csr() and octavia-diskimage-retrofit will fail if re-run after a successful run so we now protect against this.

Fixes: #1343

### DIFF
--- a/zaza/openstack/charm_tests/vault/setup.py
+++ b/zaza/openstack/charm_tests/vault/setup.py
@@ -27,6 +27,13 @@ import zaza.openstack.utilities.generic
 import zaza.openstack.utilities.exceptions as zaza_exceptions
 import zaza.utilities.juju as juju_utils
 
+GET_CSR_FAIL_MSG = """
+This action will invalidate this intermediate CA chain until the signed csr is
+uploaded. During this time no new certificate requests will be processed. If
+you are sure you want to go ahead with this then please run the action with
+force=True
+""".strip().replace('\n', ' ')
+
 
 def get_cacert_file():
     """Retrieve CA cert used for vault endpoints and write to file.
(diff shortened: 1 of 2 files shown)


dosaboy:
still testing

dosaboy:
tested fresh run and re-run. seems good.

zaza/openstack/charm_tests/vault/setup.py
```diff
@@ -162,19 +162,24 @@ def auto_initialize(cacert=None, validation_application='keystone', wait=True,
...
-        intermediate_csr,
-        cakey.decode(),
-        cacertificate.decode(),
-        generate_ca=True)
-    action = vault_utils.run_upload_signed_csr(
-        pem=intermediate_cert,
-        root_ca=cacertificate,
-        allowed_domains='openstack.local')
+    if action.data['status'] == 'failed':
+        logging.info("vault get_csr() failed with force=False so assuming it "
```
freyes, line 174: I'm not comfortable with this assumption, we should address in the charm to not fail OR give you information on why it failed that can be used here to validate the assumption.

freyes (changes requested):

zaza/openstack/charm_tests/vault/setup.py
dosaboy: the charm supports using force=True but i dont think we want to use that here right? As an alternative I can check the status message so it is sure that we are getting the expected failure but this would be a change to zaza to include 'message' in https://github.com/openstack-charmers/zaza/blob/d3f1d55ca23434d0eb2404d76f178722e472ffa3/zaza/model.py#L1092

freyes: > As an alternative I can check the status message so it is sure that we are getting the expected failure [...]

this sounds more reasonable, let's take this approach.

dosaboy: done.

freyes (approved):
LGTM. Thanks for addressing the feedback

PR #1352, opened 2026-04-28
Fix barbican policyd tests

munirsidd:
The `BarbicanTests.get_client_and_attempt_operation()` method was manually resolving the barbican endpoint via `keystone_client.service_catalog.url_for()`,  which fails because session-based keystone clients don't populate service_catalog.

- Add `get_barbican_session_client()` to `zaza.openstack.utilities.openstack`, following the same pattern as `get_heat_session_client()` and other service clients
- Update `BarbicanTests` to use `openstack_utils.get_barbican_session_client()`, which lets the barbican client resolve the endpoint from the session automatically


dosaboy:
have tested this using master branch of barbican charm and all good

PR #1358, opened 2026-04-29
Launch vms with http proxy

dosaboy:
If an http proxy is available we need to set it in launched vms to allow them to access external sites.


freyes (approved):
LGTM.

I didn't find any other places where the proxy bits should be added to.

PR #1369, opened 2026-06-01
Retry pacemaker remote

freyes:
Make the test for pacemake-remote more robust by using tenacity to retry and cover for race conditions while pacemaker settles, and drop hardcoded node name in favor of picking a node from the available units.


MylesJP (approved):
LGTM

PR #1373, opened 2026-06-11
Work around Juju unit error after reboot by awaiting symlink recreation

fnordahl:
After enable_hugepages_vfio_on_hvs_in_vms reboots a VM, Juju may still be recreating symlinks for unit tools when the first config-changed hook fires, causing the unit to enter an error state (LP: #2077936).

Add a _wait_for_juju_symlinks_after_reboot method that monitors /var/log/juju/machine-*.log for both the Reboot and symlink recreation messages before allowing execution to continue. This is called after enable_hugepages returns from rebooting, both on the success and error-recovery paths.

Assisted-by: GLM-5.1, opencode


freyes (approved):
LGTM

PR #1379, opened 2026-06-11
Retry create_additional_port_for_machines on ConnectionTimeout

dosaboy:
If this is called when keystone services are for whatever reason being restarted a timeout occurs and a retry generally works.

### DIFF
--- a/zaza/openstack/utilities/openstack.py
+++ b/zaza/openstack/utilities/openstack.py
@@ -64,6 +64,7 @@ from designateclient.client import Client as DesignateClient
 from keystoneclient.v2_0 import client as keystoneclient_v2
 from keystoneclient.v3 import client as keystoneclient_v3
 from keystoneauth1 import session
+from keystoneauth1.exceptions.connection import ConnectTimeout
 from keystoneauth1.identity import (
     v3,
     v2,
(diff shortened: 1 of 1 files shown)


xtrusia (approved):
LGTM
Maybe better to add a stop condition though, e.g. stop=tenacity.stop_after_attempt(10)

zaza/openstack/utilities/openstack.py
```diff
@@ -982,6 +983,9 @@ def get_charm_networking_data(limit_gws=None):
         other_config)
 
 
+@tenacity.retry(wait=tenacity.wait_exponential(multiplier=1, max=60),
+                reraise=True, stop=tenacity.stop_after_attempt(10),
+                retry=tenacity.retry_if_exception_type(ConnectTimeout))
```
freyes, line 988: I like the idea, and this function is already taking care of not adding more ports to the instance if it already has a port.

zaza/openstack/utilities/openstack.py
freyes, line 1036: should we treat this in a more atomic way?, in case the `create_port` succeeds, but the `interface_attach` doesn't?, in that scenario we won't reusing the already created port, because `neutronclient.list_ports(device_id=server.id)['ports']`, the port won't be in this list.

freyes (changes requested):

PR #1386, opened 2026-06-17
Add subnet pool prefix to no_proxy

ghadi-rahme:
This should avoid issues with VMs connected to the loadbalancer

### DIFF
--- a/zaza/openstack/configure/guest.py
+++ b/zaza/openstack/configure/guest.py
@@ -24,6 +24,7 @@ import zaza.openstack.utilities.openstack as openstack_utils
 import zaza.openstack.charm_tests.nova.utils as nova_utils
 import zaza.openstack.utilities.exceptions as openstack_exceptions
 import zaza.utilities.deployment_env as deployment_env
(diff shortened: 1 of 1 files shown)


zaza/openstack/configure/guest.py
```diff
@@ -116,10 +117,15 @@ def get_default_userdata():
     inside launched guest vms.
     """
     deploy_env = deployment_env.get_deployment_context()
+    no_proxy = deploy_env.get('TEST_NO_PROXY') or ''
+    subnetpool = OVERCLOUD_NETWORK_CONFIG.get('subnetpool_prefix', '')
+
+    if subnetpool and subnetpool not in no_proxy:
+        no_proxy = '{},{}'.format(no_proxy, subnetpool) if no_proxy else subnetpool
```
dosaboy, line 124: this wont be enough for jammy which needs expanded list of addresses. im not convinced it is needed anyway. when we need to access the vm we can add its address to no_proxy e.g. when we do curl/wget

PR #1398, opened 2026-07-02
[stable/yoga] Fix designate-bind ServiceIPs test wait states

xtrusia:

### DIFF
--- a/zaza/openstack/charm_tests/designate_bind/tests.py
+++ b/zaza/openstack/charm_tests/designate_bind/tests.py
@@ -16,6 +16,13 @@
 import logging
 import os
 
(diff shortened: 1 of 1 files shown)


xtrusia:
> just a thought but, [d083dde](https://github.com/openstack-charmers/zaza-openstack-tests/commit/d083dde10c772a1fd9b1915707167d7723e245a5) will now exist in all branches except stable/zed, is it perhaps worth backporting there too?

ah right, I should have mentioned it it was already in zed few years ago
https://github.com/openstack-charmers/zaza-openstack-tests/commit/d5746dddca66e8526bed743fc718946e290b0a72

PR #1401, opened 2026-07-02
[stable/antelope] Fix designate-bind ServiceIPs test wait states

xtrusia:

### DIFF
--- a/zaza/openstack/charm_tests/designate_bind/tests.py
+++ b/zaza/openstack/charm_tests/designate_bind/tests.py
@@ -55,8 +55,7 @@ class DesignateBindServiceIPsTest(test_utils.OpenStackBaseTest):
                                 reraise=True,
                                 stop=stop_after_attempt(10)):
             with attempt:
-                configured_ips = zaza_model.run_on_unit(self.UNIT,
-                                                        "ip addr")
+                configured_ips = zaza_model.run_on_unit(self.UNIT, "ip addr")
                 self.assertIn(self.VIP, configured_ips["Stdout"])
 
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/designate_bind/tests.py
```diff
@@ -44,26 +45,32 @@ def setUp(self):
...
         logging.info("Removing service IP configuration from %s unit.",
                      self.UNIT)
         config["service_ips"] = ""
         zaza_model.set_application_config(self.APPLICATION, config)
-        zaza_model.wait_for_application_states()
+        zaza_model.wait_for_application_states(states=states)
 
-        configured_ips = zaza_model.run_on_unit(self.UNIT, "ip addr")
-        self.assertNotIn(self.VIP, configured_ips["Stdout"])
+        for attempt in Retrying(wait=wait_fixed(2),
```
freyes, line 70: this is no longer a clean cherry-pick from the original commit - https://github.com/openstack-charmers/zaza-openstack-tests/commit/8e968022e604559a0a066bd2bf3bd93eed0439fe - can you explain where this piece of code came from?

also cherry-pick needs to be created with the `-x` flag

freyes (changes requested):

zaza/openstack/charm_tests/designate_bind/tests.py
xtrusia: thank you, I added it as needed but after your comment, found out there is already patch. so cherry-picked them as well.

PR #1402, opened 2026-07-02
[stable/bobcat] Fix designate-bind ServiceIPs test wait states

xtrusia:


freyes (approved):
LGTM

PR #1409, opened 2026-07-03
Convert TEST_*_PROXY to *_proxy env vars

dosaboy:
If TEST_{NO,HTTP,HTTPS}_PROXY are provided, convert them to
http_proxy etc so that code will use them. This will apply
to all code under zaza.openstack.

### DIFF
--- a/zaza/openstack/__init__.py
+++ b/zaza/openstack/__init__.py
@@ -18,6 +18,16 @@ import os
 import tempfile
 
 
+def _set_proxy_vars():
+    # Set proxy vars if provided so that all code gets to use it.
+    for key, val in os.environ.items():
+        if not key.startswith('TEST_') or 'PROXY' not in key:
+            continue
+
+        _key = key.partition('TEST_')[2]
+        os.environ[_key.lower()] = val
+
+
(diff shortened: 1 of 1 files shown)


zaza/openstack/charm_tests/neutron/setup.py
```diff
@@ -80,6 +79,14 @@
     "fip_service_subnet_cidr": "100.64.0.0/24"
 }
 
+# Set proxy vars if provided so that all code gets to use it.
```
freyes: setting this at module definition level will make all code after this to get the side effect of setting these env variable.

This is a side effect that shouldn't escape this module's scope, although if that's an intentional behavior, I think it should be defined in `zaza/openstack/__init__.py` so this gets set no matter when/what imported this code ... and even in that scenario sounds easy to get into weirdness

freyes (changes requested):

zaza/openstack/charm_tests/neutron/setup.py
dosaboy: ah yes i like that, ill move it

freyes (approved):
LGTM. @dosaboy got positive results running functests with this patch.

PR #1424, opened 2026-07-23
Add manila policyd tests

munirsidd:
Add `ManilaTests` to policyd tests, using the `share:get_all` rule to verify policy overrides work with the Manila charm.


dosaboy (approved):
lgtm and has been tested with manila charm

